PR #70, opened 2021-08-10
Generate arbitrary named entity

### LINKED ISSUE #69: Error for sqlx_migrations table when generating entities
Hi guys!

Bumped in to a tiny error just now when trying to generate entities for my Postgres database. I've used sqlx-cli for the migrations.

```bash
❯ sea-orm-cli generate entity
Error writing files: failed to resolve mod `sqlx_migrations`: /Users/nilskanevad/projects/kekou-api does not exist
```

The default table name for sqlx migrations is `_sqlx_migrations` - Notice the underscore.

The codegen creates a `_sqlx_migrations.rs` file but the table_name is getting set to `sqlx_migrations`

```rust
//! src/entity/_sqlx_migrations.rs
impl EntityName for Entity {
  fn table_name(&self) -> &str {
    "sqlx_migrations"
  }
}
```

After changing the table name in my DB to `sqlx_migrations` instead of `_sqlx_migrations` and generating the entities again it works without a issue.

So, it seems that somewhere, somehow the cli/codegen gets confused by the "_" in the tablename. Can't find where in the code though

billy1624:
Resolve #69

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,8 +25,6 @@ jobs:
           toolchain: stable
           override: true
 
-      - uses: Swatinem/rust-cache@v1
-
       - uses: actions-rs/cargo@v1
         with:
           command: build
@@ -59,8 +57,6 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-
-      - uses: Swatinem/rust-cache@v1
 
       - uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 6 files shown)


tyt2y3:
When the check all passes, we should release a new codegen
And remove the path = in cli.toml

billy1624:
I guess it's okay to keep the line?

https://github.com/launchbadge/sqlx/blob/master/sqlx-macros/Cargo.toml
https://docs.rs/crate/sqlx-macros/0.5.5/source/Cargo.toml

tyt2y3:
I think the problem of path = is it will trick the CI server to not use the production build.

billy1624:
> I think the problem of path = is it will trick the CI server to not use the production build.

Ok, so I publish `sea-orm-codegen` & `cli-orm-cli` first, then remove the `path = "..."` from this PR?

PR #74, opened 2021-08-11
Support schemas in queries

### LINKED ISSUE #65: Support schemas in queries
As far as I'm aware, there seems to be a lack of support for querying from schemas. For example, switching from public schema to a different schema.

A possible solution is that `EntityName` should support a new function called `schema_name` perhaps:
```rust
pub trait EntityName {
    fn schema_name(&self) -> Option<&str> {
        None
    }
}
```
Where `None` would not include a schema selector in the queries.. and `Some(&str)` would append the schema before the table name in queries.

This would change queries from:
```sql
SELECT "users"."id", "users"."email" FROM "users" ...
```
to:
```sql
SELECT "user"."users"."id", "user"."users"."email" FROM "user"."users" ...
-- or
SELECT "public"."users"."id", "public"."users"."email" FROM "public"."users" ...
```

billy1624:
Resolve #65

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.1", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.12.8" }
+sea-query = { version = "^0.12.11" }
 sea-strum = { version = "^0.21", features = ["derive", "sea-orm"] }
 serde = { version = "^1.0", features = ["derive"] }
 sqlx = { version = "^0.5", optional = true }
--- a/src/entity/base_entity.rs
+++ b/src/entity/base_entity.rs
@@ -3,7 +3,7 @@ use crate::{
     ModelTrait, PrimaryKeyToColumn, PrimaryKeyTrait, QueryFilter, Related, RelationBuilder,
     RelationTrait, RelationType, Select, Update, UpdateMany, UpdateOne,
 };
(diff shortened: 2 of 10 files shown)


billy1624:
Ok, let's leave the `RelationDef` for another PR?

tyt2y3:
Yes agreed

billy1624:
Feel like this part should be handled by `sea-query`, mapping from `TableRef` to table name `Iden`.

https://github.com/SeaQL/sea-orm/blob/3d9b7457f6ace3fd2b498ee6c45aef3ca3d8f2fd/src/query/helper.rs#L298-L306

tyt2y3:
> Feel like this part should be handled by `sea-query`, mapping from `TableRef` to table name `Iden`.
> 
> https://github.com/SeaQL/sea-orm/blob/3d9b7457f6ace3fd2b498ee6c45aef3ca3d8f2fd/src/query/helper.rs#L298-L306

Did I wrote this code?

billy1624:
@tyt2y3 check check

billy1624:
> > Feel like this part should be handled by `sea-query`, mapping from `TableRef` to table name `Iden`.
> > https://github.com/SeaQL/sea-orm/blob/3d9b7457f6ace3fd2b498ee6c45aef3ca3d8f2fd/src/query/helper.rs#L298-L306
> 
> Did I wrote this code?

I wrote

tyt2y3:
Okay, well I don't have a thought yet. Good enough for now if it works.

billy1624:
> Okay, well I don't have a thought yet. Good enough for now if it works.

Hahaaa, okay!

Issue #76, opened 2021-08-11
Escape column names that conflict with Rust keywords in CLI


Acidic9:
When generating models with the sea-orm-cli, if you have a column with a name that conflicts with a Rust keyword, the cli crashes and is not handled correctly.

For example, an SQL table:
```
| column_name | data_type |
|-------------------------|
| id          | serial    |
| type        | varchar   |
```
I get an error:
```
11 | ... { pub id : i64 , pub type : Option < String > }
   |                          ^^^^ expected identifier, found keyword
```
This is caused because the column `type` conflicts with the Rust keyword `type`.

Any rust columns conflicting should be renamed to resolve this issue.
https://doc.rust-lang.org/reference/keywords.html

tyt2y3:
Any suggestion on the naming convention? How about an underscore at the end?

Acidic9:
I didn't make any suggestions for this because I am not sure 😄 

I guess `type_` with an underscore could work.. or `type_col` (adding `_col`). I'm not sure of any better options than these.

billy1624:
I think we can leave it for the user to decide how to resolve the name conflict.

Personally think that neither `r#type` nor `type_` be a good option. For me, I would simply choose another name to avoid conflicting with Rust keywords.

```rust
#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel)]
pub struct Model {
    pub id: i32,
    pub type_of: String,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveCustomColumn)]
pub enum Column {
    Id,
    TypeOf,
}

impl IdenStatic for Column {
    fn as_str(&self) -> &str {
        match self {
            Self::TypeOf => "type",
            _ => self.default_as_str(),
        }
    }
}
```

Acidic9:
The problem is that users will integrate existing databases with SeaORM.. and usually you don't want your framework / language to dictate your database, it should be flexible enough to avoid issues like this

billy1624:
Good point!

PR #79, opened 2021-08-12
Customize column names

### LINKED ISSUE #43: Support non snake-case column name
We assume all columns in database table are in snack-case.

```rust
#[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
pub enum Column {
    Id,
    Name,
}
```

We need to support an optional derive attribute to override this assumption.

```rust
#[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
#[DeriveColumn(case = "camel")]
pub enum Column {
    Id,
    Name,
}
```

Or, customize column name of each variants

```rust
#[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
pub enum Column {
    #[DeriveColumn(name = "ID")]
    Id,
    Name,
}
```

billy1624:
Resolve #43


Acidic9:
> I insist that after we released a new version of sea-orm-macro, we bump the dependency and remove the `path =` from toml to ensure that our CI uses the released version.

I noticed that it used path also, wasn't sure if it was a mistake or not.

This is awesome, you guys work so fast. Is there any examples of how to use this feature to rename columns?

tyt2y3:
https://github.com/SeaQL/sea-orm/blob/8b30ef74a22153ebdb2e0cf65afd411fe68ba8a6/src/tests_cfg/filling.rs#L19-L34

billy1624:
We want to keep it simple and avoid using excessive macros loll

tyt2y3:
Though, I think we can instead use override:
```rust
 impl IdenStatic for Column { 
     // the default impl is self.as_str()
     fn rename(&self) -> &str { 
         match self { 
             Self::Id => "id", 
             _ => self.as_str(), 
         } 
     } 
 } 
```

Acidic9:
Ohh okay I see 😄 Serde uses a macro for renaming, but I guess the `default_as_str()` approach works fine too.. just a little bit more boiletplate code

Acidic9:
Does the renaming also need to be implemented for `DeriveModel`? I'm not sure how to handle a model with fields which need to be renamed.

billy1624:
Our assumption on naming model attributes & column enum variants:
- All model attributes are in snake-case and exists in column enum variants
  - Model attributes: `id`, `name`
- All column variants are in camel-case and exists in model attributes
  - Enum variants: `Id`, `Name`

So, that means if you want to rename one of the column / attribute you have to also update the other half of it.

Acidic9:
I'm still a little confused, for the column you would rename it with `DeriveCustomColumn` and `impl IdenStatic for Column`.. but how do we rename a field on the model too? For exampe, if a field conflicts with a Rust keyword.

billy1624:
So basically the `impl IdenStatic for Column` is just for constructing SQL statement, i.e. how we refer this column in SQL statement.

Take your example in #76... we might do the followings

```rust
#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel)]
pub struct Model {
    pub id: i32,
    pub type_of: String,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveCustomColumn)]
pub enum Column {
    Id,
    TypeOf,
}

impl IdenStatic for Column {
    fn as_str(&self) -> &str {
        match self {
            Self::TypeOf => "type",
            _ => self.default_as_str(),
        }
    }
}
```

Acidic9:
Ah perfect, it all makes sense now thank you @billy1624 ... again XD !!!

PR #80, opened 2021-08-13
Custom column names with default

billy1624:
As suggested on #79


billy1624:
Now we have...

https://github.com/SeaQL/sea-orm/blob/7063ff6ec983fbbaa8e8c396d292fd5c5370b261/src/tests_cfg/filling.rs#L19-L36

tyt2y3:
Remember to release a new ver and change our Cargo

PR #84, opened 2021-08-14
Fix compile errors on different runtime

billy1624:


sea-orm-cli/Cargo.toml
```diff
@@ -29,7 +29,9 @@ sea-schema = { version = "^0.2.4", default-features = false, features = [
 sqlx = { version = "^0.5", default-features = false, features = [ "mysql", "postgres" ] }
 
 [features]
-default = [ "runtime-async-std-native-tls" ]
+# Warning: removing this line means one cannot simply run `cargo install sea-orm-cli`.
+# They have to run `cargo install sea-orm-cli --features runtime-async-std-native-tls`
+# default = [ "runtime-async-std-native-tls" ]
```
tyt2y3, line 34: sea-orm-cli is an app. People shouldn't care about which runtime we are using.
I think we should ship a default runtime.
If it's so problematic to be in the same workspace, we can split it out to its own workspace, or even another repo.

billy1624: Agree, split it out to its workspace worked!

tyt2y3: Cool. Let's try.

billy1624:
Check this @tyt2y3

tyt2y3:
Nice

Issue #86, opened 2021-08-15
Compile error '`Sync` is not implemented' under tokio


Acidic9:
Related issue: https://github.com/SeaQL/sea-query/issues/105

With the following function in a tokio async runtime, I get an error saying:
```
future created by async block is not `Send`
help: the trait `Sync` is not implemented for `(dyn Iden + 'static)`
```
```rust
async fn update(
    db: &DatabaseConnection,
    active_model: model::ActiveModel,
) -> Result<bool, Box<dyn std::error::Error>> {
    Entity::update(active_model)
        .exec(db)
        .await
        .map_err(Box::new)?;

    Ok(true)
}
```

SeaORM does not seem to be thread safe. Perhaps this is due to sea-query's Iden type.

tyt2y3:
Can you setup an minimal reproducible example under /examples?

Acidic9:
I've fixed it locally... the quick fix seems to be changing the definition of `trait Iden` in sea-query to:
```rust
pub trait Iden: Send + Sync {
    ...
}
```
This is to enforce the type is Sync

tyt2y3:
Great finding. Can we instead do it inside SeaORM, changing IdenStatic instead of Iden?
```rust
pub trait IdenStatic: Iden + Copy + Debug + Send + Sync + 'static {
```

I am trying to understand the problem

Acidic9:
I've tried removing Send + Sync from `Iden` trait in sea-query and adding it to `IdenStatic` like you suggested, but the errors come back.

T inside `Arc<T>` should be Send + Sync for the whole Arc to be Send + Sync, but it's not required.
https://doc.rust-lang.org/std/sync/struct.Arc.html#thread-safety
It says:
> `Arc<T>` will implement `Send` and `Sync` as long as the `T` implements `Send` and `Sync`. Why can’t you put a non-thread-safe type `T` in an `Arc<T>` to make it thread-safe? This may be a bit counter-intuitive at first: after all, isn’t the point of `Arc<T>` thread safety? The key is this: `Arc<T>` makes it thread safe to have multiple ownership of the same data, but it doesn’t add thread safety to its data.

tyt2y3:
Okay, good to know. Thanks.

tyt2y3:
Btw, any idea how can I add a test case to our test suite?

Acidic9:
> Btw, any idea how can I add a test case to our test suite for this?

[...]

PR #91, opened 2021-08-17
Test against all supported runtime

billy1624:
Related to #86

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,9 +13,6 @@ jobs:
   test:
     name: Unit Test
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        runtime: [async-std-native-tls, async-std-rustls, actix-native-tls, actix-rustls, tokio-native-tls, tokio-rustls]
     steps:
       - uses: actions/checkout@v2
 
@@ -25,13 +22,14 @@ jobs:
           toolchain: stable
           override: true
 
+      - uses: Swatinem/rust-cache@v1
+
       - uses: actions-rs/cargo@v1
         with:
           command: build
           args: >
             --all
             --exclude 'sea-orm-example-*'
-            --features default,runtime-${{ matrix.runtime }}
 
       - uses: actions-rs/cargo@v1
         with:
@@ -39,7 +37,6 @@ jobs:
           args: >
             --all
             --exclude 'sea-orm-example-*'
-            --features default,runtime-${{ matrix.runtime }}
 
   cli:
     name: CLI
@@ -56,6 +53,8 @@ jobs:
           toolchain: stable
           override: true
 
+      - uses: Swatinem/rust-cache@v1
+
       - uses: actions-rs/cargo@v1
         with:
           command: install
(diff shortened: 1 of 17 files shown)


tyt2y3:
Brilliant. A macro ?

src/runtime.rs
```diff
@@ -0,0 +1,33 @@
+#[cfg(any(
+    feature = "runtime-async-std",
+    all(
```
tyt2y3, line 3: This any all combo is a little strange?

billy1624:
Macro! :P

src/runtime.rs
billy1624: To support simple `cargo test` without specifying any runtime preference

billy1624: Default use async-std

tyt2y3: I still don't understand why is it
```
async-std OR (!async-std AND !actix AND !tokio)
```
if any runtime is specified, then we use that runtime.
if no runtime is specified, then we simply do nothing?

tyt2y3:
As per our discussion, I think we'd try:

Using `smol` async runtime for unit & doc tests

Require at least one of the production runtimes available for running integration tests.

.github/workflows/rust.yml
```diff
@@ -174,27 +207,31 @@ jobs:
...
     name: Postgres
     runs-on: ubuntu-20.04
+    needs: check
     env:
       DATABASE_URL: "postgres://root:root@localhost"
     strategy:
       matrix:
-        version: [13.3, 12.7, 11.12, 10.17, 9.6.22]
-        runtime: [async-std-native-tls]
+        version: [13.3, 12.7]
```
tyt2y3, line 232: We could not support old versions of Postgres?

billy1624: We can just thinking if it's too much loll

billy1624: Let me add it back then

billy1624:
Btw... for Postgres integration tests, this happened from time to time...

> thread 'left_join' panicked at 'could not insert baker: Query("error returned from database: column \"bakery_id\" is of type integer but expression is of type boolean")', tests/relational_tests.rs:51:6

https://github.com/SeaQL/sea-orm/runs/3362229412?check_suite_focus=true#step:6:107

billy1624:
Correct! 👻

tyt2y3:
> > Btw... for Postgres integration tests, this happened from time to time...
> > > thread 'left_join' panicked at 'could not insert baker: Query("error returned from database: column "bakery_id" is of type integer but expression is of type boolean")', tests/relational_tests.rs:51:6
> > 
> > 
> > https://github.com/SeaQL/sea-orm/runs/3362229412?check_suite_focus=true#step:6:107
> 
> 💩 from time to time? there is a cause of non-determinism?

Well. That's bad. @samsamai would you by chance has some guess?

billy1624:
All these suffer the same error loll
- https://github.com/SeaQL/sea-orm/runs/3367814059?check_suite_focus=true#step:6:106
- https://github.com/SeaQL/sea-orm/runs/3362229412?check_suite_focus=true#step:6:107
- https://github.com/SeaQL/sea-orm/runs/3361018305?check_suite_focus=true#step:7:107
- https://github.com/SeaQL/sea-orm/runs/3360297887?check_suite_focus=true#step:7:107

tyt2y3:
I have a remote idea
https://github.com/SeaQL/sea-query/blob/b384d4d26aea0ae2ef1bc112558d2cd07a798544/src/driver/sqlx_postgres.rs#L7

tyt2y3:
I have a theory now. The prepared statement is exectued elsewhere with an integer value, so when it's executed 2nd time, the protocol would complain type mismatch.

tyt2y3:
https://www.postgresql.org/docs/13/protocol-message-formats.html

billy1624:
We can merge this PR frist? As this issue is solely related to sea-query.

PR #93, opened 2021-08-18
Implement `FromStr` for `DeriveColumn`

Acidic9:
Adds an implementation for enums which derive `DeriveColumn` making them import `std::str::FromStr`.

The `FromStr` supports the column names as snake_case and mixedCase.
```rust
#[derive(DeriveColumn)]
pub enum Column {
    Id,
    Name,
    CakeId,
}

Column::from_str("id");       // Ok(Column::Id)
Column::from_str("name");     // Ok(Column::Name)
Column::from_str("cake_id");  // Ok(Column::CakeId)
Column::from_str("cakeId");   // Ok(Column::CakeId)
Column::from_str("non-existent column");  // Err(ParseColumnErr)
```

I've written a test for this in `/src/entity/column.rs`.

### DIFF
--- a/sea-orm-macros/src/derives/column.rs
+++ b/sea-orm-macros/src/derives/column.rs
@@ -1,6 +1,6 @@
-use heck::SnakeCase;
+use heck::{MixedCase, SnakeCase};
 use proc_macro2::{Ident, TokenStream};
-use quote::{quote, quote_spanned};
+use quote::{format_ident, quote, quote_spanned};
 use syn::{Data, DataEnum, Fields, Variant};
 
 pub fn impl_default_as_str(ident: &Ident, data: &Data) -> syn::Result<TokenStream> {
@@ -41,6 +41,44 @@ pub fn impl_default_as_str(ident: &Ident, data: &Data) -> syn::Result<TokenStrea
     ))
 }
 
+pub fn impl_col_from_str(ident: &Ident, data: &Data) -> syn::Result<TokenStream> {
+    let parse_error_iden = format_ident!("Parse{}Err", ident);
+
+    let data_enum = match data {
+        Data::Enum(data_enum) => data_enum,
+        _ => {
+            return Ok(quote_spanned! {
+                ident.span() => compile_error!("you can only derive DeriveColumn on enums");
+            })
+        }
+    };
+
+    let columns = data_enum.variants.iter().map(|column| {
+        let column_iden = column.ident.clone();
+        let column_str_snake = column_iden.to_string().to_snake_case();
+        let column_str_mixed = column_iden.to_string().to_mixed_case();
+        quote!(
+            #column_str_snake | #column_str_mixed => Ok(#ident::#column_iden)
+        )
+    });
+
+    Ok(quote!(
+        #[derive(Debug, Clone, Copy, PartialEq)]
+        pub struct #parse_error_iden;
+
+        impl std::str::FromStr for #ident {
+            type Err = #parse_error_iden;
+
+            fn from_str(s: &str) -> Result<Self, Self::Err> {
+                match s {
+                    #(#columns),*,
+                    _ => Err(#parse_error_iden),
+                }
+            }
+        }
+    ))
+}
+
 pub fn expand_derive_column(ident: &Ident, data: &Data) -> syn::Result<TokenStream> {
     let impl_iden = expand_derive_custom_column(ident, data)?;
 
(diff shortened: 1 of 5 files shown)


src/query/combine.rs
```diff
@@ -167,7 +167,7 @@ mod tests {
                 .left_join(fruit::Entity)
                 .select_also(fruit::Entity)
                 .filter(cake::Column::Id.eq(1))
-                .filter(fruit::Column::Id.eq(2))
+                .filter(ColumnTrait::eq(&fruit::Column::Id, 2))
```
Acidic9, line 170: I had to specify ColumnTrait here as I've added `#[derive(PartialEq)]` to `fruit::Column` to test the `FromStr` results.

tyt2y3: Thank you. I understand now. It makes the 'eq' method ambiguous I suppose.

tyt2y3:
@billy1624 Thoughts?

billy1624:
Interesting! Can I know the use case of it? @Acidic9

Acidic9:
In my case, it's useful with graphql for only selecting the columns specified by the users graphql query. This is the main use case, I'm not sure of many others.

billy1624:
So you parse GraphQL query and construct the SQL with SeaORM? Cool!

billy1624:
I think this could be useful to anyone wants to parse arbitrary string input into concrete column enum variant. Without the need to write a custom `match` or chained `if` statements.

Acidic9:
Yeah exactly!
In my case I use it with:
```rust
let columns: Vec<_> = ctx
    .field()
    .selection_set()
    .filter_map(|selection| coupon::Column::from_str(selection.name()).ok())
    .collect();
    
columns
    .into_iter()
    .fold(Coupon::find().select_only(), |acc, col| acc.column(col))
    .all(db)
    .await?;
```
So it selects only columns from the graphql query.

Acidic9:
This also makes the use case for `#[derive(DeriveModel, DerivePartialModel)]` very useful, so you can select a dynamic subset of columns, and use `.into_model::<cake::PartialModel>()` where `PartialModel` is every column wrapped in `Option<T>`.

billy1624:
> This also makes the use case for `#[derive(DeriveModel, DerivePartialModel)]` very useful, so you can select a dynamic subset of columns, and use `.into_model::<cake::PartialModel>()` where `PartialModel` is every column wrapped in `Option<T>`.

For the column already wrapped in `Option<T>` will it be wrapped twice in `PartialModel`?

Acidic9:
> For the column already wrapped in `Option<T>` will it be wrapped twice in `PartialModel`?

I started implementing `DerivePartialModel`... and no I was thinking it will only be wrapped in one `Option<T>`, never `Option<Option<T>>`. I don't see any use in having the multi-layered option in this case.

billy1624:
> > For the column already wrapped in `Option<T>` will it be wrapped twice in `PartialModel`?
> 
> I started implementing `DerivePartialModel`... and no I was thinking it will only be wrapped in one `Option<T>`, never `Option<Option<T>>`. I don't see any use in having the multi-layered option in this case.

Yes, I was thinking the same thing.

Acidic9:
I've made a PR for implementing `PartialModel`: https://github.com/SeaQL/sea-orm/pull/95

sea-orm-macros/src/derives/column.rs
```diff
@@ -41,6 +41,44 @@ pub fn impl_default_as_str(ident: &Ident, data: &Data) -> syn::Result<TokenStrea
...
+            return Ok(quote_spanned! {
+                ident.span() => compile_error!("you can only derive DeriveColumn on enums");
+            })
+        }
+    };
+
+    let columns = data_enum.variants.iter().map(|column| {
+        let column_iden = column.ident.clone();
+        let column_str_snake = column_iden.to_string().to_snake_case();
+        let column_str_mixed = column_iden.to_string().to_mixed_case();
```
tyt2y3, line 59: Why do you want to favour MixedCase over CamelCase?
The different between them is in Camel, the first character is also captialized:
```
CamelCase
mixedCase
```

tyt2y3:
We might as well 'inherit' `std::str::FromStr` for `ColumnTrait`

sea-orm-macros/src/derives/column.rs
Acidic9: Yes because in my case graphql uses `mixedCase`.

I actually always thought camel case was `camelCase`, not `CamelCase`.

tyt2y3: Good to know. So the GraphQL convention is to use `camelCase`. Well maybe that's a `heck` naming thing.

PR #95, opened 2021-08-18
WIP: Add `DerivePartial` macro

Acidic9:
*Work in progress.*

Adds a new derive macro called `DerivePartialModel`. This is similar to an `ActiveModel`, but instead wraps each column in an `Option<T>`.

This is useful for querying a subset of columns without defining a new type, which is particularly useful for dynamic queries built at runtime based on user input.

Example usage:
```rust
#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel, DerivePartialModel)]
pub struct Model {
    pub id: i32,
    pub name: String,
    pub difficulty: Option<i32>,
}
```
This generates a new struct called `PartialModel` and implements `FromQueryResult` on it:
```rust
// Generated code ...

#[derive(Clone, Default, Debug, PartialEq)]
pub struct PartialModel {
    pub id: Option<i32>,
    pub name: Option<String>,
    pub difficulty: Option<i32>,
}

impl FromQueryResult for PartialModel { ... }
```
You can use the `PartialModel` struct by calling `.into_model::<PartialModel>()` on a query.
```rust
// Example usage
let cakes = Cake::find()
    .select_only()
    .column(cake::Column::Name)
    .column(cake::Column::Difficulty)
    .into_model::<cake::PartialModel>()
    .all(db)
    .await?;
```

I've marked the pull request as Work In Progress to give time for feedback, and because I have not written any tests for this yet (but works in my personal project).

### DIFF
--- a/sea-orm-macros/src/derives/mod.rs
+++ b/sea-orm-macros/src/derives/mod.rs
@@ -4,6 +4,7 @@ mod column;
 mod entity;
 mod from_query_result;
 mod model;
+mod partial_model;
 mod primary_key;
 
 pub use active_model::*;
@@ -12,4 +13,5 @@ pub use column::*;
 pub use entity::*;
 pub use from_query_result::*;
 pub use model::*;
+pub use partial_model::*;
 pub use primary_key::*;
--- a/sea-orm-macros/src/derives/partial_model.rs
+++ b/sea-orm-macros/src/derives/partial_model.rs
@@ -1,0 +1,92 @@
+use proc_macro2::{Ident, TokenStream};
+use quote::{quote, quote_spanned};
+use syn::{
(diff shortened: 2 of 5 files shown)


billy1624:
I think it's a good feature to have if someone want to aggregate on some column without selecting all columns. Then, they can simply call `.into_model::<PartialModel>()`.

https://github.com/SeaQL/sea-orm/blob/9ee4c24114763ec4048816435ec0846bd9fe21ad/tests/relational_tests.rs#L55-L74

tyt2y3:
@Acidic9 are you still using this, or do you have new ideas based on `SimpleInput`?

Acidic9:
@tyt2y3 I think for now it can be closed... the SimpleInput would replace this actually

PR #100, opened 2021-08-21
#86 Test

tyt2y3:
Just for experimenting why `Arc` is necessary.


tyt2y3:
It does not seem to fail even if the updated test suite
@acidic9 can you advice how to actually reproduce the original error?

Acidic9:
Did you try the comment I posted here? https://github.com/SeaQL/sea-orm/issues/86#issuecomment-899032492

tyt2y3:
> @tyt2y3 Did you try the comment I posted here? [#86 (comment)](https://github.com/SeaQL/sea-orm/issues/86#issuecomment-899032492)

Thanks I reproduced it now. Okay so the key is `tokio::spawn`.

Issue #59, opened 2021-07-27
Transforming an Entity into TableCreateStatement


tyt2y3:
https://github.com/SeaQL/sea-orm/blob/08acc2a06a88a1198c27cf73c3d42dee299b0ec7/tests/common/setup/schema.rs#L29-L51

To avoid duplication of information, we can actually generate a SeaQuery SchemaStatement entirely from the Entity definition.
It would be useful in 2 scenarios:

1. writing test cases i.e. original model is MySQL but we want to test in SQLite. this is particularly helpful if the migration files are written in raw SQL. we effectively automatically translated from MySQL to SQLite!

2. writing SQLite applications. in this case, apps don't need a separate migration tool. everything should be done by the app itself. so it's actually helpful to automatically bootstrap a SQLite database from an Entity file

tyt2y3:
@scrblue Would you be interested in this task?

PR #102, opened 2021-08-23
Create Table Statement

bobbyng626:
#59


billy1624:
Hey @bobbyng626, can you add some test cases? I saw you got some for debugging during development.

E.g. asserting the output of `create_table_statement_of` against a handwritten CreateTableStatement

src/entity/create_stmt.rs
```diff
@@ -0,0 +1,172 @@
...
+    use crate::{CreateStatementOf, tests_cfg};
+
+    #[test]
+    fn test_create_statement_tests_cfg_cake() {
+        let create_statement = tests_cfg::cake::Entity::create_table_statement_of(tests_cfg::cake::Entity);
+        let table = format!("{:?}", create_statement.get_table_name());
+        let columns = format!("{:?}", create_statement.get_columns());
+        let relations = format!("{:?}", create_statement.get_foreign_key_create_stmts());
+        let indexs = format!("{:?}", create_statement.get_indexes());
+        let result = format!("{:?}", create_statement);
```
tyt2y3, line 165: We should not be using the debug formatter to construct our test case.
@billy1624 please help changing to concrete structs

src/entity/create_stmt.rs
```diff
@@ -0,0 +1,172 @@
+use crate::{ColumnDef as Sea_Orm_Column_Def, EntityTrait, PrimaryKeyTrait, RelationDef};
+use crate::entity::column::ColumnTrait;
+use crate::entity::relation::RelationTrait;
+use crate::entity::primary_key::PrimaryKeyToColumn;
+use sea_query::{Alias, ColumnDef as Sea_Query_Column_Def, ColumnType as Sea_Query_Column_Type, ForeignKeyCreateStatement, IndexCreateStatement, SeaRc, TableCreateStatement, TableRef};
+pub use sea_strum::IntoEnumIterator;
+pub trait CreateStatementOf
```
tyt2y3, line 7: Perhaps we dont need a trait here

src/entity/create_stmt.rs
```diff
@@ -0,0 +1,172 @@
...
+            if sea_orm_column_def.unique {
+                sea_query_column_def.unique_key();
+                index.unique();
+            }
+            if sea_orm_column_def.indexed {
+                index.table(entity);
+                index.col(col);
+                stmt.index(&mut index);
+            }
+            match Sea_Query_Column_Type::from(sea_orm_column_def.col_type) {
```
tyt2y3, line 79: why are we using Camel_Snake here?

src/entity/create_stmt.rs
```diff
@@ -0,0 +1,172 @@
+use crate::{ColumnDef as Sea_Orm_Column_Def, EntityTrait, PrimaryKeyTrait, RelationDef};
+use crate::entity::column::ColumnTrait;
+use crate::entity::relation::RelationTrait;
+use crate::entity::primary_key::PrimaryKeyToColumn;
+use sea_query::{Alias, ColumnDef as Sea_Query_Column_Def, ColumnType as Sea_Query_Column_Type, ForeignKeyCreateStatement, IndexCreateStatement, SeaRc, TableCreateStatement, TableRef};
+pub use sea_strum::IntoEnumIterator;
+pub trait CreateStatementOf
+{
+    fn create_table_statement_of<E>(entity: E) -> TableCreateStatement where E: EntityTrait {
```
tyt2y3, line 9: Perhaps we only need a function named `entity_to_table_create_statement`

src/entity/mod.rs
```diff
@@ -6,6 +6,7 @@ mod model;
 pub mod prelude;
 mod primary_key;
 mod relation;
+mod create_stmt;
```
tyt2y3, line 9: I am not sure how should we name this module, perhaps `schema`?

src/entity/create_stmt.rs
billy1624: Done!

billy1624:
Thinking should we include `on_update` & `on_delete` settings on `sea_orm::Relation`

https://github.com/bobbyng626/sea-orm/blob/1ad1767457a5e320692b9a61db7d60560695c11c/tests/common/setup/schema.rs#L66-L73

tyt2y3:
> Thinking should we include `on_update` & `on_delete` settings on `sea_orm::Relation`

Seemingly yes

tyt2y3:
> > Thinking should we include `on_update` & `on_delete` settings on `sea_orm::Relation`
> 
> 
> 
> Seemingly yes

We don't have to change the codegen for now.

I think it's worth to add because then we don't have to modify the original test cases.

billy1624:
Btw... why we have a seperate `sea_orm::ColumnType` which mirroring `sea_query::ColumnType`?
https://github.com/bobbyng626/sea-orm/blob/1ad1767457a5e320692b9a61db7d60560695c11c/src/entity/column.rs#L13-L36

Cargo.toml
```diff
@@ -32,7 +32,7 @@ futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.1.1", optional = true }
-sea-query = { version = "^0.15", features = ["thread-safe"] }
+sea-query = { version = "^0.15", git = "https://github.com/SeaQL/sea-query.git", features = ["thread-safe"] }
```
billy1624, line 35: Use sea-query on crates.io after new version published

tyt2y3:
Two reasons:

1) I want it to have the def method. It shouldn't be a problem though. We can make a custom trait indeed.

2) I think for example some numeric columns, the precision parameter is completely irrelevant and is being deprecated anyway.

So either we break SeaQuery, or we have to do the extra mapping in SeaORM.

Cargo.toml
tyt2y3: Yea 0.16 is coming

tyt2y3: [0.16.0](https://github.com/SeaQL/sea-query/releases/tag/0.16.0)

tyt2y3: image

PR #104, opened 2021-08-23
Represent several relations between same types

### LINKED ISSUE #89: Represent several relations between same types
Example problematic cases:

1. Entities A et B have several different many-to-many relations
2. Entity A has several one-to-many relations to entity B (Eg: `COUPLE.member1 -> USER` and `COUPLE.member2 -> USER`)

In these cases there are several issues:

1. the codegen uses A/B entity names to generate `Relation` variants, so we obtain duplicate variants. I guess it could use the join table or foreign key entity name
2. the `Related` trait is typed on A/B entities, so same issue. So I believe fixing this requires a breaking conception change...

billy1624:
Resolve #89

### DIFF
--- a/tests/common/bakery_chain/baker.rs
+++ b/tests/common/bakery_chain/baker.rs
@@ -81,4 +81,22 @@ impl Related<super::cake::Entity> for Entity {
     }
 }
 
+pub struct BakedForCustomer;
+
+impl Linked for BakedForCustomer {
+    type FromEntity = Entity;
+
+    type ToEntity = super::customer::Entity;
+
+    fn link() -> Vec<RelationDef> {
+        vec![
+            super::cakes_bakers::Relation::Baker.def().rev(),
+            super::cakes_bakers::Relation::Cake.def(),
+            super::lineitem::Relation::Cake.def().rev(),
+            super::lineitem::Relation::Order.def(),
+            super::order::Relation::Customer.def(),
+        ]
+    }
+}
+
 impl ActiveModelBehavior for ActiveModel {}
--- a/tests/common/bakery_chain/baker.rs
+++ b/tests/common/bakery_chain/baker.rs
@@ -88,7 +88,7 @@ impl Linked for BakedForCustomer {
 
     type ToEntity = super::customer::Entity;
 
-    fn link() -> Vec<RelationDef> {
+    fn link(&self) -> Vec<RelationDef> {
         vec![
             super::cakes_bakers::Relation::Baker.def().rev(),
             super::cakes_bakers::Relation::Cake.def(),
--- a/tests/relational_tests.rs
+++ b/tests/relational_tests.rs
@@ -1,7 +1,8 @@
 use chrono::offset::Utc;
 use rust_decimal::prelude::*;
 use rust_decimal_macros::dec;
-use sea_orm::{entity::*, query::*, FromQueryResult};
+use sea_orm::{entity::*, query::*, DbErr, FromQueryResult};
+use uuid::Uuid;
 
 pub mod common;
 pub use common::{bakery_chain::*, setup::*, TestContext};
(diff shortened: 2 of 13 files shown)


billy1624:
Just a simple POC @tyt2y3

tyt2y3:
okay looks reasonable. But the intention of `find_related` is to select the `ToEntity`, and thus, it walks backwards from the `ToEntity` back to `FromEntity`

billy1624:
> okay looks reasonable. But the intention of `find_related` is to select the `ToEntity`, and thus, it walks backwards from `ToEntity` back to `FromEntity`

I walk (join) in the wrong direction here?

https://github.com/SeaQL/sea-orm/blob/31941d3af678eb46d510c31c941fa056d5d01015/src/entity/relation.rs#L38-L44

tyt2y3:
I think we were using join_rev
https://github.com/SeaQL/sea-orm/blob/8cfa14233d2d6bce27857cce8b12a95be414d2cc/src/query/helper.rs#L121-L128

tests/common/bakery_chain/baker.rs
```diff
@@ -81,4 +81,22 @@ impl Related<super::cake::Entity> for Entity {
...
 }
 
+pub struct BakedForCustomer;
+
+impl Linked for BakedForCustomer {
+    type FromEntity = Entity;
+
+    type ToEntity = super::customer::Entity;
+
+    fn link() -> Vec<RelationDef> {
```
tyt2y3, line 91: I think we should have self here, i.e. `fn link(&self)`

tyt2y3:
At least for the scope of this PR we are not changing the codegen.
But as a rough idea, I think these 'links' does not belong to the Entity file.
I am thinking to put them in a dedicated module because it works across many Entities.

tests/common/bakery_chain/baker.rs
billy1624: Okay, will update it now.

Btw... what is the reason behind this? lolll

tyt2y3: I think it will open a lot of possibilities when the struct is not a unit struct.

billy1624: Got it

tests/relational_tests.rs
```diff
@@ -474,3 +475,202 @@ pub async fn having() {
...
+       LEFT JOIN `cake` ON `cakes_bakers`.`cake_id` = `cake`.`id`
+       LEFT JOIN `lineitem` ON `cake`.`id` = `lineitem`.`cake_id`
+       LEFT JOIN `order` ON `lineitem`.`order_id` = `order`.`id`
+       LEFT JOIN `customer` ON `order`.`customer_id` = `customer`.`id`
+    */
+    let baked_for_customers = Baker::find()
+        .find_also_linked(baker::BakedForCustomer)
+        .all(&ctx.db)
+        .await?;
+    println!("{:#?}", baked_for_customers);
```
tyt2y3, line 671: Please assert against the content 🥗

billy1624: Updated

tests/relational_tests.rs
```diff
@@ -673,7 +673,7 @@ pub async fn linked() -> Result<(), DbErr> {
         .group_by(customer::Column::Name)
         .order_by_asc(baker::Column::Id)
         .order_by_asc(customer::Column::Id)
-        .into_model::<BakerLite, CustomerLite>()
+        .into_model()
         .all(&ctx.db)
         .await?;
```
billy1624, line 678: Wowww getting rid of turbofish :P

tests/relational_tests.rs
```diff
@@ -474,3 +475,239 @@ pub async fn having() {
...
+
+    #[derive(Debug, FromQueryResult, PartialEq)]
+    struct CustomerLite {
+        name: String,
+    }
+
+    let baked_for_customers: Vec<(BakerLite, Option<CustomerLite>)> = Baker::find()
+        .find_also_linked(baker::BakedForCustomer)
+        .select_only()
+        .column_as_prefixed(baker::Column::Name, "A_", baker::Column::Name)
```
tyt2y3, line 668: Actually, we can make the `column_as` 2nd parameter accept an `IntoIdentity`, then we won't need to introduce new API?

billy1624: Done

PR #106, opened 2021-08-25
Example of integrating with Rocket

tyt2y3:


tyt2y3:
What problems are you facing?

billy1624:
Hey @samsamai, can you push the latest draft up here? Even it can't be compiled.

samsamai:
Hi @tyt2y3 @billy1624 , yeah I've been trying to clean up the code it tonight but I'm having some weird problems, it is giving errors that I wasn't getting before. I just rebooted and updated rust... 🤞

samsamai:
Still doing it, it's weird, I'm going back to one of my commits from before but getting lots of these:
```
error[E0425]: cannot find function, tuple struct or tuple variant `ColumnFromStrErr` in module `sea_orm`
  --> /Users/sam/Code/github/sea-orm/src/tests_cfg/fruit.rs:20:40
   |
20 | #[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
   |                                        ^^^^^^^^^^^^ not found in `sea_orm`
   |
   = note: this error originates in the derive macro `DeriveColumn` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Not sure what it's doing with test_cfg when my code is a crate inside examples

billy1624:
Try `cargo update`

samsamai:
Thanks! that was it!

samsamai:
@billy1624 I pushed up the version that I have, sorry it is really messy as I was trying everything to just get it to work. 
It needs this repo to run: `git@github.com:samsamai/Rocket.git`, branch: ` ss/seaorm-contrib` so if you pull that down locally and change the `examples/rocket_example/Cargo.toml` and also `contrib/db_pools/lib/Cargo.toml` in this rocket repo, it should work locally.

`cargo run` in the rocket_example dir, and then `http://127.0.0.1:8000/sqlx`

It's hard coded to use mysql (see the rocket side) just to get it working.

tyt2y3:
Okay, so that means we need to open a PR on Rocket!

billy1624:
Hey @samsamai, I think we can `impl rocket_db_pools::Pool for sea_orm:: Database` inside `sea-orm` without the need of forking `rocket`?

https://github.com/SeaQL/sea-orm/blob/87d4f09931df6210e8745a2235ac1cc86303c053/src/driver/rocket_mysql.rs#L1-L39

tyt2y3:
Looks good so far. I am just puzzling, what's prevent us from generalizing over MySQL and Postgres?

samsamai:
We should be able to, I'll have to fix the implementation of the rocket driver first, right now it is not using a pool and just had mysql hardcoded in so we could get things working.

tyt2y3:
okay, thank you for the explanation!

samsamai:
@tyt2y3 Rocket expects us to implement `rocket_db_pools::Pool` for the struct that we have chosen to derive database on.
This impl needs an `init` where the database pool is initialized with various options, and a `get` which returns a connection from the pool.

billy1624:
Any other problem we need to solve? @samsamai

samsamai:
I think it would help with the Rocket integration if there was a general Sea-Orm pool type, which would generalize over all the different DB types. Currently we have `SqlxMySqlPoolConnection`, `SqlxPostgresPoolConnection` and `SqlxSqlitePoolConnection` but I can't see anything like a `SeaOrmPool` that generalizes over them.

If there was such a thing then the rocket driver would look something like this in pseudo:

```
#[rocket::async_trait]
impl rocket_db_pools::Pool for sea_orm::SeaOrmPool {
    type Error = sea_orm::DbErr;

    type Connection = sea_orm::DatabaseConnection;

    async fn init(figment: &Figment) -> Result<Self, Self::Error> {
      // let pool = Connect to database via url
      // return Ok(pool)
    }

    async fn get(&self) -> Result<Self::Connection, Self::Error> {
        // return a connection from the pool
        // return Ok(self.pool.acquire().await)
    }
}
```

tyt2y3:
okay, I think we can name it `sea_orm::RocketDbPool`
@samsamai would you need help from @billy1624 on this?

samsamai:
@tyt2y3 Yes sure, @billy1624 help is always appreciated. I'll make a start on it and ask if I get stuck.

examples/rocket_example/Rocket.toml
```diff
@@ -1,5 +1,9 @@
...
 
-[global.databases]
-blog = { url = "sqlite::memory:" }
+# Mysql
+# make sure to enable "sqlx-mysql" feature in Cargo.toml, i.e default = ["sqlx-mysql"]
+# url = "mysql://root:@localhost/rocket_example"
+
+# Postgres
+# make sure to enable "sqlx-postgres" feature in Cargo.toml, i.e default = ["sqlx-postgres"]
+url = "postgres://root:root@localhost/rocket_example"
```
tyt2y3, line 9: Just a tiny tiny thing. We should comment out this line too?

tyt2y3:
Looks really good now!
@samsamai is there still anything missing?
@billy1624 what's your thought?

examples/rocket_example/src/sqlx/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[cfg(feature = "sqlx-mysql")]
+    let db_url = "mysql://root:@localhost/rocket_example";
+    #[cfg(feature = "sqlx-postgres")]
+    let db_url = "postgres://root:root@localhost/rocket_example";
+
+    let conn = sea_orm::Database::connect(db_url).await.unwrap();
+
+    let _create_post_table = setup::create_post_table(&conn).await;
+    Ok(rocket)
+}
```
billy1624, line 92: Can we retrieve the db_url defined in `examples/rocket_example/Rocket.toml` from `Rocket<Build>`?

billy1624: I guess we can get the connection directly?
https://github.com/SergioBenitez/Rocket/blob/8cae077ba1d54b92cdef3e171a730b819d5eeb8e/examples/databases/src/sqlx.rs#L71-L82

tyt2y3: Yes, seems so

samsamai:
I've been adding some templates so that the example renders html and is more interactive. I will check that in soon and then have a look at getting db_url in the run_migrations.

billy1624:
Wow we got an UI for it? Cool! @samsamai

examples/rocket_example/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+        },
+    )
+}
+
+#[get("/<id>")]
+async fn read(conn: Connection<Db>, id: i64) -> Option<Json<post::Model>> {
+    let post: Option<post::Model> = Post::find_by_id(id)
+        .one(&conn)
+        .await
+        .expect("could not find post");
```
tyt2y3, line 73: Will this panic be handled by rocket or it will halt the entire server?
Otherwise, if the post is not found, we should return a `404`.

examples/rocket_example/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+async fn delete(conn: Connection<Db>, id: i32) -> Flash<Redirect> {
+    let post: post::ActiveModel = Post::find_by_id(id)
+        .one(&conn)
+        .await
+        .unwrap()
+        .unwrap()
+        .into();
+    let _result = post.delete(&conn).await.unwrap();
+
+    Flash::success(Redirect::to("/"), "Post successfully deleted.")
```
tyt2y3, line 91: Similarly, I think we should handle not found and return 404 here.

examples/rocket_example/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    let post = post_form.into_inner();
+
+    let _post = post::ActiveModel {
+        title: Set(post.title.to_owned()),
+        text: Set(post.text.to_owned()),
+        ..Default::default()
+    }
+    .save(&conn)
+    .await
+    .expect("could not insert post");
```
tyt2y3, line 44: Similarly, we should instead return 500 internal server error.

examples/rocket_example/src/main.rs
tyt2y3: Oh I tried, it is 404 indeed.

tyt2y3:
I think we are missing a link to /read the post?

samsamai:
@tyt2y3 I added an edit page which serves dual purpose.

samsamai:
This is ready to launch, I reckon. 😉

tyt2y3:
image

PR #112, opened 2021-08-27
Add dynamic last insert id for postgres

Acidic9:
This adds support for issue https://github.com/SeaQL/sea-orm/issues/103

I'm sure you guys may want to re-write this from scratch.. but this is my take on it.
I implemented it in my own forked version, and it may not be completely compatible with SeaORM's official codebase, but I'm submitting it as a PR in case you guys want to take a look at my implementation.

Feel free to close this PR, I'll leave it open just to make sure it gets seen. This code works for my personal project.

Edit:
I noticed there's some other code included in this PR not related to last insert ID. Though a majority of the code is for implementing it. This PR is just a guide in case any devs want to reference it when implementing it in SeaORM.

See commits:
https://github.com/SeaQL/sea-orm/pull/112/commits/d300174dff33de831a913534f70d29dcd8ab17f8
https://github.com/SeaQL/sea-orm/pull/112/commits/58d3db926b19f5038ea823e451e5f8f9cfc394d4
For the main code changes made to make this work.

PR #117, opened 2021-08-29
`InsertResult` to return the primary key's type

### LINKED ISSUE #103: `InsertResult` to return the primary key's type
Currently, the `last_insert_id` from InsertResult simply uses the ExecResult's last_insert_id.
However, at least for Postgres with `returning`, we should be able to retrieve the primary key column for the Entity and return the value in its native type. For example, a `uuid` instead of an `i64`.

billy1624:
Resolve #103

Special thanks to @Acidic9 for providing inspiration on PR #112

### DIFF
--- a/src/executor/execute.rs
+++ b/src/executor/execute.rs
@@ -1,6 +1,3 @@
-use crate::TryGetable;
-use std::str::FromStr;
-
 #[derive(Debug)]
 pub struct ExecResult {
     pub(crate) result: ExecResultHolder,
@@ -21,36 +18,23 @@ pub(crate) enum ExecResultHolder {
 // ExecResult //
 
 impl ExecResult {
-    pub fn last_insert_id<T>(&self) -> T
-    where
-        T: TryGetable + Default + FromStr,
-    {
+    pub fn last_insert_id(&self) -> u64 {
         match &self.result {
             #[cfg(feature = "sqlx-mysql")]
-            ExecResultHolder::SqlxMySql(result) => result
-                .last_insert_id()
-                .to_string()
-                .parse()
-                .unwrap_or_default(),
+            ExecResultHolder::SqlxMySql(result) => result.last_insert_id(),
             #[cfg(feature = "sqlx-postgres")]
-            ExecResultHolder::SqlxPostgres(result) => {
-                res.try_get("", "last_insert_id").unwrap_or_default()
-            }
+            ExecResultHolder::SqlxPostgres(_) => panic!("Should not retrieve last_insert_id this way"),
             #[cfg(feature = "sqlx-sqlite")]
             ExecResultHolder::SqlxSqlite(result) => {
                 let last_insert_rowid = result.last_insert_rowid();
                 if last_insert_rowid < 0 {
                     panic!("negative last_insert_rowid")
                 } else {
-                    last_insert_rowid.to_string().parse().unwrap_or_default()
+                    last_insert_rowid as u64
                 }
             }
             #[cfg(feature = "mock")]
-            ExecResultHolder::Mock(result) => result
-                .last_insert_id
-                .to_string()
-                .parse()
-                .unwrap_or_default(),
+            ExecResultHolder::Mock(result) => result.last_insert_id,
         }
     }
 
--- a/src/executor/execute.rs
+++ b/src/executor/execute.rs
@@ -23,7 +23,9 @@ impl ExecResult {
             #[cfg(feature = "sqlx-mysql")]
             ExecResultHolder::SqlxMySql(result) => result.last_insert_id(),
             #[cfg(feature = "sqlx-postgres")]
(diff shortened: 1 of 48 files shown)


src/executor/execute.rs
```diff
@@ -21,23 +21,36 @@ pub(crate) enum ExecResultHolder {
 // ExecResult //
 
 impl ExecResult {
```
tyt2y3, line 23: Is there a way we can keep `ExecResult` as is? It is just a wrapper over SQLx's QueryResult. We should not be changing ExecResult. We should only change InsertResult.

billy1624: Let me think...

src/executor/insert.rs
```diff
@@ -83,15 +83,21 @@ where
...
             let res = conn.query_one(statement).await?.unwrap();
-            crate::query_result_into_exec_result(res)?
+            res.try_get("", "last_insert_id").unwrap_or_default()
         }
-        _ => db.execute(statement).await?,
+        _ => {
+            db.execute(statement).await?
+            .last_insert_id()
+            .to_string()
+            .parse()
```
tyt2y3, line 96: Yeah this looks much more sensible. May I also know why is it necessary to `to_string` and `parse` again?

billy1624: Because we cannot convert e.g. `u64` to `i32` directly

billy1624:
If this passes all tests, I will start re-think about all these trait bounds

https://github.com/SeaQL/sea-orm/blob/4e5a8d2dc553a26668c89b5740bb6a0d4b5e33f3/src/entity/primary_key.rs#L10-L18

src/executor/insert.rs
tyt2y3: how about this https://stackoverflow.com/questions/28273169/how-do-i-convert-between-numeric-types-safely-and-idiomatically?

billy1624:
Why I got this error again?
https://github.com/SeaQL/sea-orm/runs/3454855146#step:6:104

src/executor/insert.rs
billy1624: I saw this StackOverflow this morning. I don't think it helps :(

tyt2y3: Why not?

billy1624: We can't use `TryFrom` or `From` for it

tyt2y3: Why not?

billy1624: If we want `TryFrom` or `From` `u32` to `ValueType`, this requires `ValueType: From<u32>`. Which is not reasonable

tyt2y3: But having to stringify and parse is not much more sensible either

billy1624:
> Why I got this error again?
> https://github.com/SeaQL/sea-orm/runs/3454985766?check_suite_focus=true#step:6:104

Okay, solve it on 4d0c5f0

src/executor/insert.rs
billy1624: Agree

tyt2y3:
We need a trait `TryFromU64` to replace the use of `FromStr`

Acidic9:
> We need a trait `TryFromU64` to replace the use of `FromStr`

How would TryFromU64 work with something like UUID? I don't think postgres returns UUID as an integer does it?

tyt2y3:
I think the Postgres side will fairly ignore that result. We were only trying to fix the MySQL side.

billy1624:
Wow the commit hash starts with 0e0ee0e lolll

Acidic9:
I'm curious why the trait TryFromU64 is added instead of using `From<u64>` 🤔

billy1624:
> I'm curious why the trait `TryFromU64` is added instead of using `From<u64>` or `TryFrom<u64>` 🤔
> It seems like this might cover cases for libraries with types that already implement `From<u64>` / `TryFrom<u64>`
> Sorry if this has already been discussed

I think what we have here is not an usual logic of `From<u64>` or `TryFrom<u64>`. So, we better not use the conventional std traits.

src/executor/query.rs
```diff
@@ -280,3 +288,135 @@ impl TryGetable for Decimal {
...
+            cols.len()
+        ))))
+    } else {
+        Ok(())
+    }
+}
+
+// TryFromU64 //
+
+pub trait TryFromU64: Sized {
```
tyt2y3, line 349: Awesome!

src/tests_cfg/cake_filling_price.rs
```diff
@@ -35,6 +35,8 @@ pub enum PrimaryKey {
 }
 
 impl PrimaryKeyTrait for PrimaryKey {
+    type ValueType = (i32, i32);
```
tyt2y3, line 38: Seems like we are able to support composite key now!

billy1624: Indeed

PR #128, opened 2021-09-05
EntityModel derive-macro

nappa85:


tyt2y3:
This is brilliant! It's much shorter than I thought it needed to be.
Btw, would you mind moving the code from `lib.rs` to a new file under the `derives` module, like the other macros?

nappa85:
> This is brilliant! It's much shorter than I thought it needed to be.
> Btw, would you mind moving the code from `lib.rs` to a new file under the `derives` module, like the other macros?

Yes, this is a mere copy-paste of my code, only with some renaming in the process, I was sure it would have to be split like the other macros, but yesterday I didn't had time to put on it

tyt2y3:
Sure, no problem. I think we'd be pretty close after this PR!

nappa85:
I've added support for indexed and unique attributed, I stopped on default_value because of the new method it relies on

nappa85:
I've made an hypothesis on default_value and default_expr implementation, I'm not sure it's what you intended

tyt2y3:
> I've made an hypothesis on default_value and default_expr implementation, I'm not sure it's what you intended

Yes they are, accurately.

Thank you very much for this. 

For the scope of this PR, the only thing I'd ask for is probably the optional impl of 'schema_name' besides 'table_name' in the EntityTrait.

nappa85:
I've seen the conflicts, I can update the macro to auto-fill PrimaryKeyTrait::ValueType with primary key type(s)

PR #130, opened 2021-09-06
Remove generics on find_by_id

billy1624:


tyt2y3:
I think we should not remove the panics, there is still a possibility for it to happen, though the cause would ONLY be a wrongly defined Entity.

billy1624:
Panics that we can't get rid of lolll

tyt2y3:
Oh, unless we have a derive macro to return the number of enum variants at compile time, then match it against the arity of the type def at compile time!
How does it sound?

tyt2y3:
Not for this PR may be

PR #132, opened 2021-09-07
Add ActiveModel insert & update

### LINKED ISSUE #110: Implement `insert` & `update` for `ActiveModel`
To avoid manually `exec` the insert `ActiveModel`. In below, we cannot simply `save` it because this is a junction table and the composite primary key should be set during insert.

https://github.com/SeaQL/sea-orm/blob/bfaa7d4539afc033cb3f240fa4fed4ff1344c7c0/tests/relational_tests.rs#L562-L569

Perhaps `impl` it here

https://github.com/SeaQL/sea-orm/blob/8cfa14233d2d6bce27857cce8b12a95be414d2cc/sea-orm-macros/src/derives/active_model.rs#L39-L47

### LINKED ISSUE #129: Remove implicit filter in update method
TLDR;

Instead of using the `Set` primary key in an active model for updating an entity, use a manual `.filter` method.
```rust
Fruit::update(pear)
    .filter(fruit::Column::Id.eq(123)) // Explicitly filter for the ID only
    .exec(db).await?;
```

And add a new method directly to an `ActiveModel` which uses the `Set` primary key.
```rust
let pear = fruit::ActiveModel { id: Set(123), ... };
pear.update().exec(db).await?; // No need for `.filter`, it uses the ID already
```

---

Currently to perform an update on multiple fields, it looks like you do it with:
```rust
Fruit::update_many()
    .col_expr(fruit::Column::CakeId, Expr::value(Value::Null))
    .filter(fruit::Column::Name.contains("Apple"))
    .exec(db)
    .await?;
```
But if you want to update 10 columns, do you need to manually add `.col_expr` for each column?

And doing the following will update the pears name with the ID `123`.
```rust
let pear = pear::ActiveModel { 
    id: Set(123),
    name: Set("Green Pear".to_string()),
};
Fruit::update(pear).exec(db).await?;
```
The issue with this, is that if the id is not set, a panic will occur.
https://github.com/SeaQL/sea-orm/blob/de1f5d397df26659ef11ee9cc97251964503c748/src/query/update.rs#L89-97

But perhaps it'd make sense for the API to remove the lines 89 - 97 (in the link above) and force the user to explicitly add the `WHERE id = '...'` clause:
```rust
Fruit::update(pear)
    .filter(fruit::Column::Id.eq(123))
    .exec(db).await?;
```
The benefit of this, is it allows users to write more complex filtering rather than a simple `id = ...`.
The down side is that if the ActiveModel has the ID as `Set`... then it'll actually try to update the ID in the query. But I don't consider this so much as a down side, because some people may actually want to update a primary key.

The active model could have an `.update` method which could keep the current behaviour and use the set primary key to perform an update.
Eg:
```rust
let pear = pear::ActiveModel { ... };
pear.update(db).await?; // Update the pear in place using it's set ID.
```

billy1624:
Resolve #110 #127

### DIFF
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -36,24 +36,6 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
             #(pub #field: sea_orm::ActiveValue<#ty>),*
         }
 
-        impl ActiveModel {
-            pub async fn insert(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
-                sea_orm::insert_and_select_active_model::<Self, Entity>(self, db).await
-            }
-
-            pub async fn update(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
-                sea_orm::update_active_model::<Self, Entity>(self, db).await
-            }
-
-            pub async fn save(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
-                sea_orm::save_active_model::<Self, Entity>(self, db).await
-            }
-
-            pub async fn delete(self, db: &sea_orm::DatabaseConnection) -> Result<sea_orm::DeleteResult, sea_orm::DbErr> {
-                sea_orm::delete_active_model::<Self, Entity>(self, db).await
-            }
-        }
-
         impl std::default::Default for ActiveModel {
             fn default() -> Self {
                 <Self as sea_orm::ActiveModelBehavior>::new()
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -37,6 +37,14 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
         }
 
         impl ActiveModel {
+            pub async fn insert(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
+                sea_orm::insert_and_select_active_model::<Self, Entity>(self, db).await
+            }
+
+            pub async fn update(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
+                sea_orm::update_active_model::<Self, Entity>(self, db).await
+            }
+
             pub async fn save(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
                 sea_orm::save_active_model::<Self, Entity>(self, db).await
             }
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -212,7 +212,10 @@ where
     Ok(am)
 }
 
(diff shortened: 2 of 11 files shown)


sea-orm-macros/src/derives/active_model.rs
```diff
@@ -36,16 +36,6 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
...
 
-        impl ActiveModel {
-            pub async fn save(self, db: &sea_orm::DatabaseConnection) -> Result<Self, sea_orm::DbErr> {
-                sea_orm::save_active_model::<Self, Entity>(self, db).await
-            }
-
-            pub async fn delete(self, db: &sea_orm::DatabaseConnection) -> Result<sea_orm::DeleteResult, sea_orm::DbErr> {
-                sea_orm::delete_active_model::<Self, Entity>(self, db).await
-            }
-        }
```
billy1624, line 47: We finally get rid of these codes

tyt2y3: Nice

tyt2y3 (commented):
Yes. Please withheld the breaking change to the `Updater` API in this PR. Thanks

billy1624:
I should open another PR for the `async_trait` thing?

src/entity/model.rs
```diff
@@ -2,7 +2,7 @@ use crate::{DbErr, EntityTrait, Linked, QueryFilter, QueryResult, Related, Selec
 pub use sea_query::Value;
 use std::fmt::Debug;
 
-pub trait ModelTrait: Clone + Debug {
```
tyt2y3, line 5: Oh we needed to add `Send` to `ModelTrait`

billy1624: Yes, this this what I'm taking

src/entity/active_model.rs
```diff
@@ -65,9 +67,71 @@ pub trait ActiveModelTrait: Clone + Debug {
...
+        let exec = <Self::Entity as EntityTrait>::insert(am).exec(db);
+        let res = exec.await?;
+        // TODO: if the entity does not have auto increment primary key, then last_insert_id is a wrong value
+        // FIXME: Assumed valid last_insert_id is not equals to Default::default()
+        if <<Self::Entity as EntityTrait>::PrimaryKey as PrimaryKeyTrait>::auto_increment()
+            && res.last_insert_id != <<Self::Entity as EntityTrait>::PrimaryKey as PrimaryKeyTrait>::ValueType::default()
+        {
+            let find = <Self::Entity as EntityTrait>::find_by_id(res.last_insert_id).one(db);
+            let found = find.await;
+            let model: Option<<Self::Entity as EntityTrait>::Model> = found?;
```
billy1624, line 84: We need to add `ModelTrait: Sync` because of this

tyt2y3: 🤔 `Send` is fair I guess. `Sync` would be too harsh.

tyt2y3:
> I should open another PR for the `async_trait` thing?

Keep this PR as-is is fine.

src/entity/active_model.rs
billy1624: `Send` I mean loll It's a typo

tyt2y3:
I rebased on master

billy1624:
Looks good to me!

tyt2y3 (approved):
smart one

Issue #105, opened 2021-08-24
More compact Entity format


nappa85:
I was trying your crate, seems promising, but the way you choose to represent models is too verbose.
For example I've a DB table with 53 fields, using your system I've written like 210 LOC, repeating the same informations like 3 times.

I think everything could be done with a single proc_macro, that takes a standard Rust struct and replaces it with a module with the same name of the struct, with Model struct with more or less the same contents of the original struct, Column enum with structs fields as variants, automatic impl ColumnTrait for Column with a type mapping, where the mapping fails you can use a field notation to override it (comes handy for custom types), table name and primary key can be specified with a struct notation, what's missing?
Relations, well, this have to be specified externally, but you can point the enum with another notation, and if the notation isn't present you generate a void enum.

After that, if I can add something more, I often use Cow on my entities, because this way I'm not forced to clone Strings when updating/creating. With your system I can't, because Model doesn't support a lifetime. It's not a big problem, but it blocks also other custom types with a lifetime

Acidic9:
I haven't thought of this before, but I think I can agree with you somewhat.

For example, `Entity`, `Model`, `Column`, `PrimaryKey`, `PrimaryKeyTrait` & `ColumnTrait` could be combined into a single struct definition with derives.
From:
```rust
#[derive(Copy, Clone, Default, Debug, DeriveEntity)]
pub struct Entity;

impl EntityName for Entity {
    fn schema_name(&self) -> Option<&str> {
        Some("public")
    }

    fn table_name(&self) -> &str {
        "users"
    }
}

#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel)]
pub struct Model {
    pub id: Uuid,
    pub created_at: DateTimeWithTimeZone,
    pub updated_at: DateTimeWithTimeZone,
    pub username: String,
    pub password: String,
    pub email: Option<String>,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
pub enum Column {
    Id,
    CreatedAt,
    UpdatedAt,
    Username,
    Password,
    Email,
}


#[derive(Copy, std::clone::Clone, std::fmt::Debug, EnumIter, DerivePrimaryKey)]
pub enum PrimaryKey {
    Id,
}

impl PrimaryKeyTrait for PrimaryKey {
    fn auto_increment() -> bool {
        false
    }
}

#[derive(Copy, Clone, Debug, EnumIter)]
pub enum Relation {}

[...]

PR #135, opened 2021-09-07
Entity format

tyt2y3:
#105

### DIFF
--- a/sea-orm-macros/src/derives/entity_model.rs
+++ b/sea-orm-macros/src/derives/entity_model.rs
@@ -100,20 +100,8 @@ pub fn expand_derive_entity_model(data: Data, attrs: Vec<Attribute>) -> syn::Res
                                                     ));
                                                 }
                                             } else if name == "auto_increment" {
-                                                if let Lit::Str(litstr) = &nv.lit {
-                                                    auto_increment = match litstr.value().as_str() {
-                                                        "true" => true,
-                                                        "false" => false,
-                                                        _ => {
-                                                            return Err(Error::new(
-                                                                field.span(),
-                                                                format!(
-                                                                    "Invalid auto_increment = {}",
-                                                                    litstr.value()
-                                                                ),
-                                                            ))
-                                                        }
-                                                    };
+                                                if let Lit::Bool(litbool) = &nv.lit {
+                                                    auto_increment = litbool.value();
                                                 } else {
                                                     return Err(Error::new(
                                                         field.span(),
--- a/sea-orm-macros/src/derives/entity_model.rs
+++ b/sea-orm-macros/src/derives/entity_model.rs
@@ -1,0 +1,237 @@
+
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::quote;
+use syn::{Attribute, Data, Fields, Lit, Meta, parse::Error, punctuated::Punctuated, spanned::Spanned, token::Comma};
+
+use convert_case::{Case, Casing};
+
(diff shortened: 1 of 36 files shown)


Acidic9:
Looking awesome!

We don't want to use the shorter `#[sea(...)]` instead of `#[sea_orm(...)]`?

Also, if we wanted this to be more correct (something I even forgot to do), we should add `::` before the types. For example `impl ::std::default::Default ... ` and `impl ::sea_orm::...`, but it's not super important.

Do you think it's worth updating all the macros to take in an `DeriveInput` parameter and return `syn::Result<TokenStream>`, and use the struct to store the data?
Now it has a few inconsistencies, and updating it like that would allow a macro_rules for simplifying the lib.rs file.
Eg:
```rust
macro_rules! create_derive {
    ($name:ident $(, $( $attrs:ident ),+)?) => {
        paste! {
            #[proc_macro_derive($name, attributes($( $( $attrs ),*)?))]
            pub fn [< $name:snake >](input: TokenStream) -> TokenStream {
                let input = parse_macro_input!(input as DeriveInput);
                derives::[<expand_ $name:snake>](input)
                    .unwrap_or_else(Error::into_compile_error)
                    .into()
            }
        }
    }
}

create_derive!(DeriveEntity, sea);
create_derive!(DerivePrimaryKey);
create_derive!(DeriveColumn);
create_derive!(DeriveCustomColumn);
create_derive!(DeriveModel, sea);
create_derive!(DeriveModelColumn, sea);
create_derive!(DeriveModelPrimaryKey, sea);
create_derive!(DeriveActiveModel);
create_derive!(DeriveActiveModelBehavior);
create_derive!(DeriveRelation, sea);
create_derive!(FromQueryResult);
```
*(the macro above works)*

tyt2y3:
> We don't want to use the shorter `#[sea(...)]` instead of `#[sea_orm(...)]`?
> 
I guess the norm is to use the crate's name itself

> Also, if we wanted this to be more correct (something I even forgot to do), we should add `::` before the types. For example `impl ::std::default::Default ... ` and `impl ::sea_orm::...`, but it's not super important.
> 
I think it's fine if someone wants to alias something else as sea_orm.

> Do you think it's worth updating all the macros to take in an `DeriveInput` parameter and return `syn::Result<TokenStream>`, and use the struct to store the data?
> 
> Now it has a few inconsistencies, and updating it like that would allow a macro_rules for simplifying the lib.rs file.

Yes I do think it's a good idea.

Acidic9:
> I guess the norm is to use the crate's name itself?

I think it's fine having a shorter version, nothing will conflict with `sea`. For example, the async-graphql crate uses `#[graphql(...)]`, instead of the full `async-graphql`.

> I think it's fine if someone wants to alias something else as sea_orm.

Sure, it's a very small thing. But I think it's usually a standard thing when writing macros, mainly for `::std::...` std library types.

tyt2y3:
I hope `sea_orm` is not too long to be obstructive )
I do agree on the `::std` thing, but we actually `use create as sea_orm` ourselves
https://github.com/SeaQL/sea-orm/blob/a186338ff5794553bdd0b53fdcb4310deeece0a5/src/tests_cfg/fruit.rs#L1

sea-orm-macros/src/derives/entity_model.rs
```diff
@@ -0,0 +1,247 @@
...
+            match self {
+                #columns_trait
+            }
+        }
+    }
+
+    #entity_def
+
+    #primary_key
+        });
```
billy1624, line 246: Btw... there are some clippy warning here loll

billy1624 (changes requested):

billy1624 (approved):

PR #137, opened 2021-09-09
Add model validation with database

Acidic9:
Checks every struct that derives `DeriveModel` with the matching database table ensuring the fields are valid.
This is done through a feature flag `"model-validation"`.

Provides warnings and errors for mismatches in the models.

Preview of warnings

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ default = [
     "with-uuid",
 ]
 macros = ["sea-orm-macros"]
+model-validation = ["sea-orm-macros/model-validation"]
 mock = []
 with-json = ["serde_json", "sea-query/with-json"]
 with-chrono = ["chrono", "sea-query/with-chrono"]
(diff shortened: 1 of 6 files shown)


tyt2y3:
Interesting. I guess not everyone would like to have it work at compile time?
I am thinking of adding a `validate` command to `sea-orm-cli` instead

Acidic9:
Yes that would be nice too.. but I don't know how to scan all the models and their fields in the source code.

The only way I can think to do that is to generate json data of the project with rustdoc, but that can only be done in nightly as of now.

Acidic9:
This feature is not wanted?

tyt2y3:
I most certainly closed this accidentally.

But it did seem to be a bit ambitious in one attempt.

Is it possible for this feature to be vended by a separate crate?

tyt2y3:
I think a `schema validation` functionality is definitely a great feature.
But I was thinking of doing it 'start-time' vs 'compile-time',
meaning we'd probably add a line `Schema::check()` (something like that) in the entry point of the program, before the web server starts.

Can you share more on the benefits of doing this compile time?

Acidic9:
Hm interesting point. Maybe an external package for this might be a good approach actually to avoid SeaORM from getting bloated. I'll take some time to have a think about this. `Schema::check()` also sounds nice!

PR #143, opened 2021-09-10
Codegen for compact Entity format

### LINKED ISSUE #138: Codegen for compact Entity format
I think it's better for us to preserve the functionality to generate the current long Entity format, and can be switched on with a command line arguemnt.

- [x] update codegen
- [x] update sea-orm-cli to generate short format by default
- [x] update docs

billy1624:
Resolve #138


billy1624:
We should be able to generate correct `ForeignKeyAction` for each relation given that we have it now.

billy1624:
Hey @tyt2y3, see the generated compact entity files here bd6e5f8

tyt2y3:
Can't comment there. Why the contact_details is string instead of json?

billy1624:
> Can't comment there. Why the contact_details is string instead of json?

See it here #146 

Fixed the json thing on SeaQL/sea-schema@e7f6a37

tyt2y3:
How about the `Decimal` type?

```
#[sea_orm(column_type = "Decimal(Some((19, 4)))")]
```

without this, the ColumnDef won't compile

billy1624:
Can we provide a default column type for Decimal?

billy1624:
Also, should we fully quantify custom types? E.g. Decimal, Json

tyt2y3:
The problem is the Decimal type does not work without specifying the decimal places

tyt2y3:
> Also, should we fully quantify custom types? E.g. Decimal, Json

Basically yes

billy1624:
> The problem is the Decimal type does not work without specifying the decimal places

We can provide a default for it if it can't be discovered? Let say Decimal((10,2))

tyt2y3:
> > The problem is the Decimal type does not work without specifying the decimal places
> 
> We can provide a default for it if it can't be discovered? Let say Decimal((10,2)) as default?

I'd rather let it not compile

billy1624:
> Looks good to me! Anything missing?

Docs to Add / Edit

1. Document `compact` / `expanded` options for `sea-orm-cli generate entity`
2. Introduce compact entity syntax & `DeriveEntityModel` macro attributes

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -297,25 +320,83 @@ impl EntityWriter {
...
+            .iter()
+            .map(|col| {
+                let mut attrs: Punctuated<_, Comma> = Punctuated::new();
+                if primary_keys.contains(&col.name) {
+                    attrs.push(quote! { primary_key });
+                    if !col.auto_increment {
+                        attrs.push(quote! { auto_increment = false });
+                    }
+                }
+                if let Some(ts) = col.get_col_type_attrs() {
```
tyt2y3, line 344: We forgot one thing. That is, `nullable` when a col_type is involved.
See https://github.com/SeaQL/sea-orm/issues/105#issuecomment-915922635

billy1624: Oh, will fix this

Issue #142, opened 2021-09-10
Transactions support


nappa85:
Coming from https://github.com/SeaQL/sea-orm/pull/118, I start a new issue con go on with the analisys:

What @tyt2y3 proposed:
> I think transaction is highly driver specific. I don't think we can cleanly wrap and generalizes over different database+driver.
> Perhaps we can start by exposing some of the underlying types and methods from SQLx.
> 
> https://docs.rs/sqlx/0.5.7/sqlx/trait.Connection.html#method.transaction

It can be a starting point, if we're able to maintain sea-orm typing.
What I mean is, sqlx Connection::transaction takes an fn(&mut Transaction), we would need to wrap it up so inside the function we can still use our DatabaseConnection type.

And it's kind of an overkill IMHO, Transaction is only a sugar wrapper over a normal connection, where a START TRANSACTION has been executed.

I would prefere an approach like this: https://docs.rs/mysql_async/0.28.0/mysql_async/struct.Conn.html#method.start_transaction
where Transaction is simply a tiny wrapper over Connection and can be used in any method that takes `&Conn` because it `impl Deref<Conn>`

About isolation levels, mysql and postgres are quite similars:
https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html
https://www.postgresql.org/docs/9.5/sql-set-transaction.html

Sqlite has transactions too, it has a little different dialect, like BEGIN [TRANSACTION] instead of START TRANSACTION, and isolation levels have to be set using the PRAGMA keyword, like https://www.sqlite.org/pragma.html#pragma_read_uncommitted

tyt2y3:
There is a reason that SQLx takes in a `FnOnce` for transaction. It is using a connection pool, this is to prevent that the user forget to COMMIT or ROLLBACK thereafter.

I think, probably, the SeaORM API can take a [`Vec<Statement>`](https://docs.rs/sea-orm/0.x/sea_orm/struct.Statement.html) and execute them in a single transaction. This is rather simplistic, and does not cover all use cases.

All in all, I think transaction in an async context is a complex matters that requires more investigation.

nappa85:
Transactions are often used to lock to a certain state (isolation) to be sure that there are not changes between select queries. How do you cover that case with a `Vec<Statement>`?

I can try to see if I can fit inside sqlx apis with a wrapper, if it covers all scenarios

tyt2y3:
[...]

PR #149, opened 2021-09-11
Transaction support

nappa85:
As discussed on https://github.com/SeaQL/sea-orm/issues/142,
I've implemented transactions adding a DbConnection trait.

As always, I'm open to discussion, this time more that before since I've touched a lot of core things around


nappa85:
It has been a long hard way out of hell, but I think I finished this PR

tyt2y3:
Thank you so much for the hard work! I am trying to go through the extensive change.

Several questions popping in my head:

1) How'd it be like to actually perform a transaction?
2) Why is the `Mutex` needed?
3) Why is the `&*conn` needed in Rocket example?

nappa85:
>     1. How'd it be like to actually perform a transaction?
It should work like sqlx::Connection::transaction, I'll write a test ASAP
>     2. Why is the `Mutex` needed?
The Mutex is needed because `sqlx::Executor` is implemented only for `&mut sqlx::Connection`
>     3. Why is the `&*conn` needed in Rocket example?
Didn't understood deeply myself that one, `Connection` impl `Deref<DatabaseConnection>` but it seems that the compiler isn't able to match that with the generic.
Haven't thought about that too much to be honest

nappa85:
Had to move to `Pin<Box<dyn Future>>` to get out of a lifetime hell...

tyt2y3:
Thank you so much for the expedition! I think, lifetime + async is probably 'fighting two Ferris' at once.

The only things left to discuss, is probably the namings )

I am thinking of renaming `DbConnection` to `ConnectionTrait` or `DatabaseFunctions`.

@billy1624 several followups we should do after merging:
1. remove the `is_mock_connection` checks and still make the test suite pass
2. investigate how to remove the extra `*` from Rocket example `save(&*conn)`
3. enhance our `MockConnection` such that it also 'supports' transaction

nappa85:
> I am thinking of renaming `DbConnection` to `ConnectionTrait` or `DatabaseFunctions`.
Ok, I'll rename to ConnectionTrait ASAP

>     1. remove the `is_mock_connection` checks and still make the test suite pass
If you look at my commits, I've tried, but you'll end-up having the query built conditionally on sqlx-postgres feature, so you'll end up chasing the format, that it isn't what you want to do on a test

tyt2y3:
> If you look at my commits, I've tried, but you'll end-up having the query built conditionally on sqlx-postgres feature, so you'll end up chasing the format, that it isn't what you want to do on a test

I get what you mean. So to clarify the intention, I think `Mock` should behave like Postgres if and only if the `backend` is Postgres, regardless of whether the `sqlx-postgres` feature exists. That means, yes, we should not remove the `is_mock_connection` check, but instead make it independent of the `sqlx-postgres` feature.

nappa85:
> I get what you mean. So to clarify the intention, I think `Mock` should behave like Postgres if and only if the `backend` is Postgres, regardless of whether the `sqlx-postgres` feature exists. That means, yes, we should not remove the `is_mock_connection` check, but instead make it independent of the `sqlx-postgres` feature.

Well, if the query is built "postgres-style" independently from sqlx-postgres feature, the `is_mock_connection` method isn't needed anymore

nappa85:
Sorry, it's monday morning...
I should have done a different branch for the last commit, it's a fix on the proc-macro we made, the Decimal column type was instantiated without the Option-al argument

tyt2y3:
Oh, by the way, you should add `n***85@the***ing.com` to your GitHub account (https://github.com/settings/emails), that way you'd be properly credited!

billy1624:
I think the `MockConnetion` can provide an API like...

```rust
async fn to_json_2() -> Result<(), DbErr> {
    let db = MockDatabase::new(DbBackend::Postgres)
        .append_query_results(vec![vec![maplit::btreemap! {
            "id" => Into::<Value>::into(128), "name" => Into::<Value>::into("apple")
        }]])
        .into_connection();

    db.transaction::<_, (), DbErr>(|db| Box::pin(async move {
        assert_eq!(
            cake::Entity::find().into_json().one(db).await.unwrap(),
            Some(serde_json::json!({
                "id": 128,
                "name": "apple"
            }))
        );

        Ok(())
    })).await.unwrap();

    // With empty transaction log if rollback
    assert_eq!(
        db.into_transaction_log(),
        vec![
            Transaction::from_sql_and_values(
                DbBackend::Postgres,
                r#"INSERT INTO "cake" ("cake"."id", "cake"."name") VALUES ($1, $2)"#,
                vec![128u64.into(), "apple".into()]
            ),
        ]
    );

    Ok(())
}
```

nappa85:
> That's fine.
> Oh, by the way, you should add your commit email `n***85@the***ing.com` to your GitHub account (https://github.com/settings/emails) (or commit using your account's primary email), that way you'd be properly credited!
> Last time I actually rewrote the commit author (https://github.com/SeaQL/sea-orm/commits?author=nappa85)

You're right, that happens because I'm working from my office laptop...

PR #156, opened 2021-09-14
Transaction Support

tyt2y3:
#142 continue


tyt2y3:
Let's get movin @billy1624

nicoulaj:
Maybe you could include `ConnectionTrait` in the prelude ? Btw, `QuerySelect` and `QueryOrder`are also missing (`QueryFilter`, `QueryResult` are included so it feels a bit inconsistent).

tyt2y3:
@nappa85 well, I tried to rebase the branch onto master, which is quite daunting indeed. Let me know if anything is missing.

nappa85:
Seems ok to me, it's quite difficult to spot a difference on this ocean of
changes, but the key points seems to be there

Il giorno sab 18 set 2021 alle ore 17:01 Chris Tsang <
***@***.***> ha scritto:

PR #158, opened 2021-09-15
DatabaseConnection impl Sync, Send and Clone

### LINKED ISSUE #152: `DatabaseConnection` is not `Send`, `Sync` and `Clone`
SeaORM's `DatabaseConnection` type is not `Send`, `Sync` and `Clone`.
These should be easily added as `DatabaseConnection` is a wrapper of the Sqlx pool types, which themselves are `Send`, `Sync` and `Clone`.

On the Sqlx [`Pool`](https://docs.rs/sqlx/0.5.7/sqlx/struct.Pool.html) documentation:
> `Pool` is `Send`, `Sync` and `Clone`, so it should be created once at the start of your application/daemon/web server/etc. and then shared with all tasks throughout its lifetime.

This is usually needed for many apps, such as an actix server, so it can be sent between threads to aquire new connections.

billy1624:
Resolve #152


tyt2y3 (commented):
I think the problem is MockConnection is not intended for cloning,
it would mess up the mock query result or exec result.

tyt2y3 (approved):
Reasonable

PR #164, opened 2021-09-15
Transaction with references test

nappa85:
This test covers the quite common case where an external reference is needed inside a transaction.
I haven't found other solution than using a concrete fn with explicit lifetimes, this test can be used as a working example for users

### DIFF
--- a/tests/query_tests.rs
+++ b/tests/query_tests.rs
@@ -1,6 +1,7 @@
 pub mod common;
 
 pub use common::{bakery_chain::*, setup::*, TestContext};
+use sea_orm::{DatabaseTransaction, DbErr};
 pub use sea_orm::entity::*;
 pub use sea_orm::{QueryFilter, ConnectionTrait};
 
@@ -222,8 +223,6 @@ pub async fn find_all_filter_with_results() {
     feature = "sqlx-postgres"
 ))]
 pub async fn transaction() {
-    use sea_orm::DbErr;
-
     let ctx = TestContext::new("find_all_filter_with_results").await;
 
     ctx.db.transaction::<_, (), DbErr>(|txn| Box::pin(async move {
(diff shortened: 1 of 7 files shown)


nappa85:
Even if the tests doesn't pass for unknown reasons, I'm not still 100% satisfied of this solution, I've a real life situation where this isn't enough and I can't replicate that without porting all the environment.

nappa85:
After an entire day lost fighting this issue, I still haven't a solution.
It's impossible to not use "pinboxed" futures and still describe an HRTB that describes both the Fn and it's return type.
Helper traits aren't a solution because the compiler still doesn't recognizes closures for trait application, and without closures you can't pass references to local variables.
I think in my real-life situation the problem is that I take a reference to a slice in input and my return type references slice contents, so probably cloning I can solve the issue while still using the "pinboxed" version.

tyt2y3:
Thank you for the update. I think your scenario is quite difficult.
Yes, cloning, or taking ownership is always the fallback solution.
Hopefully the overhead is small enough to be negligible.

tyt2y3:
>     I: Iterator<Item=&'a Yadda>,
Why do you want to use a generic here instead of, sharing the ref to yaddayadda?

nappa85:
On the original code it comes from different sources, so I've preferred using the generic instead of converting the formats. But with the generic I can give a lifetime to the type too, so I don't think it's the problem

nappa85:
There is also the "worse" branch with the original format using `&'a str` as hashmap keys instead of `String`, at the moment the only solution I've found is to pass the Vec instead of the slice, move the Vec inside the transaction closure and then return it inside the tuple

tyt2y3:
As far as I know, these kind of things can only be done with an `Arc<Mutex<T>>`

nappa85:
Thank you for your suggestion.
I've made another commit for this PR, I've cleaned-up an unneeded lifetime in transaction method signature

nappa85:
I think tests are failing for a data race, can you take a look at it?

tests/query_tests.rs
```diff
@@ -222,8 +223,6 @@ pub async fn find_all_filter_with_results() {
     feature = "sqlx-postgres"
 ))]
 pub async fn transaction() {
-    use sea_orm::DbErr;
-
     let ctx = TestContext::new("find_all_filter_with_results").await;
```
tyt2y3, line 226: This context should be unique.
May be "transaction_test"?

tests/query_tests.rs
```diff
@@ -255,3 +254,48 @@ pub async fn transaction() {
...
 }
+
+#[sea_orm_macros::test]
+#[cfg(any(
+    feature = "sqlx-mysql",
+    feature = "sqlx-sqlite",
+    feature = "sqlx-postgres"
+))]
+pub async fn transaction_with_reference() {
+    let ctx = TestContext::new("find_all_filter_with_results").await;
```
tyt2y3, line 265: "transaction_with_reference_test"?

PR #167, opened 2021-09-16
Improve working with raw SQL

MuhannadAlrusayni:
(updated) This PR allow us to fetch `T: FromQueryResult` using `T::find_by_statement()` from raw sql:
```rust
// with this PR we can call `T::find_by_statement()` where `T: FromQueryResult` and return `T`
let res: Vec<SelectResult> = SelectResult::find_by_statement(Statement::from_sql_and_values(
    DbBackend::Postgres,
    r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
    vec![],
))
.all(&db)
.await?;

```

### DIFF
--- a/src/database/statement.rs
+++ b/src/database/statement.rs
@@ -1,4 +1,4 @@
-use crate::{DbBackend, FromQueryResult, SelectModel, SelectorRaw};
+use crate::DbBackend;
 use sea_query::{inject_parameters, MysqlQueryBuilder, PostgresQueryBuilder, SqliteQueryBuilder};
 pub use sea_query::{Value, Values};
 use std::fmt;
@@ -42,75 +42,6 @@ impl Statement {
             values: Some(stmt.1),
             db_backend,
         }
-    }
-
-    /// ```
-    /// # #[cfg(feature = "mock")]
-    /// # use sea_orm::{error::*, tests_cfg::*, MockDatabase, Transaction, DbBackend};
-    /// #
-    /// # let db = MockDatabase::new(DbBackend::Postgres)
-    /// #     .append_query_results(vec![vec![
-    /// #         maplit::btreemap! {
-    /// #             "name" => Into::<Value>::into("Chocolate Forest"),
-    /// #             "num_of_cakes" => Into::<Value>::into(1),
-    /// #         },
-    /// #         maplit::btreemap! {
-    /// #             "name" => Into::<Value>::into("New York Cheese"),
-    /// #             "num_of_cakes" => Into::<Value>::into(1),
-    /// #         },
-    /// #     ]])
-    /// #     .into_connection();
-    /// #
-    /// use sea_orm::{entity::*, query::*, tests_cfg::cake, FromQueryResult};
-    ///
-    /// #[derive(Debug, PartialEq, FromQueryResult)]
-    /// struct SelectResult {
-    ///     name: String,
-    ///     num_of_cakes: i32,
-    /// }
-    ///
-    /// # let _: Result<(), DbErr> = smol::block_on(async {
-    /// #
-    /// let res: Vec<SelectResult> = Statement::from_sql_and_values(
-    ///     DbBackend::Postgres,
-    ///     r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
-    ///     vec![],
-    /// )
-    /// .into_model::<SelectResult>()
-    /// .all(&db)
-    /// .await?;
-    ///
-    /// assert_eq!(
-    ///     res,
-    ///     vec![
-    ///         SelectResult {
-    ///             name: "Chocolate Forest".to_owned(),
-    ///             num_of_cakes: 1,
-    ///         },
-    ///         SelectResult {
-    ///             name: "New York Cheese".to_owned(),
-    ///             num_of_cakes: 1,
-    ///         },
-    ///     ]
-    /// );
-    /// #
-    /// # Ok(())
-    /// # });
-    ///
-    /// assert_eq!(
-    ///     db.into_transaction_log(),
-    ///     vec![Transaction::from_sql_and_values(
-    ///         DbBackend::Postgres,
-    ///         r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
-    ///         vec![]
-    ///     ),]
-    /// );
-    /// ```
-    pub fn into_model<M>(self) -> SelectorRaw<SelectModel<M>>
-    where
-        M: FromQueryResult,
-    {
-        SelectorRaw::<SelectModel<M>>::from_statement(self)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/database/statement.rs
```diff
@@ -43,6 +43,75 @@ impl Statement {
...
+    /// assert_eq!(
+    ///     db.into_transaction_log(),
+    ///     vec![Transaction::from_sql_and_values(
+    ///         DbBackend::Postgres,
+    ///         r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
+    ///         vec![]
+    ///     ),]
+    /// );
+    /// ```
+    pub fn into_model<M>(self) -> SelectorRaw<SelectModel<M>>
```
billy1624, line 109: I think we should rename `into_model()` to `select()`. As later on we might want `update()`, `delete()`...etc.

src/executor/select.rs
```diff
@@ -263,6 +263,20 @@ impl<S> SelectorRaw<S>
 where
     S: SelectorTrait,
 {
+    // Create `SelectorRaw` from Statment. Executing this `SelectorRaw` will
+    // return a type `M` which implement `FromQueryResult`.
+    //
+    // Helper function used by `Statment.into_model()`
+    pub(crate) fn from_statement<M>(stmt: Statement) -> SelectorRaw<SelectModel<M>>
```
billy1624, line 270: Perhaps we can add a alias for this method, let say `new()`.

src/database/statement.rs
MuhannadAlrusayni: Hmm, what's the difference between `select()` and  `update()`, `delete()` .. etc ?

src/executor/select.rs
MuhannadAlrusayni: I'm not sure, this function only accept `M: FromQueryResult` and `Statement` to create `SelectorRaw<..>` it's kinda narrowed to be used only by `Statement` and a type `M: FromQueryResult`, while SelectorRaw<..>` would accept more types than that.

billy1624: Sorry, I tag on the wrong line. It should be [this](https://github.com/SeaQL/sea-orm/blob/38a4d3a810e8ccf1567266646de140edc4291a21/src/database/statement.rs#L26-L34) method.

src/database/statement.rs
billy1624: - Calling `select()` we get any type that impl `FromQueryResult`

- Calling `update()` we get `UpdateResult`

- Calling `delete()` we get `DeleteResult`

This just allow us to further extends the API of `Statement` further in the future.
Btw... we are not going to implement `update()` / `delete()` in this PR unless you want to loll.

src/executor/select.rs
MuhannadAlrusayni: Oh, I'm not sure if understand your point yet, but I have not created this method (`from_sql_and_values`).

billy1624: Yes, it's not. But would it be more convenient if we introduce a new alias for it (`from_sql_and_values `)?

src/database/statement.rs
MuhannadAlrusayni: hmm, I named it `into_model()` to be consistent with the rest of the SeaORM API. If that's not good pick, we may remove this method from `Statement` and put it somewhere else ? 

Here are some options I can think of (I would prefer the 1st option if we are going to change this):
1) add a function to any type that impl `FromQueryResult` that would return `SelectorRaw<..>`
```rust
pub trait FromQueryResultExt: FromQueryResult + Sized {
    fn query_by(stmt: Statement) -> SelectorRaw<SelectModel<Self>> {
        SelectorRaw::<SelectModel<Self>>::from_statement(stmt)
    }
}

impl<T: FromQueryResult> FromQueryResultExt for T {}
```
This way we can use it like:
```rust
let res: Vec<SelectResult> = SelectResult::query_by(Statement::from_sql_and_values(
    DbBackend::Postgres,
    r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
    vec![],
))
.all(&db)
.await?;
```

2) maybe we should change `SelectorRaw::from_statement` to be public, and the example in the main comment would be:
```rust
let stmt = Statement::from_sql_and_values(
    DbBackend::Postgres,
    r#"SELECT "cake"."name", count("cake"."id") AS "num_of_cakes" FROM "cake""#,
    vec![],
);

let res: Vec<SelectResult> = SelectorRaw::from_statemnet(stmt)
.all(&db)
.await?;
```

billy1624: (2.) seems good to me.

Any comments? @tyt2y3

src/executor/select.rs
MuhannadAlrusayni: hmm, to my personal taste I would prefer to not only rename/alias this function but `from_string` as well :D , so I would suggest:
- `from_string` -> `new`
- `from_sql_and_values` -> `with_values`

I'm still not sure if this is the right place to discuss this xD that's why I was confused at 1st, maybe we should open another PR/issue to discuss this farther ?

src/database/statement.rs
MuhannadAlrusayni: maybe I should clarify why I preferred the 1st option, again consistency, option 1 follow the same way how entities get fetched:
```rust
// fetching entities 
let cheese: Option<cake::Model> = Cake::find(1).one(db).await?;

// fetching types that impl `FromQueryResult` using option 1
let res: Option<SelectResult> = SelectResult::query_by(stmt).one(db).await?;
```

tyt2y3: Actually... why not both? Both looks good.
Yes, and making a new Trait is a good idea too.
My only comment is perhaps `query` as the method name would suffice.

src/executor/select.rs
tyt2y3: Agreed. `from_sql_and_values` is not the best name. Back then I wasn't sure how many constructors it should have.
Let's split it to another PR.

src/database/statement.rs
MuhannadAlrusayni: > why not both? Both looks good.
just to make sure, here is the changes I'll made:
- remove `into_model()` from `Statemenet`
- write both options 1 and 2.

 >  perhaps query as the method name would suffice.
 I don't have strong preference here, if you like to see other options too:
 - `query()`
 - `find()` consistent with `Enitiy::find()`, I prefer this other the next one if are not going to add more variant of `find_*` in the future for `FromQueryResult`.
 - `find_by_statement()` consistent with `Entity::find_by_id()` and the name tells that it need `Statement` to be called

src/executor/select.rs
MuhannadAlrusayni: should I mark this as resolved ?

tyt2y3: Sure

src/database/statement.rs
MuhannadAlrusayni: I think this conversation should be resolved by https://github.com/SeaQL/sea-orm/pull/167/commits/e71a74b0c1cf4bc391733d35a1e8c2ac12269a04

tyt2y3:
@MuhannadAlrusayni thank you for the contribution.

On a second look I found that `Sized` is already a requirement of `FromQueryResult`, so I merged the method back into `FromQueryResult`.

PR #172, opened 2021-09-17
Actix integration example

samsamai:


examples/actix_example/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+actix-files = "0.5"
+tera = "1.8.0"
+dotenv = "0.15"
+listenfd = "0.3.3"
+
+# remove `path = ""` in your own project
+sea-orm = { path = "../../", version = "^0.2", features = [
+  "macros",
+  "sqlx-all",
+  "runtime-async-std-native-tls",
```
tyt2y3, line 27: I am confused. It shouldnt compile? I thought we needed `runtime-actix-native-tls` ?

examples/actix_example/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+use tera::Tera;
+
+mod post;
+pub use post::Entity as Post;
+mod setup;
+
+const DEFAULT_POSTS_PER_PAGE: usize = 25;
+
+struct AppState {
+    db_url: String,
```
tyt2y3, line 27: Is it possible for us to store the DbConn inside the AppState? Right now we are creating a new connection for each request, which certainly isn't the best practice.

tyt2y3 (commented):
It's nice that we are reusing the same `tera` templates.
Looks good indeed!

examples/actix_example/src/main.rs
samsamai: Yes I agree, I tried to do this without success. With both Rocket and Actix I ran into this issue. Both frameworks expect to get a database pool which they can call on to get a connection, this might be a good feature to add to SeaORM but I'm not sure how to do it.

Oodachi: > 是的，我同意，我尝试这样做但没有成功。对于 Rocket 和 Actix，我遇到了这个问题。两个框架都希望获得一个数据库池，他们可以调用它来获得连接，这可能是添加到 SeaORM 的一个很好的功能，但我不知道如何去做。

I think we should use the `Arc` packaging

Oodachi: I think we should use the `Arc` packaging

tyt2y3: Hello! Can you share a bit more about the approach in your actix-web-async-graphql-sqlx?

samsamai: > I think we should use the `Arc` packaging

Thanks Oodachi for the hint! I've committed a version using Arc to pass the database connection, all the requests will use the same connection in the case.

Oodachi: > Hello! Can you share a bit more about the approach in your actix-web-async-graphql-sqlx?

It seems unnecessary, maybe a custom type is a better choice.

```rust
pub type SharedState = Arc<State>;
```

Acidic9: Yes with Actix you can use Arc.. But I don't think you need to. Sqlx's connection is designed to be cloned, so you should be make the connection in fn main, and put the connection in the app data.

Also, make sure to create the pool *outside* `HttpServer::new`, otherwise it'll create a new pool for each thread.

tyt2y3: My bad. I thought I merged the "Add `clone` trait to `DbConn`" into `0.2.2`. Turn out I didn't. It will be included in `0.2.3` I suppose.

examples/actix_example/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+mod setup;
+
+const DEFAULT_POSTS_PER_PAGE: usize = 25;
+
+#[derive(Debug, Clone)]
+struct AppState {
+    templates: tera::Tera,
+    conn: Arc<DatabaseConnection>,
+}
+type SharedState = Arc<AppState>;
```
Acidic9, line 27: The `conn: Arc<DatabaseConnection>` does not need to be wrapped in an Arc as sqlx's connection struct is designed to be cloned already.

Perhaps this can be updated to this?
```rust
struct AppState {
    templates: Arc<tera::Tera>,
    conn: DatabaseConnection,
}
```

samsamai: Great! Thanks for the input everyone, looks like with the latest updates we don't need Arc at all.

billy1624:
@tyt2y3 regarding why we can't use `runtime-actix-native-tls` here

```toml
sea-orm = { path = "../../", version = "^0.2", features = [
  "macros",
  "sqlx-all",
  "runtime-actix-native-tls",
], default-features = false }
```

https://github.com/launchbadge/sqlx/issues/1117

tyt2y3:
> @tyt2y3 regarding why we can't use `runtime-actix-native-tls` here
> [launchbadge/sqlx#1117](https://github.com/launchbadge/sqlx/issues/1117)

Interesting

billy1624:
Hey @marlon-sousa, do you know when actix-web v4 will be stable?

tyt2y3:
Yup, the SQLx folks also said, this question comes up every other week ) haha.
The story is, yes, latest SQLx depends on a tokio version not compatible with Actix 3.

So ideally we can upgrade to Actix 4, but it's still in beta, so meanwhile it could still be meaningful to use Actix 3, as I guess it will take some time before it goes out of beta.

Performance-wise, I am not sure whether or not it will make a significance.

Finally thank you for trusting and depending on us for your project. We will together make it a success!

marlon-sousa:
Hello,

When is actix-web 4 going to be released is such a question everyone wants to know, including me ...

Still some question persists:

1. What is the difference between runtime-actix-** and runtime-tokio-** features, since actix itself uses tokio?
2. What are the consequences to imbed an async-std runtime inside a tokio runtime? Is it safe or reasonable enough to do it in production? Because as soon as this example is released everyone using actix-web will start copying and pasting it and using it for their projects in production as this is going to be the official way of doing it, as documented by the project authors.

My understanding is that every time a call on any sea-orm function happens they will be scheduled in a different event loop, and I am not sure if this is or isn't blocking from the point of view of the original caller. If a full runtime plus future poolings and such happens inside another runtime event loop this has potential to cause a huge enormous mess in actix which basicly needs that everything inside their threads be fast and light.

billy1624:
Perhaps we can have two actix examples? v3 & v4 one each

Acidic9:
Actix beta v4 has been out for quite some time. I highly suggest anyone starting a new project with actix should use the beta v4. v3 of Actix uses an outdated version of tokio and it makes it painful to work with other libraries which require tokio > v1

samsamai:
#190

PR #177, opened 2021-09-18
Transaction continued

tyt2y3:
continue #156
1. [x] remove the is_mock_connection checks and still make the test suite pass
1. [ ] investigate how to remove the extra * from Rocket example save(&*conn)
1. [ ] enhance our MockConnection such that it also 'supports' transaction


tyt2y3:
> investigate how to remove the extra * from Rocket example save(&*conn)

It seems that we need a dynamic dispatch version `&dyn ConnectionTrait` to really eliminate the `&*conn` quirk.

PR #182, opened 2021-09-21
`find_linked` join with table alias

### LINKED ISSUE #163: `find_linked()` generates invalid SQL for self referencing associations
The implementation of `find_linked()` does not use aliases on join statements, so for self referencing many to many associations such as `TABLE -> RELATES_TO -> TABLE` the generated query is invalid. Other ORMs typically generate autoincremented aliases for each relation in this case.

billy1624:
Resolve #163

### DIFF
--- a/src/query/join.rs
+++ b/src/query/join.rs
@@ -74,7 +74,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::tests_cfg::{cake, cake_filling, cake_filling_price, filling, fruit};
+    use crate::tests_cfg::{cake, cake_filling, cake_filling_price, entity_linked, filling, fruit};
     use crate::{ColumnTrait, DbBackend, EntityTrait, ModelTrait, QueryFilter, QueryTrait};
     use pretty_assertions::assert_eq;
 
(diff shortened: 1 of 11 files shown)


tyt2y3:
Can we have a more comprehensive test case with more levels of joins?

billy1624:
Added some more test cases 1a61f09

src/query/join.rs
```diff
@@ -76,6 +76,7 @@ where
 mod tests {
     use crate::tests_cfg::{cake, cake_filling, cake_filling_price, filling, fruit};
     use crate::{ColumnTrait, DbBackend, EntityTrait, ModelTrait, QueryFilter, QueryTrait};
+    use pretty_assertions::assert_eq;
 
     #[test]
     fn join_1() {
```
tyt2y3, line 82: Can we have a slightly more elaborate unit test after here, even if the relation does not make sense?

tyt2y3 (changes requested):

src/query/join.rs
billy1624: More level of joins?

tyt2y3: Yes, 3 or 4 levels maybe?

billy1624: Checkout a4f2e3c

tyt2y3:
Can you move the Link definitions outside of the Entity files? perhaps to a separate file called `entity_links`?
I feel like it is cluttering the Entity file a bit

billy1624:
Yes we can!

PR #184, opened 2021-09-21
Streaming support

nappa85:
Streaming support is often a plus, for example when you don't want to allocate the entire resultset only to iterate it and apply some transformation.

This implementation uses the crate ouroboros to cover the self-reference problem in a safe way, the streaming itself is based on sqlx

If you like it, I can use the same technique to create an independent Transaction object that doesn't live only inside a closure, as it's implemented now

### DIFF
--- a/tests/stream_tests.rs
+++ b/tests/stream_tests.rs
@@ -2,7 +2,8 @@ pub mod common;
 
 pub use common::{bakery_chain::*, setup::*, TestContext};
 pub use sea_orm::entity::*;
-pub use sea_orm::{QueryFilter, ConnectionTrait};
+pub use sea_orm::{QueryFilter, ConnectionTrait, DbErr};
+use futures::StreamExt;
 
 #[sea_orm_macros::test]
 #[cfg(any(
@@ -11,8 +12,6 @@ pub use sea_orm::{QueryFilter, ConnectionTrait};
     feature = "sqlx-postgres"
 ))]
 pub async fn stream() -> Result<(), DbErr> {
-    use futures::StreamExt;
-
     let ctx = TestContext::new("stream").await;
 
     let bakery = bakery::ActiveModel {
--- a/tests/stream_tests.rs
+++ b/tests/stream_tests.rs
@@ -10,7 +10,7 @@ pub use sea_orm::{QueryFilter, ConnectionTrait};
     feature = "sqlx-sqlite",
     feature = "sqlx-postgres"
 ))]
-pub async fn stream() {
+pub async fn stream() -> Result<(), DbErr> {
     use futures::StreamExt;
 
     let ctx = TestContext::new("stream").await;
@@ -21,19 +21,18 @@ pub async fn stream() {
         ..Default::default()
     }
     .save(&ctx.db)
-    .await
-    .expect("could not insert bakery");
+    .await?;
 
     let result = Bakery::find_by_id(bakery.id.clone().unwrap())
        .stream(&ctx.db)
-        .await
-        .unwrap()
+        .await?
         .next()
         .await
-        .unwrap()
-        .unwrap();
+        .unwrap()?;
 
     assert_eq!(result.id, bakery.id.unwrap());
 
     ctx.delete().await;
+
+    Ok(())
 }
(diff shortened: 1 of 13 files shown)


tests/stream_tests.rs
```diff
@@ -0,0 +1,39 @@
...
+    .expect("could not insert bakery");
+
+    let result = Bakery::find_by_id(bakery.id.clone().unwrap())
+       .stream(&ctx.db)
+        .await
+        .unwrap()
+        .next()
+        .await
+        .unwrap()
+        .unwrap();
```
billy1624, line 34: Can we replace some of the `unwrap()` with `?` :P

nappa85: It's only one unwrap more than the `find_by_id_with_result` test that I've used as base for this test :P
TTo use `?` I have to make a test that returns `Result`, is it ok?

billy1624: Thanks!!

tyt2y3:
Let's discuss a bit on the background.

So you have a large result set, and you want to retrieve it slowly?

I think the potential drawback with streaming slowly is that it occupies the connection and it actually hurts concurrency.

On the other hand, we can use a paginator which fetches by batch and each row at once, though the query has to be repeated each batch. 

Okay so back to the code, why do we need a new connection enum? Is it because we have to rule out MockConnection?

nappa85:
> So you have a large result set, and you want to retrieve it slowly?
Slowly? Streaming would be faster, now slower.
Let's have, for example
```rust
Entity::find().all(conn).await
```
How does it works?
Well, the workflow is:
* `sqlx::Executor::fetch_all` calls `sqlx::Executor:fetch` that returns the stream, then uses try_collect that allocates the stream into a Vec
* `sea_orm::driver::sqlx_{mysql|postgres|sqlite}::query_all` trasforms back the Vec into an Iterator, applies Into::into and allocates back into a Vec
* `sea_orm::executor::select::Selector::all` trasforms back the Vec into an Iterattor, applies SelectorTrait::from_raw_query_result and allocates back into a Vec

So for a single query we have 3 different allocations.
Using streams data is allocated only at the end, only if who is using the stream wants to allocate.
This is a big speed-up, and a big memory improvement.

> I think the potential drawback with streaming slowly is that it occupies the connection and it actually hurts concurrency.
The way you use connections is that, for every query take one connection from the pool, use it once and put it back on the pool.
You don't have any guarantee that on two consecutive queries the connection is the same.
The only difference with stream is that the connection is used for the entire life of the stream, that yes, can be longer because you can take the stream and never consume it, but that doesn't hurt concurrency, because other queries would use other connections, the worst that can happen is a really badly written program that takes so many streams and never consumes them that it reach db server's max connections number, but also this error is managed, there won't be a panic.

> Okay so back to the code, why do we need a new connection enum? Is it because we have to rule out MockConnection?
I've made the Connection enum simply to hide out sqlx traits, or we would end up with `QueryStream<sqlx::MySql>`, `QueryStream<sqlx::Postgres>` or `QueryStream<sqlx::Sqlite>`
I don't want to rule out MockConnection, simply didn't had the time to think about it, and first of all I wanted to find out if streaming was a possible implementation and discuss the design with you

tyt2y3:
Okay understand now.
I originally thought you'd do some computational processing or IO operation meanwhile consuming the stream.

> I've made the Connection enum simply to hide out sqlx traits, or we would end up with QueryStream<sqlx::MySql>, QueryStream<sqlx::Postgres> or QueryStream<sqlx::Sqlite>

Then perhaps we can simply call it `SqlxQueryStream` )

nappa85:
> Then perhaps we can simply call it `SqlxQueryStream` )

Well, it would be a bit uglly, and I think the enum would be the solution to implement streaming for the mocked connection.

If I'm going to apply the same technique for the Transaction, the enum would be useful there too

tyt2y3:
> If you like it, I can use the same technique to create an independent Transaction object that doesn't live only inside a closure, as it's implemented now

Oh I get what you mean now. Sounds like a better implementation indeed.

tyt2y3:
Close in favor of #199

PR #189, opened 2021-09-23
add debug_query macro make get raw_sql becomes simply.

baoyachi:
Let's take insertion as an example,before we want get sea-orm raw_sql, it's may be import trait namespace with :`use sea_orm::QueryTrait;`

```rust
use sea_orm::QueryTrait;
let pear = fruit::ActiveModel {
            name: Set("Pear".to_owned()),
            ..Default::default() // all other attributes are `Unset`
        };

let raw_sql = Entity::insert(pear).build(DatabaseBackend::MySql).to_string();
``` 

Now, you can use `debug_query` macro get raw_sql becomes simply.It does not need to specify a specific `DbBackend`,but can be obtained through `get_database_backend` function with `DatabaseConnection`. That not import trait namespace, the  
[debug_query](https://github.com/SeaQL/sea-orm/pull/189/files#diff-107bf27907dbd02679a1e1bf776f50c5a2eac62da3a6cbbdd240bbafe29035dfR72-R76) macro make use simply.
```rust
let pear = fruit::ActiveModel {
            name: Set("Pear".to_owned()),
            ..Default::default() // all other attributes are `Unset`
        };

let raw_sql = debug_query!(&c,&conn).to_string();
assert_eq!(raw_sql,r#"INSERT INTO "cake" ("id", "name") VALUES (1, 'Apple Pie')"#);

let raw_sql = debug_query_fmt!(&c,DbBackend::Sqlite);
assert_eq!(raw_sql,r#"INSERT INTO `cake` (`id`, `name`) VALUES (1, 'Apple Pie')"#);
```

* debug_query!: support accpeted `&DatabaseConnection` reference or `DbBackend` type .
* debug_query_fmt!: with `Statement` to_string().


baoyachi:
Sorry,wait a minute, there is a better solution

billy1624:
Hey @baoyachi, if we change the `db_backend` arg and make it accept anything that `impl IntoDbBackend` would this make the API ergonomic enough?

https://github.com/SeaQL/sea-orm/blob/6861802a6d9c6eabe554cb984a56d55dca7b2e62/src/query/traits.rs#L16-L23

Introduce a new trait `IntoDbBackend` and implement it...
```rust
pub trait IntoDbBackend { ... }

impl IntoDbBackend for DatabaseConnection { ... }

impl IntoDbBackend for DatabaseBackend { ... }
```

Then we can write...
```rust
let raw_sql = Entity::insert(pear).build(&db).to_string();

let raw_sql = Entity::insert(pear).build(DatabaseBackend::MySql).to_string();
```

baoyachi:
@billy1624  That's great. I thought about this before sending PR, which may involve large changes. Now I think I can do that.

billy1624:
Any comments? @tyt2y3

tyt2y3:
Okay so, I would prefer `debug_query` to output a string and remove `debug_query_fmt`.
This method overloading strategy is the first time I've ever used!

baoyachi:
> Okay so, I would prefer `debug_query` to output a string and remove `debug_query_fmt`.
> This function overloading strategy is the first time I've ever used!

@tyt2y3 I have removed `debugged_ query_fmt`  macro and use `debug_query` instead.
Added `gen_statment` macro.This seems necessary because it receives one of the multiple parameters of `DatabaseConnection` or `DbBackend`.

src/database/connection.rs
```diff
@@ -16,14 +16,32 @@ pub enum DatabaseConnection {
...
 
+pub trait IntoDbBackend {
+    fn build(&self, statement: &impl StatementBuilder) -> Statement;
+
+    fn get_query_builder(&self) -> Box<dyn QueryBuilder>;
+}
+
 #[derive(Debug, Copy, Clone, PartialEq)]
-pub enum DatabaseBackend {
+pub enum DbBackend {
```
tyt2y3, line 26: I know probably no one is using `DatabaseBackend`. But removing it is bad.

src/database/connection.rs
```diff
@@ -16,14 +16,32 @@ pub enum DatabaseConnection {
 
 pub type DbConn = DatabaseConnection;
 
+pub trait IntoDbBackend {
```
tyt2y3, line 19: I think there is actually no need to introduce a new trait here.
There is only 1 impl and it's not being used in any trait bound.
It might have been useful before, but seemingly not in this final solution.
So I'd suggest moving the methods back to its original `impl DbBackend` block.

src/error.rs
```diff
@@ -24,6 +24,8 @@ impl std::error::Error for ColumnFromStrErr {}
 
 impl std::fmt::Display for ColumnFromStrErr {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        use chrono::format;
```
tyt2y3, line 27: This line also seems to be redundant (I am not very sure though).

tyt2y3 (changes requested):
Sorry for the late review. When I go through a PR, I'd actually try to justify every line change.
As such, the more line changes are there, the more effort it takes to go through.
Thank you for your contribution nonetheless!

src/error.rs
baoyachi: Removed

src/database/connection.rs
baoyachi: Removed `IntoDbBackend `,roll back to original impl DbBackend block.

tyt2y3:
Thank you

PR #192, opened 2021-09-24
Codegen classify ordinary relation wrongly as conjunct relation

### LINKED ISSUE #186: Primary & foreign key referencing same column panics
Hi, I've got the following SQL, which crashes the generator. 
Commenting out the foreign key line, or changing the primary key to `PRIMARY KEY (a)` makes it work fine, but I'm not sure why, as I think what I'm doing is valid?

```sql
CREATE TABLE IF NOT EXISTS foo (
  a TEXT NOT NULL,
  PRIMARY KEY (a)
);

CREATE TABLE IF NOT EXISTS bar (
  a TEXT NOT NULL,
  b TEXT NOT NULL,

  FOREIGN KEY (a) REFERENCES foo (a),
  PRIMARY KEY (a, b)
);
```

The error I get is:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/callym/.cargo/registry/src/github.com-1ecc6299db9ec823/sea-orm-codegen-0.2.3/src/entity/transformer.rs:69:83
```

Running v0.2.3

The example is maybe a bit unclear, my use case is trying to store plant information (family, genus, species, etc), where (species name) isn't unique (there's tons of plants with a species name of e.g. 'alba'), but (genus name + species name) **is** unique.

billy1624:
Resolve #186

### DIFF
--- a/sea-orm-codegen/src/entity/transformer.rs
+++ b/sea-orm-codegen/src/entity/transformer.rs
@@ -63,6 +63,8 @@ impl EntityTransformer {
             entities.insert(table_name.clone(), entity.clone());
             for (i, mut rel) in relations.into_iter().enumerate() {
                 let is_conjunct_relation = entity.primary_keys.len() == entity.columns.len()
+                    && rel.columns.len() == 2
+                    && rel.ref_columns.len() == 2
                     && entity.primary_keys.len() == 2;
                 match is_conjunct_relation {
                     true => {


tyt2y3:
Anything need to add?

billy1624:
It's better to have an end to end test case for codegen. But it would be included on other big PR.

tyt2y3:
okay

Issue #200, opened 2021-09-27
Postgres URI scheme designator can be  be either `postgresql://` or `postgres://`


billy1624:
> The URI scheme designator can be either `postgresql://` or `postgres://`.

Extracted from https://www.postgresql.org/docs/current/libpq-connect.html

baoyachi:
May be need define const with DbType scheme in sea-orm.

PR #201, opened 2021-09-27
Adapter postgres url scheme 

baoyachi:
#200


billy1624:
Ohoh what happened? loll

PR #202, opened 2021-09-27
Adapter postgres url scheme

baoyachi:
working for #200 

* adapter postgres url scheme
* add new method in `DbBackend` with name `url_starts_with`.

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -12,21 +12,40 @@ pub use transaction::*;
 
 use crate::DbErr;
 
+#[derive(Debug)]
+pub enum DbScheme {
+    Postgres,
+    Mysql,
+    Sqlite,
+}
+
+impl DbScheme {
+    pub fn starts_with(self, base_url: &str) -> bool {
+        match self {
+            DbScheme::Postgres => {
+                base_url.starts_with("postgres://") || base_url.starts_with("postgresql://")
+            }
+            DbScheme::Mysql => base_url.starts_with("mysql://"),
+            DbScheme::Sqlite => base_url.starts_with("sqlite:"),
+        }
+    }
+}
+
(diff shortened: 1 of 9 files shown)


src/database/mod.rs
```diff
@@ -12,21 +12,40 @@ pub use transaction::*;
 
 use crate::DbErr;
 
+#[derive(Debug)]
+pub enum DbScheme {
+    Postgres,
+    Mysql,
+    Sqlite,
+}
```
billy1624, line 20: Thanks!! @baoyachi 

- Perhaps we can add the `starts_with()` method to `DatabaseBackend` instead of introducing a new enum `DbScheme`?
- And rename the `starts_with()` method to `url_starts_with()`?

https://github.com/SeaQL/sea-orm/blob/1eb0e5cbb278413b327b1512e66b591ef66cffb7/src/database/connection.rs#L19-L24

baoyachi: It's good 👍 .

tyt2y3:
Thank you so much for this. I think my only nitpick would be to think of a better name for the method `url_starts_with`.
Because in left to right order, `url starts with mysql` makes sense.
But in reverse, `mysql starts with url` does not.
May be, I suggest, to use `is_prefix_of` as in `mysql is prefix of url`.
Sounds good?

baoyachi:
> Thank you so much for this. I think my only nitpick would be to think of a better name for the method `url_starts_with`.
> Because in left to right order, `url starts with mysql` makes sense.
> But in reverse, `mysql starts with url` does not.
> May be, I suggest, to use `is_prefix_of` as in `mysql is prefix of url`.
> Sounds good?

Done.

tyt2y3 (approved):
Awesome!

PR #205, opened 2021-09-27
Drop `Default` trait bound of `PrimaryKeyTrait::ValueType`

### LINKED ISSUE #171: Why `PrimaryKeyTrait::ValueType: Default` is needed?
> Hi, it seems that `PrimaryKeyTrait::ValueType` bound to impl `Default`, can we remove this bound ? since some types cannot implement `Default`, for example I have used `Username(String)` as primary key for my table/entity.

> or maybe replace `Default` bound with function, something like fn `default_value() -> Option<Self::ValueType>` for those types that can generate default value.

> this issue blocks me from porting my project to use SeaORM

Originally posted on Discord by @MuhannadAlrusayni

billy1624:
Resolve #171

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::*, DatabaseConnection, DeleteResult, EntityTrait, Iterable, PrimaryKeyToColumn,
-    PrimaryKeyTrait, Value,
+    PrimaryKeyValue, Value,
 };
 use async_trait::async_trait;
 use std::fmt::Debug;
@@ -70,23 +70,18 @@ pub trait ActiveModelTrait: Clone + Debug {
     async fn insert(self, db: &DatabaseConnection) -> Result<Self, DbErr>
     where
         <Self::Entity as EntityTrait>::Model: IntoActiveModel<Self>,
+        <<Self as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey:
+            PrimaryKeyValue<<Self as ActiveModelTrait>::Entity>,
     {
         let am = self;
         let exec = <Self::Entity as EntityTrait>::insert(am).exec(db);
         let res = exec.await?;
-        // Assume valid last_insert_id is not equals to Default::default()
-        if res.last_insert_id
-            != <<Self::Entity as EntityTrait>::PrimaryKey as PrimaryKeyTrait>::ValueType::default()
-        {
-            let found = <Self::Entity as EntityTrait>::find_by_id(res.last_insert_id)
-                .one(db)
-                .await?;
-            match found {
-                Some(model) => Ok(model.into_active_model()),
-                None => Err(DbErr::Exec("Failed to find inserted item".to_owned())),
-            }
-        } else {
-            Ok(Self::default())
+        let found = <Self::Entity as EntityTrait>::find_by_id(res.last_insert_id)
+            .one(db)
+            .await?;
+        match found {
+            Some(model) => Ok(model.into_active_model()),
+            None => Err(DbErr::Exec("Failed to find inserted item".to_owned())),
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/entity/active_model.rs
```diff
@@ -70,23 +70,18 @@ pub trait ActiveModelTrait: Clone + Debug {
     async fn insert(self, db: &DatabaseConnection) -> Result<Self, DbErr>
     where
         <Self::Entity as EntityTrait>::Model: IntoActiveModel<Self>,
+        <<Self as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey:
+            PrimaryKeyValue<<Self as ActiveModelTrait>::Entity>,
```
billy1624, line 74: .

src/entity/base_entity.rs
```diff
@@ -299,6 +299,8 @@ pub trait EntityTrait: EntityName {
     fn insert<A>(model: A) -> Insert<A>
     where
         A: ActiveModelTrait<Entity = Self>,
+        <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey:
+            PrimaryKeyValue<<A as ActiveModelTrait>::Entity>,
```
billy1624, line 303: .

billy1624 (commented):
I'm thinking how to get rid of all these...
```rust
// This
<<Self as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey:
            PrimaryKeyValue<<Self as ActiveModelTrait>::Entity>,

// And this
<<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey:
            PrimaryKeyValue<<A as ActiveModelTrait>::Entity>,
```

billy1624:
Got this error, somehow...

> Error: Exec("error returned from database: database table is locked: metadata")

https://github.com/SeaQL/sea-orm/runs/3719025255?check_suite_focus=true#step:5:290

tyt2y3:
Thanks. It looks good generally.

Though I don't understand why we have to add a new trait `PrimaryKeyValue`.

I think the existing APIs provides sufficient info for us to implement `extract_primary_key_value_from_model`.

src/executor/insert.rs
```diff
@@ -85,21 +91,27 @@ where
 {
     type PrimaryKey<A> = <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey;
     type ValueTypeOf<A> = <PrimaryKey<A> as PrimaryKeyTrait>::ValueType;
-    let last_insert_id = match db.get_database_backend() {
-        DbBackend::Postgres => {
+    let last_insert_id_opt = match db {
+        #[cfg(feature = "sqlx-postgres")]
+        DatabaseConnection::SqlxPostgresPoolConnection(conn) => {
```
tyt2y3, line 96: Oh. and here. the behavior has been changed since 85eadd38d222b07621ff5b5ca55b650295bbe682

tyt2y3 (changes requested):

src/executor/insert.rs
billy1624: Let me merge from latest master and check

billy1624:
> Though I don't understand why we have to add a new trait PrimaryKeyValue.

I don't want to include `get_primary_key_value` inside `PrimaryKeyTrait` because this will requires changing either...
- Add type def `Entity: EntityTrait`
- Add generic `E: EntityTrait`

> I think the existing API provides sufficient info for us to implement extract_primary_key_value_from_model.

I think the only way to extract either `(A, B)` (tuple value) or `A` (single value) is to `impl PrimaryKeyValue` inside `DerivePrimayKey` and counting the number of primary key columns

tyt2y3:
I think, since our identity is only a tuple of 1, 2 or 3,
There can be ways to accomplish this without introducing a new trait.

billy1624:
Something like this?

https://github.com/SeaQL/sea-orm/blob/beb3ec62dc9f0a33c72dd8f176e217d23c3226aa/src/entity/base_entity.rs#L242-L257

tyt2y3:
Still WIP?

billy1624:
> I think, since our identity is only a tuple of 1, 2 or 3, there can be ways to accomplish this without introducing a new trait.
> 
> Iterable actually is able to tell us the arity I suppose.

Stuck on this

tyt2y3:
> Something like this?

Yes, should be.

billy1624:
I can avoid add a new trait by putting `extract_primary_key_value` method in `ActiveModelTrait`. And implement it inside proc_macros. How is that sounds?

tyt2y3:
No we don't need that either

PR #208, opened 2021-09-27
Support feteching T, (T, U), (T, U, P)..etc where they impl TryGetableMany

### LINKED ISSUE #180: Support feteching `T`, `(T, U)`, `(T, U, P)`..etc where they impl `TryGetableMany`
Hi, I think we need a way to fetch data from database without putting into a `Model`, instead we use tuple or single type, here is what's in my mind:
```rust
// where bool: TryGetableMany
let is_locked: Option<bool> = cake::Account::find()
    .select_only()
    .column(account::Column::IsLocked)
    .into_values() // this would return `Selector<?>`
    .one(db)
    .await?;
```

We can also extend this to `TryGetableMany` it self, that would allow us to write:
```rust
let stmt = "select name, count(post.user_id) as post_count .. group by name";
let user_posts = <(String, i32)>::from_statment(stmt, ["name", "post_count"]) // this would return `Selector<?>`
    .one(db)
    .await?;
```

sounds cool?

MuhannadAlrusayni:
Resolve #180

This is not the best API for sure! but we can start with this. One blocker that I got was `SelectorTrait::from_raw_query_result` is a function, which doesn't help when I want to store some data into `SelectGetableValue`, so I used generic type `C` to work around this, but yeah we don't get very ergonomic API.

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -314,6 +314,15 @@ where
     }
 }
 
+impl<T> TryGetableMany for (T,)
+where
+    T: TryGetableMany,
+{
+    fn try_get_many(res: &QueryResult, pre: &str, cols: &[String]) -> Result<Self, TryGetError> {
+        T::try_get_many(res, pre, cols).map(|r| (r,))
+    }
+}
+
 impl<A, B> TryGetableMany for (A, B)
 where
     A: TryGetable,
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "mock")]
 use crate::debug_print;
-use crate::DbErr;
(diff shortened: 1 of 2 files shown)


MuhannadAlrusayni:
if we could convert `SelectorTrait::from_raw_query_result` to be a method instead of function, that would help us have way better API in this PR.

But I think this is breaking change API.

billy1624:
Thanks!! @MuhannadAlrusayni 

> if we could convert SelectorTrait::from_raw_query_result to be a method instead of function, that would help us have way better API in this PR.

I think this is worth doing as this enable us to store metadata that is useful when performing `from_raw_query_result`.  And it's potentially useful for downstream types that implement `SelectorTrait`.

tyt2y3 (approved):
I think we can use `sea_query::Iden` to replace `ToString`.
I will make the changes.

PR #210, opened 2021-09-28
Update `ActiveModelBehavior` API

### LINKED ISSUE #161: Update `ActiveModelBehavior` API allowing custom validation logic
Issue raised by @MuhannadAlrusayni, thank you so much for the suggestions!

> Looking at `ActiveModelBehavior` I wounder what it's used for ? is it for controlling how the model get updated ? but these methods doesn't return `Result<..>`.
> I have some business logic that restrict how my model get updated/inserted and I thought that `ActiveModelBehavior` would help me here to do so, other wise I would need not to expose my `account::Entity` struct since SeaORM doesn't provide a way to restrict update/insert to match my business logic.

> let me tell you my use case, it is that I plan to make my backend into group of crates that's independents of each other, and I want each crates to be targeting a specific domain problem, so let's say I have crate called `users`, it's used to manage users accounts, sessions, rules. and I have another crate called `stroes`.
> 
> `stroes` crate depend on `users`, since each `store` should belong to a `User`.
> 
> the problem is that `users` crate encapsulate it's own data and their business logic, and when I use SeaORM I have to expose `user::Entity` to other crates such as `stores`, but since SeaORM doesn't provide a way to inject my business logic validation I cannot expose `user::Entity` to other crates at all.
> I am not sure if explained my problem enough.

billy1624:
Resolve #161

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -10,8 +10,7 @@ pub struct ActiveValue<V>
 where
     V: Into<Value>,
 {
-    // Don't want to call ActiveValue::unwrap() and cause panic
-    pub(self) value: Option<V>,
+    value: Option<V>,
     state: ActiveValueState,
 }
 
@@ -74,7 +73,7 @@ pub trait ActiveModelTrait: Clone + Debug {
         macro_rules! next {
             () => {
                 if let Some(col) = cols.next() {
-                    if let Some(val) = self.get(col.into_column()).value {
+                    if let Some(val) = self.get(col.into_column()).into_value() {
                         val
                     } else {
                         return None;


MuhannadAlrusayni:
Great job @billy1624, there is a few things that would be cool to have in this PR.

**1st)** is to rollback db changes on failure (in some cases?).
**2nd)** I think we should change the error type (`DbErr`) to user defined error associated with `ActiveModelBehavior::Error` type or `Box<dyn Error + Send + Sync + 'static`>` which would result to make the user code cleaner, I would prefer to use user defined error type for more flexibility in the API. Something like this maybe?:
```rust
pub trait ActiveModelBehavior: ActiveModelTrait {
    Type Error: std::error::Error;

    fn after_save(self, insert: bool, db: &DatabaseConnection) -> Result<Self, Self::Error> {
        Ok(self)
    }

    // .. omitted
}
```

billy1624:
> 1st) is to rollback db changes on failure.

This requires Transaction support and it's not merged into master yet.

> 2nd) I think we should change the error type (DbErr) to user defined error associated with ActiveModelBehavior::Error type or Box<dyn Error + Send + Sync + 'static>` which would result to make the user code cleaner, I would prefer to use user defined error type for more flexibility in the API.

Perhaps we can introduce `DbErr::Custom` then user can convert their custom errors with `.map_err(|e| DbErr::Custom( ... ))`. And we don't have to force user define extra type def `ActiveModelBehavior::Error`.

```rust
#[derive(Debug)]
pub enum DbErr {
    Conn(String),
    Exec(String),
    Query(String),
    Custom(Box<dyn Error>),
}
```

MuhannadAlrusayni:
Using `DbError::Custom` is fine pick, but I would still like the signature of `ActiveModelBehavior::*` to return `Result<Self, Box<dyn Error>>` so user code don't need to do the mapping every time they override these methods, we can do these mapping in SeaORM right ?

MuhannadAlrusayni:
we can also have an alias for such a result type like this:
```rust
// we can also name this BehaviorResult<T> or other good name, so we can add it in the prelude 
type Result<T> = Result<T, Box<dyn Error>>;

pub trait ActiveModelBehavior: ActiveModelTrait {
    fn before_save(self, insert: bool, db: &DatabaseConnection) -> Result<Self> { .. } 

    // .. omited
}
```

MuhannadAlrusayni:
Or another option is to `impl<T: Error> From<T> for DbErr` if possible so user can use `?` inside these functions.

billy1624:
> Or another option is to `impl<T: Error> From<T> for DbErr` if possible so user can use `?` inside these functions.

But this conversion will have a long lasting impact. Converting all errors defined in third-party crate into `DbErr::Custom`. Is this a good idea? @tyt2y3

I think it's okay as long as we don't change this conversion afterward. Otherwise, this will be a headache for downstream users.

tyt2y3:
So I think it is better to *not* provide a blanket implementation for now.

It looks good otherwise.

PR #211, opened 2021-09-28
Throw error if none of the db rows are affected

### LINKED ISSUE #181: Change the behavior of update() to return error when rows_affected is 0
`update()` can be used with filters. One might, for example, update a record but only if some column of itself still has some value set.

example:

```rust
MyEntity::update(ActiveRecord {
    id: Set(5),
    status: Set("ACTIVE"),
})
.filter(column::Status.eq("READY_FOR_ACTIVATION")
.exec(&db)
.await
```

This is useful to handle idempotency problems: I want to update status but only if, in the meantime, it hasn't yet been updated by another oprocess or something similar.

In this situation (or if I provide an id which does not exist) I have to know if my update has occurred successfully.

One can arg that I could use update_many, but using update gives me the advantage to have the active model itself be returned, which as far as I understand is not the case with update_many.

_Originally posted by @marlon-sousa in https://github.com/SeaQL/sea-orm/discussions/127#discussioncomment-1358559_

We should add a new Error variant, perhaps `RecordNotFound`

billy1624:
Resolve #181

### DIFF
--- a/src/executor/update.rs
+++ b/src/executor/update.rs
@@ -82,3 +82,108 @@ async fn exec_update(statement: Statement, db: &DatabaseConnection) -> Result<Up
         rows_affected: result.rows_affected(),
     })
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{entity::prelude::*, tests_cfg::*, *};
+    use pretty_assertions::assert_eq;
+
+    #[smol_potat::test]
+    async fn update_record_not_found_1() -> Result<(), DbErr> {
+        let db = MockDatabase::new(DbBackend::Postgres)
+            .append_query_results(vec![
+                vec![cake::Model {
+                    id: 1,
+                    name: "Cheese Cake".to_owned(),
+                }],
+                vec![],
+                vec![],
+            ])
+            .append_exec_results(vec![
+                MockExecResult {
+                    last_insert_id: 0,
+                    rows_affected: 1,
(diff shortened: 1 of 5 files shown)


tyt2y3:
We need unit tests using the mock connection on this behaviour too

billy1624:
> We need unit tests using the mock connection on this behaviour too

Added!

src/executor/update.rs
```diff
@@ -73,7 +73,117 @@ where
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{entity::prelude::*, tests_cfg::*, *};
+    use pretty_assertions::assert_eq;
+
+    #[smol_potat::test]
+    async fn update_record_not_found_1() -> Result<(), DbErr> {
+        let db = MockDatabase::new(DbBackend::Postgres)
+            .append_query_results(vec![
```
tyt2y3, line 94: I guess this is not needed. Let me try.

tyt2y3 (changes requested):
I have noticed a problem. this behavior should only applies to update_one, while for update many, it's fine to have updated 0 rows.

billy1624:
> I have noticed a problem. this behavior should only applies to update_one, while for update many, it's fine to have updated 0 rows.

Fixed

tyt2y3 (approved):

PR #214, opened 2021-09-30
Add support for `QueryValue`

Acidic9:
This PR is related to https://github.com/SeaQL/sea-query/pull/146

With these changes, users can implement `QueryValue` on their own types and use them in all queries including inserts and selects.

Implementing an enum would look like this:
```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, sqlx::Type)]
#[sqlx(type_name = "plan")]
#[sqlx(rename_all = "lowercase")]
pub enum Plan {
    Free,
    Basic,
    Pro,
}

impl QueryValue for Plan {
    fn query_value(&self, query_builder: &dyn QueryBuilder) -> String { self.to_string().query_value(query_builder) }
    fn primitive_value(&self) -> PrimitiveValue { self.to_string().into() }
    fn cast_as(&self) -> Option<&'static str> { Some("plan") }
}

impl str::FromStr for Plan {
    type Err = PlanFromStrError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "free" => Ok(Self::Free),
            "basic" => Ok(Self::Basic),
            "pro" => Ok(Self::Pro),
            _ => Err(PlanFromStrError),
        }
    }
}

impl ToString for Plan {
    fn to_string(&self) -> String {
        match self {
            Self::Free => "free".to_string(),
            Self::Basic => "basic".to_string(),
            Self::Pro => "pro".to_string(),
        }
    }
}

impl TryGetable for Plan {
    fn try_get(res: &QueryResult, pre: &str, col: &str) -> Result<Self, TryGetError> {
        let column = format!("{}{}", pre, col);
        match &res.row {
            QueryResultRow::SqlxPostgres(row) => {
                use sqlx::Row;
                sqlx::postgres::PgRow::try_get::<Option<Plan>, _>(row, column.as_str())
                    .map_err(|e| TryGetError::DbErr(sqlx_error_to_query_err(e)))
                    .and_then(|opt| opt.ok_or(TryGetError::Null))
            }
        }
    }
}
```
But all these impls could easily be hidden and simplified using a single derive macro.

---

Another example for a custom email domain type in postgres:
```sql
CREATE DOMAIN email AS citext CHECK (
  VALUE ~ '^[a-zA-Z0-9.!#$%&''*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$'
);
```
```rust
#[derive(Clone, Debug, PartialEq, Eq, sqlx::Type)]
#[sqlx(type_name = "citext")]
pub struct Email(pub String);

impl QueryValue for Email {
    /// Returns the value as an escaped string safe for use in SQL queries.
    fn query_value(&self, query_builder: &dyn QueryBuilder) -> String {
        self.0.query_value(query_builder) // self.0 is a string, so calling .query_value() on it automatically escapes it and wraps it in quotes
    }

    /// Primitive value for use in Database.
    fn primitive_value(&self) -> PrimitiveValue {
        self.0.clone().into() // self.0 is a &String, and `PrimitiveValue` implements `From<String>`
    }

    // No need for cast_as for DOMAIN types
}

// This could very easily use a macro to generate this.
impl TryGetable for Email {
    fn try_get(res: &QueryResult, pre: &str, col: &str) -> Result<Self, TryGetError> {
        let column = format!("{}{}", pre, col);
        match &res.row {
            QueryResultRow::SqlxPostgres(row) => {
                use sqlx::Row;
                sqlx::postgres::PgRow::try_get::<Option<Email>, _>(row, column.as_str())
                    .map_err(|e| TryGetError::DbErr(sqlx_error_to_query_err(e)))
                    .and_then(|opt| opt.ok_or(TryGetError::Null))
            }
        }
    }
}
```

### DIFF
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -82,7 +82,7 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
 
             fn set(&mut self, c: <Self::Entity as EntityTrait>::Column, v: sea_orm::Value) {
                 match c {
-                    #(<Self::Entity as EntityTrait>::Column::#name => self.#field = sea_orm::ActiveValue::set(v.unwrap()),)*
+                    #(<Self::Entity as EntityTrait>::Column::#name => self.#field = sea_orm::ActiveValue::set(v.primitive_value().unwrap()),)*
                     _ => panic!("This ActiveModel does not have this field"),
                 }
             }
--- a/sea-orm-macros/src/derives/model.rs
+++ b/sea-orm-macros/src/derives/model.rs
@@ -144,7 +144,7 @@ impl DeriveModel {
 
                 fn set(&mut self, c: <Self::Entity as sea_orm::entity::EntityTrait>::Column, v: sea_orm::Value) {
                     match c {
-                        #(<Self::Entity as sea_orm::entity::EntityTrait>::Column::#column_idents => self.#field_idents = v.unwrap(),)*
+                        #(<Self::Entity as sea_orm::entity::EntityTrait>::Column::#column_idents => self.#field_idents = v.primitive_value().unwrap(),)*
                         _ => panic!(#missing_field_msg),
                     }
                 }
--- a/src/database/statement.rs
+++ b/src/database/statement.rs
@@ -1,12 +1,12 @@
 use crate::DbBackend;
+pub use sea_query::Value;
 use sea_query::{inject_parameters, MysqlQueryBuilder, PostgresQueryBuilder, SqliteQueryBuilder};
-pub use sea_query::{Value, Values};
 use std::fmt;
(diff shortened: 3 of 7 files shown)


tyt2y3:
It really seems cool! At least for the API it offers.

Why the `cast` attribute is needed? In what cases where a cast is not needed?

Acidic9:
> Why the cast attribute is needed? In what cases where a cast is not needed?

The cast will wrap your type in a sql cast. In the case of the plan example, `CAST('free' as plan)`.
I think cast is only needed for `CREATE TYPE`, but `CREATE DOMAIN` does not require a cast.

PR #219, opened 2021-10-03
Add tables option to CLI generate entity

Ka1wa:
I have a big database of which I only need a couple tables generated. 

This PR introduces the `tables` option to the `sea-orm-cli generate entity` command  as follows:
```
-t, --tables <TABLES>                      Generate entity file for specified tables only (comma seperated)
```

This allows me to do the following:
`sea-orm-cli generate entity -o src/entity -t words,words_definitions`

Finally I made the option conflict with the existing `include-hidden-tables` because that seemed most logical.


tyt2y3:
Thank you for the contribution. Nicely done indeed.

PR #222, opened 2021-10-04
Transaction 3

tyt2y3:
Continue #199


nappa85:
> Was there a reason where `TransactionError` is a standalone enum instead of being a variant of DbErr? If not, I'd merge it into DbErr, and use a `Box<dyn Error>` for simplicity.

I don't like "Boxdyning" things, adds an indirection that is often avoidable.
If I remember correctly, `TransactionError` was made to wrap the error returned from the closure, but distinct between transaction begin errors and closure errors.

nappa85:
> @nappa85 have you tried nested transaction, i.e. multiple begin() and mutiple commit() on same connection?

Not extensively TBH, but should be trivial...
Have you found problems?

tyt2y3:
Just to remind myself of the nested transaction implementation of SQLx

https://github.com/launchbadge/sqlx/blob/b6e127561797fe9aababa24ec640275ecb9b42af/sqlx-core/src/transaction.rs#L218

tyt2y3:
> Is this complicated enough? 12800468b1b9c72be8a28cf5a993c74a9cd1a8ed @tyt2y3 

Yes but we need more

PR #224, opened 2021-10-04
Escape rust keywords with `r#` raw identifier

### LINKED ISSUE #217: sea-orm-cli: Entity generation fails if column name is a reserved keyword
Hello,

I am very excited by this project, and I think the work you're doing is amazing.

I encountered a problem, when I was trying to use `sea-orm-cli` to generate entities. I got an error as one of the columns in my database was named `type` which is a reserved keyword. `rustc` suggest fixing it by writing `r#type` instead in the generated code.

Keep up the great work!

billy1624:
Resolve #217


billy1624:
r#self and r#Self are not possible. 🦀 won't compile loll

tyt2y3:
okay so https://internals.rust-lang.org/t/raw-identifiers-dont-work-for-all-identifiers/9094
then how can we handle `self`, `crate`, `super`?

billy1624:
See d6b53abd9500bd02dd00f4121968111820060422

JellyWX:
looks that `macro` is missing from the identifier list here

billy1624:
Hey @JellyWX, do you mean `macro` isn't included in the list of keyword?

JellyWX:
Yes, seems so. Just ran into the generation issue with a table named "macro"

billy1624:
I think that's another problem. Could you please elaborate it on a new issue? Thanks!!

PR #240, opened 2021-10-11
Add derive `DeriveIntoActiveModel` and `IntoActiveValue` trait

tqwewe:
Creates a new derive macro `DeriveIntoActiveModel` for implementing [`IntoActiveModel`](https://docs.rs/sea-orm/0.2.6/sea_orm/entity/trait.IntoActiveModel.html) on structs. This is useful for creating your own struct with only partial fields of a model, for example an insert struct, or update struct.

```rust
// Define regular model as usual
#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel)]
#[sea_orm(table_name = "users")]
pub struct Model {
    pub id: Uuid,
    pub updated_at: DateTimeWithTimeZone,
    pub email: String,
    pub password: String,
    pub full_name: Option<String>,
    pub phone: Option<String>,
}

// Create a new struct with some fields omitted
#[derive(DeriveIntoActiveModel)]
pub struct NewUser {
    // id, created_at and updated_at are omitted from this struct, and will always be `ActiveValue::unset`
    pub email: String,
    pub password: String,
    pub full_name: String, // Full name is usually optional, but it can be required here
    pub phone: Option<String>, // Option implements `IntoActiveValue`, and when `None` will be `unset`
}

#[derive(DeriveIntoActiveModel)]
pub struct UpdateUser {
    pub phone: Option<Option<String>>, // Option<Option<T>> allows for Some(None) to update the column to be NULL
}

fn main() {
    let new_user = NewUser {
        email: "someone@domain.com".to_string(),
        password: "test".to_string(),
        full_name: "John Doe".to_string(),
        phone: None,
    };
    
    let active_model = new_user.into_active_model();
    // active_model can now be used in queries...
}
```

Additionally, an attribute `active_model` is available in case your active model struct has a custom name.
```rust
#[derive(DeriveIntoActiveModel)]
#[sea_orm(active_model = MyActiveModel)]
pub struct NewProduct {
    // ...
}
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -144,6 +144,12 @@ jobs:
           args: >
             --all
 
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: >
+            --manifest-path sea-orm-rocket/Cargo.toml
+
   cli:
     name: CLI
     runs-on: ${{ matrix.os }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,12 +30,13 @@ futures-util = { version = "^0.3" }
 log = { version = "^0.4", optional = true }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.2.6", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.17.0", features = ["thread-safe"] }
+sea-query = { version = "^0.17.1", features = ["thread-safe"] }
 sea-strum = { version = "^0.21", features = ["derive", "sea-orm"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1", optional = true }
 sqlx = { version = "^0.5", optional = true }
 uuid = { version = "0.8", features = ["serde", "v4"], optional = true }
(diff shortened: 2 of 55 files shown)


Acidic9:
> Is it better if the `active_model` attribute is a string?
> I discovered it is not possible to have `::` in the value
> 
> Can't compile
> ```rust
> #[derive(DeriveIntoActiveModel)]
> #[sea_orm(active_model = my_mod::MyActiveModel)]
> pub struct NewProduct {
>     // ...
> }
> ```
> 
> Should work
> ```rust
> #[derive(DeriveIntoActiveModel)]
> #[sea_orm(active_model = "my_mod::MyActiveModel")]
> pub struct NewProduct {
>     // ...
> }
> ```

Oh I wasn't aware that wouldn't work. I guess it's better to use a string then. I can create another PR to change this if you'd like?

tyt2y3:
> Oh I wasn't aware that wouldn't work. I guess it's better to use a string then. I can create another PR to change this if you'd like?

Sure, that'd be appreciated!
(And it is more in line with DeriveRelation)

PR #245, opened 2021-10-15
Use backed-up primary key as last insert id before getting it from driver

### LINKED ISSUE #244: [BUG][SQLITE] Failed to find inserted item
Hello!

Thank you for the great library!

I've run into a small problem. It seems there is a problem when inserting new data into a `Sqlite` database, in a table which primary key is a string.

I've tried using `WITHOUT ROWID` in the table definition, but no luck.

From a debug statement I added in a local version of `sea-orm`, it seems the "last inserted id" is defaulting to either `"0"` or `"1"`

Here is a minimal working example:

`main.rs`

```rust
use anyhow::Error;
use sea_orm::entity::{Set, Unset};

mod database;

#[tokio::main]
async fn main() -> Result<(), Error> {
    let database = database::get_connection().await?;

    let new_repository = database::entities::repository::ActiveModel {
        id: Set("some_id".to_string()),
        name: Set("some_name".to_string()),
        owner: Set("some_owner".to_string()),
        description: Unset(None)
    };
    let res = new_repository.insert(&database).await?;
    println!("{:#?}", res);
    Ok(())
}
```

`Cargo.toml` (truncated)
```toml
[dependencies]
tokio = { version = "1", features = [ "full" ] }
anyhow = "1"
sea-orm = { version = "^0", features = [ "sqlx-sqlite", "runtime-tokio-rustls", "macros" ], default-features = false }
```

The migration file

```sql
CREATE TABLE IF NOT EXISTS repository (
    id TEXT PRIMARY KEY,
    owner TEXT NOT NULL,
    name TEXT NOT NULL,
    description TEXT
);
```

The entity file
```rust

#[sea_orm(table_name = "repository")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: String,
    pub owner: String,
    pub name: String,
    pub description: Option<String>
}

pub enum Relation {
}

```

billy1624:
- Resolve #244

### DIFF
--- a/tests/common/bakery_chain/mod.rs
+++ b/tests/common/bakery_chain/mod.rs
@@ -7,6 +7,7 @@ pub mod customer;
 pub mod lineitem;
 pub mod metadata;
 pub mod order;
+pub mod repository;
 
 pub use super::applog::Entity as Applog;
 pub use super::baker::Entity as Baker;
@@ -17,3 +18,4 @@ pub use super::customer::Entity as Customer;
 pub use super::lineitem::Entity as Lineitem;
 pub use super::metadata::Entity as Metadata;
 pub use super::order::Entity as Order;
+pub use super::repository::Entity as Repository;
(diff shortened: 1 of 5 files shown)


tyt2y3:
We should add #244 as a test case?

billy1624:
> We should add #244 as a test case?

Sure!

PR #258, opened 2021-10-19
Enumeration

### LINKED ISSUE #252: Enumeration
This issue is to outline the design of how to allow custom Enums to be used in SeaORM Entity.

It will be supported in 3 steps:

- [x] Rust enum support (but serialize to string or integer)
- [x] Mapping to a database native enum
- [ ] Schema discovery & code generation

A proposed trait `ActiveEnum` (or `DbEnum` or `OrmEnum` etc) would be:

```rust
trait ActiveEnum {
    type Value: ValueType;

    fn to_value(&self) -> Self::Value;
    fn try_from_value(v: &Self::Value) -> Result<Self, DbErr>;
    fn db_type() -> ColumnDef;
}
```

A sample model would be like:

````rust
enum Category {
    Big,
    Small,
}

struct Model {
    category: Category;
}

impl ActiveEnum for Category {
    type Value = String;

    fn to_value(&self) -> Self::Value {
        match self {
            Big => "B",
            Small => "S",
        }.to_owned()
    }
    fn from_value(v: &Self::Value) -> Result<Self, DbErr> {
        match v.as_ref {
            "B" => Ok(Self::Big),
            "S" => Ok(Self::Small),
            _ => Err(DbErr::TypeErr(format!("unexpected value for Category: {}", v))),
        }
    }
    fn db_type() -> ColumnDef {
        ColumnType::String(Some(1)).def
    }
}
````

Such a trait should be all it takes for using a custom enum (we need to eliminate all boilerplate as in https://github.com/SeaQL/sea-orm/discussions/196#discussioncomment-1379259). And it will be converted automatically during Select, Insert and Update.

We can also make a `DeriveActiveEnum` macro to make it easier:

```rust
#[derive(DeriveActiveEnum)]
#[sea_orm(db_type = "String(Some(1))")]
enum Category {
    #[sea_orm(string_value = "B")]
    Big,
    #[sea_orm(string_value = "S")]
    Small,
}
```

(because of the excess `String("B")`, we'd better divide into string_value and num_value)

After some discussion with @Acidic9, to support native enum in Postgres, we basically have to cast everywhere:

```SQL
SELECT CAST("category" AS text) from "mytable";
INSERT INTO "mytable" ("category") VALUES (CAST($1 AS CategoryEnum));
UPDATE "mytable" SET "category" = CAST($1 AS CategoryEnum);
```

For MySQL, casting is not needed.

Ref: #196

billy1624:
- Resolve #252
- Related to https://github.com/SeaQL/sea-query/issues/131
- Depends on SeaQL/sea-query#170


billy1624:
Btw... I will try to document all newly added types starting from this PR. Plz check the docs and give me some feedback on it @tyt2y3

billy1624:
Ready for review and merge @tyt2y3

tyt2y3:
Sorry for creating conflict

billy1624:
No worries, I will resolve it

src/error.rs
```diff
@@ -11,6 +11,8 @@ pub enum DbErr {
     RecordNotFound(String),
     /// A custom error
     Custom(String),
+    /// Error occurred while parsing value into [ActiveEnum](crate::ActiveEnum)
```
tyt2y3, line 14: Just a slight change, I think it is a more general error, which can be documented as 'Error occurred while parsing value into target type'

tests/common/features/active_enum.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[sea_orm(num_value = 0)]
+    Black,
+    #[sea_orm(num_value = 1)]
+    White,
+}
+
+#[derive(Debug, Clone, PartialEq, DeriveActiveEnum)]
+#[sea_orm(
+    rs_type = "String",
+    db_type = r#"Enum("tea".to_owned(), vec!["EverydayTea".to_owned(), "BreakfastTea".to_owned()])"#
```
tyt2y3, line 39: Oh, this shouldn't be the intended usage?

billy1624: This allow create_enum_from_entity and create_table_from_entity got the required info and do their job. I'm thinking how to pull the enum variants (Vec<String>) part out.

tyt2y3: Yup it certainly looks clumsy

billy1624: Agree

tyt2y3: To avoid duplication of information, we should be able to delegate the value definitions to the actual enum definition.

billy1624: Probably storing `Box<Trait>`

billy1624: > Probably storing Box<Trait>

I think it's not possible...

```
error[E0038]: the trait `active_enum::ActiveEnumIden` cannot be made into an object
   --> src/entity/column.rs:68:17
    |
68  |     EnumTry(Box<dyn ActiveEnumIden>),
    |                 ^^^^^^^^^^^^^^^^^^ `active_enum::ActiveEnumIden` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> src/entity/active_enum.rs:112:27
    |
112 | pub trait ActiveEnumIden: Clone {
    |           --------------  ^^^^^ ...because it requires `Self: Sized`
    |           |
    |           this trait cannot be made into an object...
```

While we have to implement `Clone` for the trait, because...

```rust
#[derive(Debug, Clone, PartialEq)]
pub enum ColumnType { ... }
```

tyt2y3: If we require ActiveEnum to extend Iterable, then we can provide a default impl that returns all values.

```rust
fn values() -> Vec<String> {
    Self::IntoIter()...
}
```

Then the ColumnDef simply call this at runtime.

billy1624: Sounds like a good plan

billy1624: See bb78a1d

tyt2y3:
I think we are pretty much done

https://github.com/SeaQL/sea-orm/blob/bb78a1d7097be2fc6cecceb547580ff8fd47250b/tests/common/features/active_enum.rs#L1-L43

https://github.com/SeaQL/sea-orm/blob/bb78a1d7097be2fc6cecceb547580ff8fd47250b/tests/active_enum_tests.rs#L1-L68

@Acidic9 @nicoulaj request for comment 👀

Acidic9:
Ooh I really like the look of this. I had a bried look at the linked code snippets you shared @tyt2y3 and it looks really explicit. I like how you can use an enum as an integer too!

nicoulaj:
Just tried it on a complex project and it mostly worked flawlessly, awesome work !

Two things I noticed:

- When using database enum, `ColumnTrait::eq` does not seem to cast the value, it generates `WHERE mycol = ?` and the query is invalid as a consequence
- The new macros generate code with `use sea_query` instead of `use sea_orm::sea_query`, forcing users to add an explicit dependency on sea-query

src/entity/active_enum.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    /// Try to convert the corresponding value into enum variant.
+    fn try_from_value(v: &Self::Value) -> Result<Self, DbErr>;
+
+    /// Get the database column definition of this active enum.
+    fn db_type() -> ColumnDef;
+
+    /// Get the name of all enum variants
+    fn values() -> Vec<Self::Value> {
+        Self::iter().map(|s| s.to_value()).collect()
```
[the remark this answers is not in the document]
billy1624: It won't loll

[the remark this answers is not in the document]
billy1624:
```
error[E0631]: type mismatch in function arguments
   --> src/entity/active_enum.rs:125:26
    |
115 |     fn to_value(&self) -> Self::Value;
    |     ---------------------------------- found signature of `for<'r> fn(&'r Self) -> _`
...
125 |         Self::iter().map(Self::to_value).collect()
    |                          ^^^^^^^^^^^^^^ expected signature of `fn(Self) -> _`

error[E0599]: the method `collect` exists for struct `std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>`, but its trait bounds were not satisfied
   --> src/entity/active_enum.rs:125:42
    |
125 |         Self::iter().map(Self::to_value).collect()
    |                                          ^^^^^^^ method cannot be called on `std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>` due to unsatisfied trait bounds
    |
   ::: /Users/billy/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/adapters/map.rs:61:1
    |
61  | pub struct Map<I, F> {
    | -------------------- doesn't satisfy `_: Iterator`
    |
    = note: the following trait bounds were not satisfied:
            `<for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value} as FnOnce<(Self,)>>::Output = _`
            which is required by `std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>: Iterator`
            `for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}: FnMut<(Self,)>`
            which is required by `std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>: Iterator`
            `std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>: Iterator`
            which is required by `&mut std::iter::Map<<Self as IntoEnumIterator>::Iterator, for<'r> fn(&'r Self) -> <Self as active_enum::ActiveEnum>::Value {<Self as active_enum::ActiveEnum>::to_value}>: Iterator`

Some errors have detailed explanations: E0599, E0631.
For more information about an error, try `rustc --explain E0599`.
error: could not compile `sea-orm` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed
```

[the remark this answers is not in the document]
billy1624: This works lol

```rust
pub trait ActiveEnum: Sized + Iterable {
    
    ...

    /// Convert enum variant into the corresponding value.
    fn to_value(&self) -> Self::Value;

    ...

    /// Convert an owned enum variant into the corresponding value.
    fn into_value(self) -> Self::Value {
        Self::to_value(&self)
    }

    /// Get the name of all enum variants
    fn values() -> Vec<Self::Value> {
        Self::iter().map(Self::into_value).collect()
    }
}
```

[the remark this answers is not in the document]
billy1624: See 2b841b1b5d5bcea74ffc0cd99951c7f9a0955be5

billy1624:
> * The new macros generate code with `use sea_query` instead of `use sea_orm::sea_query`, forcing users to add an explicit dependency on sea-query

Nice catch! Fixed on f64f1e9216eaee57822ef2071960ff27072d06ec

billy1624:
> * When using database enum, `ColumnTrait::eq` does not seem to cast the value, it generates `WHERE mycol = ?` and the query is invalid as a consequence

That's bad. See da9f3ea60ca8b56013270dff07bfee5eed4fd9c4. Thanks!! @nicoulaj

billy1624:
After second thought, it only make sense to cast string to enum type inside `=` & `!=` comparison. See 10f3de0f9da0738903c6cb962d11e371f1d2f278

tyt2y3 (approved):
It's been a long road

PR #263, opened 2021-10-20
CI cleaning

AngelOnFira:
This PR is to clean up CI a bit, and separate out some tests into their own files. It also adds a RustFmt check.

Specifically, all of the tests for postgres/mysql/sqlite were all moved to respective test files (postgres.yml for example). All of the general unit tests, clippy, and other generic items were added to a `rust-ci.yml` file.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,9 +11,8 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-
-  clippy:
-    name: Clippy
+  clippy-fmt:
+    name: Clippy + Fmt
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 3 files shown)


tyt2y3:
We can give it a try. But let's put rust fmt before everything so it fails fast.

The current workflow also have two stages, so if compilation fails, subsequent tests would abort.

I tried to look for a combined status badge for all workflows, but couldn't.

billy1624:
So we basically keep the unified old workflow and add a `cargo fmt` stage before it :P

AngelOnFira:
I guess it wouldn't be the best, you could see the rendered version [here](https://github.com/SeaQL/sea-orm/blob/88a0c1fd585add40cef72da9509d66fab4eaad7b/README.md).

image

PR #276, opened 2021-10-28
Soft Delete - 1

### LINKED ISSUE #220: [Feature Request] Soft Delete
### Use case 

* Sometimes we can't always perform hard deletions in production.
* This is what most ORMs do

### The Ideal way

* Respect the `deleted_at` column (`pub deleted_at: Option<DateTimeWithTimeZone>`)
* All `deleted_at` checkings should run behind the scenes instead manually adding `AND deleted_at IS NULL`
* `deleted_at` columns should be indexed for performance reasons

## Delete

```rust
let res: DeleteResult = fruit.delete(db).await?; // ID = 10
```

Should generate the SQL below

```sql
-- Soft Delete - If the column deleted_at exists
UPDATE `fruit` SET `deleted_at`="2020-10-29 10:23" WHERE id = 10

-- Hard Delete - If the column deleted_at doesn't exist
DELETE FROM `fruit` WHERE `fruit`.`id` = 10'
```

## Select

```rust
let cheese: Option<cake::Model> = Cake::find_by_id(1).one(db).await?;
```

Should generate the SQL below

```sql
-- If the column deleted_at exists
SELECT * FROM cake WHERE id = 1 AND deleted_at IS NULL

-- If the column deleted_at doesn't exist
SELECT * FROM cake WHERE id = 1
```

## Golang GORM

PS: I am a Go developer most of the time.

GORM, the popular ORM for Go, has defined a [default struct](https://gorm.io/docs/models.html#gorm-Model) called `gorm.Model`.

```go
// gorm.Model definition
type Model struct {
  ID        uint           `gorm:"primaryKey"`
  CreatedAt time.Time
  UpdatedAt time.Time
  DeletedAt gorm.DeletedAt `gorm:"index"`
}
```

And any struct with `gorm.Model` in it will automatically become soft-delete aware.

```go
type User struct {
  gorm.Model
  Name       string
  Phone      string
}
```

If one really needs to perform hard deletion, here's how in GORM:

```go
db.Delete(&order)
// UPDATE orders SET deleted_at="2020-10-29 10:23" WHERE id = 10;

db.Unscoped().Delete(&order)
// DELETE FROM orders WHERE id=10;
```

billy1624:
- Resolve #220
- Depends on SeaQL/sea-query#175


tests/common/features/soft_delete_many_to_many.rs
```diff
@@ -0,0 +1,742 @@
...
+                    "INNER JOIN `soft_delete_child` ON `soft_delete_child`.`id` = `soft_delete_junction`.`child_id`",
+                    "WHERE `soft_delete`.`deleted_at` IS NULL",
+                    // FIXME: No way to know if the junction have soft delete enabled with only RelationDef on hand
+                    // "AND `soft_delete_junction`.`deleted_at` IS NULL",
+                    "AND `soft_delete_child`.`deleted_at` IS NULL",
+                    "AND `soft_delete_child`.`id` = 11",
+                ]
+                .join(" ")
+            );
+        }
```
billy1624, line 187: Currently, we have no way to tell if the junction table have soft delete enabled or not with `Related<T>`.

tests/common/features/soft_delete_many_to_many.rs
```diff
@@ -0,0 +1,742 @@
...
+                    "LEFT JOIN `soft_delete_junction` AS `r0` ON `soft_delete_child`.`id` = `r0`.`child_id`",
+                    "LEFT JOIN `soft_delete` AS `r1` ON `r0`.`parent_id` = `r1`.`id`",
+                    "WHERE `soft_delete_child`.`deleted_at` IS NULL",
+                    // FIXME: No way to know if the junction have soft delete enabled with only RelationDef on hand
+                    // "AND `r0`.`deleted_at` IS NULL",
+                    "AND `r1`.`deleted_at` IS NULL",
+                ]
+                .join(" ")
+            );
+        }
```
billy1624, line 240: Also, we can't tell if all the intermediate tables have soft delete enabled for `Linked` relation.

ericsampson:
How is the soft-delete column indexed?  I took a quick scan through the PR, but as I’m not familiar with this code base I couldn’t find this information. Thanks!

billy1624:
Column indexing is not included in the scope of this PR

ericsampson:
is column indexing planned to be addressed later?
Without having that option, this feature can have a significant performance impact on every operation that now needs to filter on this column where it didn't before.

billy1624:
Hey @ericsampson,

You can mark the column as `indexed` here. But this would not automatically make your column to be indexed in database.
https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-structure#additional-properties

By marking a column as `indexed` will help you generate SQL statement to create the table in database together with column indexes.
https://www.sea-ql.org/SeaORM/docs/generate-database-schema/create-table

tyt2y3:
Seems like we are too far behind master

billy1624:
> Seems like we are too far behind master

Yes, but the conflict is minimal

ericsampson:
@billy1624 thank you for the pointer, apologies for the noise, I'm new to these repos! Cheers

tyt2y3:
We'd probably want to revisit this now

billy1624:
> We'd probably want to revisit this now

Sure!!

billy1624:
I'll work on this from scratch on another branch

PR #287, opened 2021-11-02
Support `Vec<u8>` primary key

### LINKED ISSUE #286: Support for BYTEA Postgres primary keys
Currently any column marked as a primary key needs to satisfy `TryFromU64`.

A BYTEA (`Vec<u8>`) does not satisfy this (and, I don't think ever could).

Would it be possible to add support for this type of primary key?  We use it a lot in our application

billy1624:
- Resolve #286

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -539,5 +539,7 @@ macro_rules! try_from_u64_string {
 
 try_from_u64_string!(String);
 
+try_from_u64_err!(Vec<u8>);
+
 #[cfg(feature = "with-uuid")]
 try_from_u64_err!(uuid::Uuid);
--- a/tests/byte_primary_key_tests.rs
+++ b/tests/byte_primary_key_tests.rs
@@ -23,7 +23,7 @@ pub async fn create_and_update(db: &DatabaseConnection) -> Result<(), DbErr> {
     use common::features::byte_primary_key::*;
 
     let model = Model {
-        id: vec![1],
+        id: vec![1, 2, 3],
         value: "First Row".to_owned(),
     };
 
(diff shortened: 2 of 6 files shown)


billy1624:
Hey @tyt2y3, I was wondering why `ColumnType::Binary` is defined as without length in sea-orm?

Because of that, I can't seems to get around [this](https://github.com/SeaQL/sea-orm/runs/4077110887?check_suite_focus=true#step:6:189).

```
Error: Exec("error returned from database: 1170 (42000): BLOB/TEXT column 'id' used in key specification without a key length")
```

tyt2y3:
The `BYTEA` type in Postgres does not have a length.

tyt2y3:
@robot-head can you share your schema definition? I bet the index should have a key length

billy1624:
> Hey @tyt2y3, I was wondering why `ColumnType::Binary` is defined as without length in sea-orm?
> 
> Because of that, I can't seems to get around [this](https://github.com/SeaQL/sea-orm/runs/4077110887?check_suite_focus=true#step:6:189).
> 
> ```
> Error: Exec("error returned from database: 1170 (42000): BLOB/TEXT column 'id' used in key specification without a key length")
> ```

Btw... this only failed on MySQL & MariaDB. Postgres is working perfectly

tyt2y3:
Well, can we default `1024` it inside SeaORM?

billy1624:
> Well, can we default `1024` it inside SeaORM?

For MySQL & MariaDB only and leave Postgres untouched?

tyt2y3:
Edited.

billy1624:
> Okay, then we should use `sea_query::ColumnType::Binary(Some(1024))` explicitly for the test case.

That would work but how about ordinary user that want to use [`create_table_from_entity`](https://docs.rs/sea-orm/0.3.1/sea_orm/schema/struct.Schema.html#method.create_table_from_entity) on MySQL & MariaDB? They will fall into this trap as well if they happen to use `Vec<u8>` (`blob` db type) as primary key.

tyt2y3:
Looks good. Anything missing? I am about to make a minor release.

billy1624:
Go for it

PR #292, opened 2021-11-05
Returning

### LINKED ISSUE #183: Support returning clause to avoid database hits
Hello,

I took a look at discussions and haven't found nothing, so I am opening this issue to discuss something important.

We need to  support returning clauses in updates and insert clauses.

## Rationale

Sometimes, we need to perform batch operations, by means of multiple inserts or updates which affect more than one record and, at the same time, have the newly inserted / the updated records back in the same database operation.

This is needed, for example, when you need to insert records and have their auto incremented generated ids back to store somewhere for latter use, or when you need to update one or more records and, at the same time, get back columns you didn't have when you performed the update operation.

The way sql handles this is by using a returning clause on the query, making the database, in one operation, perform inserts / updates and also returning information you need instead of forcing you to make a latter select.

## Proposal

We propose an extension to the current sea-orm api.

## update_many

The UpdateMany struct should have another method, called returning.

Kind of code below

```
pub fn returning<T>(mut self, record_info: RecordInfo) -> Self
```

RecordInfo should be an enum like

```
pub enum RecordInfo {
  Columns(Vec<Entity::Columns>,
  FullRecord,
  PrimaryKey,
}
```

The returning function would generate a returning sql clause, as follows:

```
let returning_sql = match RecordInfo {
  Columns(c) => c.map(| col | col.to_string()).join(", "),
FullRecord => "*".into(),
  PrimaryKey: // build list of columns making part of the primary key
}
```

The insert_many would go the very same path.

The ExecResult would need to ave a vector of ActiveModels with the appropriate columns set for each returned record.

What do you think?

I could try to implement that, I would need some mentoring but I am now more focused in stablishing the api, latter we can think in the implementation.

billy1624:
- Resolve #183


billy1624:
Hey @tyt2y3, checkout the support of returning for Postgres.

src/executor/update.rs
```diff
@@ -78,17 +80,45 @@ where
...
+    mut query: UpdateStatement,
     model: A,
     db: &'a C,
 ) -> Result<A, DbErr>
 where
+    <A::Entity as EntityTrait>::Model: IntoActiveModel<A>,
     A: ActiveModelTrait,
     C: ConnectionTrait<'a>,
 {
-    Updater::new(query).check_record_exists().exec(db).await?;
```
billy1624, line 90: The old behavior will produce a error if user try to update some non-existing rows.

src/executor/update.rs
```diff
@@ -78,17 +80,45 @@ where
...
+                <A::Entity as EntityTrait>::Column::iter()
+                    .map(|c| c.into_column_ref())
+                    .collect(),
+            ));
+            SelectorRaw::<SelectModel<<A::Entity as EntityTrait>::Model>>::from_statement(
+                db_backend.build(&query),
+            )
+            .one(db)
+            .await?
+        }
```
billy1624, line 106: But now, it seems impossible to produce the same error using returning.

billy1624: And that's why we have this error [here](https://github.com/SeaQL/sea-orm/runs/4118287744?check_suite_focus=true#step:6:201)
```
thread 'main' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 Err(
<    Exec(
<        "Failed to find inserted item",
>    RecordNotFound(
>        "None of the database rows are affected",
     ),
 )

', tests/byte_primary_key_tests.rs:48:5
```

billy1624: Resolved on 6238736

billy1624:
Got this error on MariaDB 10.5+

- The error is from SQLx
  - launchbadge/sqlx#1530
- https://github.com/SeaQL/sea-orm/runs/4147966480#step:6:211

```
[2021-11-09T04:31:54Z DEBUG sea_orm::driver::sqlx_mysql] INSERT INTO `active_enum` (`category`, `color`, `tea`) VALUES (NULL, NULL, NULL) RETURNING `id`, `category`, `color`, `tea`
[2021-11-09T04:31:54Z INFO  sqlx::query] INSERT INTO `active_enum` (`category`, …; rows: 1, elapsed: 1.526ms
    
    INSERT INTO
      `active_enum` (`category`, `color`, `tea`)
    VALUES
      (?, ?, ?) RETURNING `id`,
      `category`,
      `color`,
      `tea`
    
Error: Query("no column found for name: id")
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/test/src/lib.rs:194:5
```

billy1624:
Now I'm stuck on the SQLx problem... I'm thinking maybe we just support returning for Postgres at this stage. We can leave MariaDB & Sqlite as-is, i.e. one more round trip per insert / update, if it involve updating sqlx-core.

billy1624:
The issue not only affecting MariaDB 10.5+ but SQLite 3.35.0+

- MariaDB: https://github.com/SeaQL/sea-orm/runs/4150761196
- SQLite: https://github.com/SeaQL/sea-orm/runs/4150793637

billy1624:
Just open another issue on SQLx
- launchbadge/sqlx#1531

billy1624:
As discussed, we would only support `RETURNING` on Postgres at this stage. And leaving MySQL & SQLite for later.

billy1624:
Ready for review & merge @tyt2y3

src/executor/update.rs
tyt2y3: What's the new behavior then?

src/entity/base_entity.rs
```diff
@@ -413,7 +413,7 @@ pub trait EntityTrait: EntityName {
     /// assert_eq!(
     ///     db.into_transaction_log(),
     ///     vec![Transaction::from_sql_and_values(
-    ///         DbBackend::Postgres, r#"UPDATE "fruit" SET "name" = $1 WHERE "fruit"."id" = $2 AND "fruit"."name" LIKE $3"#,
+    ///         DbBackend::Postgres, r#"UPDATE "fruit" SET "name" = $1 WHERE "fruit"."id" = $2 AND "fruit"."name" LIKE $3 RETURNING "id", "name", "cake_id""#,
```
tyt2y3, line 416: It seems to make sense if we add a more example for MySQL

src/executor/update.rs
billy1624: I manage to keep this behavior.

src/executor/update.rs
```diff
@@ -78,17 +80,61 @@ where
...
+                .one(db)
+                .await?;
+            // If we got `None` then we are updating a row that does not exist.
+            match found {
+                Some(model) => Ok(model.into_active_model()),
+                None => Err(DbErr::RecordNotFound(
+                    "None of the database rows are affected".to_owned(),
+                )),
+            }
+        }
```
billy1624, line 120: This is how I reproduce the `DbErr::RecordNotFound` error.

src/entity/base_entity.rs
billy1624: See 4d44827

src/entity/base_entity.rs
```diff
@@ -299,15 +299,56 @@ pub trait EntityTrait: EntityName {
...
     /// #
     /// # Ok(())
     /// # });
     ///
     /// assert_eq!(
     ///     db.into_transaction_log(),
     ///     vec![Transaction::from_sql_and_values(
-    ///         DbBackend::Postgres, r#"INSERT INTO "cake" ("name") VALUES ($1) RETURNING "id""#, vec!["Apple Pie".into()]
+    ///         DbBackend::Postgres,
+    ///         r#"INSERT INTO "cake" ("name") VALUES ($1) RETURNING "id""#,
```
tyt2y3, line 310: I don't get it. Shouldn't it return all columns here?

tyt2y3, line 310: I don't get it. Shouldn't it return all columns here?

src/entity/base_entity.rs
```diff
@@ -413,10 +498,69 @@ pub trait EntityTrait: EntityName {
     /// assert_eq!(
     ///     db.into_transaction_log(),
     ///     vec![Transaction::from_sql_and_values(
-    ///         DbBackend::Postgres, r#"UPDATE "fruit" SET "name" = $1 WHERE "fruit"."id" = $2 AND "fruit"."name" LIKE $3"#,
+    ///         DbBackend::Postgres,
+    ///         r#"UPDATE "fruit" SET "name" = $1 WHERE "fruit"."id" = $2 AND "fruit"."name" LIKE $3 RETURNING "id", "name", "cake_id""#,
```
tyt2y3, line 502: As in UPDATE?

src/database/db_connection.rs
```diff
@@ -267,6 +267,11 @@ impl DbBackend {
             Self::Sqlite => Box::new(SqliteQueryBuilder),
         }
     }
+
+    /// Check if the database supports `RETURNING` syntax on insert and update
+    pub fn support_returning(&self) -> bool {
+        matches!(self, Self::Postgres)
```
tyt2y3, line 273: If self is a `Mock`, we should respect the `get_database_backend` of it.

src/entity/base_entity.rs
billy1624: Actually, this is unchanged. I just added a new line after `DbBackend::Postgres,`. For Postgres, it use `RETURNING` to get the last inserted id.

src/entity/base_entity.rs
billy1624: Is related to Postgres insert?

src/entity/base_entity.rs
billy1624: While other db get the last insert id out of `execute`.

src/database/db_connection.rs
billy1624: > If self is a Mock, we should respect the get_database_backend of it.

Yes! We should

> I still don't understand why the behaviour of insert and update differs though.

Does my answer clear enough?

src/entity/base_entity.rs
tyt2y3: Here the returning clause is returning all columns, which is different from the insert case where it only returns 'I'd'

src/entity/base_entity.rs
tyt2y3: Oh so we have 2 different insert implement here?

billy1624: > Oh so we have 2 different insert implement here?

Correct! And this happens on 0.3.x as well

src/entity/base_entity.rs
billy1624: This is `update`'s docs.

> Here the returning clause is returning all columns, which is different from the insert case where it only returns 'id'

Yes! Any issue here?
Why don't we write it as `RETURNING *`?

src/entity/base_entity.rs
tyt2y3: Oh then can I also locate the other test case?

billy1624: Please find it below. I have added a new review

src/entity/base_entity.rs
tyt2y3: No issue here. I'd just want to see a consistent behaviour with insert. I could not find the counterpart test case.

src/entity/active_model.rs
```diff
@@ -147,20 +147,16 @@ pub trait ActiveModelTrait: Clone + Debug {
         C: ConnectionTrait<'a>,
     {
         let am = ActiveModelBehavior::before_save(self, true)?;
-        let res = <Self::Entity as EntityTrait>::insert(am).exec(db).await?;
-        let found = <Self::Entity as EntityTrait>::find_by_id(res.last_insert_id)
-            .one(db)
+        let am = <Self::Entity as EntityTrait>::insert(am)
```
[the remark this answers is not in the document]
billy1624: Yep!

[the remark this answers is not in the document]
billy1624: Test cases added

tyt2y3 (approved):
Nice

PR #306, opened 2021-11-09
Add `PaginatorTrait` and `CountTrait` for more constrains

YoshieraHuang:
When I build my web api using sea-orm, I want to constrain my select type to have `paginate()` method. So I added a `PaginatorTrait` and move the `paginate()` method to this trait. Meanwhile `count()` method of select type depends on `paginate()` method, but not all `paginate`able select has this method. So I added another `CountTrait` trait.

My use case of `PaginatorTrait` is like this:
```rust
// Data transfer object, aka dto
#[derive(Debug, Serialize)]
pub struct Paginated<T> {
    items: Vec<T>,
    page_size: usize,
    page_index: usize,
    num_pages: usize,
    num_items: usize,
    has_previous: bool,
    has_next: bool,
}

impl<T> Paginated<T>
{
    pub async fn query_with_map<'db, P, S, F>(
        db: &'db DatabaseConnection,
        paginate: P,
        page_size: usize,
        page_index: usize,
        predicate: F)
    -> Result<Self, DbErr>
    where
        P: PaginatorTrait<'db, DatabaseConnection, Selector = S>,
        S: SelectorTrait + 'db,
        F: FnMut(<S as SelectorTrait>::Item) -> T
    {
        let paginator = paginate.paginate(db, page_size);
        let original_items = paginator.fetch_page(page_index - 1).await?;
        let items = original_items.into_iter().map(predicate).collect();
        let num_pages = paginator.num_pages().await?;
        let num_items = paginator.num_items().await?;
        let has_previous = num_pages > 0 && page_index > 1;
        let has_next = num_pages > 0 && page_index < (num_pages - 1);
        Ok(Paginated {
            items,
            page_size,
            page_index,
            num_pages,
            num_items,
            has_previous,
            has_next
        })
    }
}
```

These traits may be used not so often but needed for me. It will be fine if this PR is merged.

By the way, if this PR will be merged, feel free to modify the docs and variable names, cause i'm not a native english speaker.

PR #307, opened 2021-11-10
Continue `PaginatorTrait`

billy1624:
## Info
- Continue #306

### DIFF
--- a/src/entity/mod.rs
+++ b/src/entity/mod.rs
@@ -117,3 +117,6 @@ pub use model::*;
 // pub use prelude::*;
 pub use primary_key::*;
 pub use relation::*;
+
+// Imports paginator utility by default making it backward compatible.
+pub use crate::PaginatorTrait;
--- a/src/entity/mod.rs
+++ b/src/entity/mod.rs
@@ -117,6 +117,3 @@ pub use model::*;
 // pub use prelude::*;
 pub use primary_key::*;
 pub use relation::*;
-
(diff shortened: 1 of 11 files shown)


src/entity/mod.rs
```diff
@@ -117,3 +117,6 @@ pub use model::*;
 // pub use prelude::*;
 pub use primary_key::*;
 pub use relation::*;
+
+// Imports paginator utility by default making it backward compatible.
+pub use crate::PaginatorTrait;
```
tyt2y3, line 122: I think it should belong to `query::*` instead of `entity::*`

billy1624: Changed on f975223

src/query/mod.rs
```diff
@@ -23,3 +23,6 @@ pub use update::*;
 pub use util::*;
 
 pub use crate::{ConnectionTrait, InsertResult, Statement, UpdateResult, Value, Values};
+
+// Imports paginator utility by default making it backward compatible.
+pub use crate::PaginatorTrait;
```
tyt2y3, line 28: Not nitpicking, but it actually can fit in the above statement.

billy1624: Can ah, just wanna put an explicit comment on it. I can remove the comment if it's unnecessary

tyt2y3: It's unnecessary.

tyt2y3 (approved):

PR #309, opened 2021-11-11
Refactor `Schema`

billy1624:


tyt2y3 (approved):
Yeah, the breaking change is unavoidable, and it seems to be a better API

Issue #322, opened 2021-11-16
Cannot derive `DeriveIntoActiveModel` for fields wrapped with `Option`


YoshieraHuang:
I meet compiling error when I use `DeriveIntoActiveModel` to convert the struct to `ActiveModel`. The struct has only part of fields and these fields are wrapped with `Option`. Here is an example script:
```rust
use sea_orm::entity::prelude::*;
use serde::{Deserialize, Serialize};

#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Serialize, Deserialize)]
#[sea_orm(table_name = "products")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub name: String,
    pub price: f64
}

impl ActiveModelBehavior for ActiveModel {}

#[derive(DeriveIntoActiveModel, Deserialize)]
pub struct UpdateProducts {
    pub name: Option<String>,
    pub price: Option<f64>
}
```

And here is the detailed compiling error:
```
the trait bound `ActiveValue<String>: From<ActiveValue<std::option::Option<String>>>` is not satisfied
the following implementations were found:
  <ActiveValue<std::option::Option<V>> as From<ActiveValue<V>>>
required because of the requirements on the impl of `Into<ActiveValue<String>>` for `ActiveValue<std::option::Option<String>>`
```
```
the trait bound `ActiveValue<f32>: From<ActiveValue<std::option::Option<f32>>>` is not satisfied
the following implementations were found:
  <ActiveValue<std::option::Option<V>> as From<ActiveValue<V>>>
required because of the requirements on the impl of `Into<ActiveValue<f32>>` for `ActiveValue<std::option::Option<f32>>`
```

The error says the `Option<T>` is converted to `ActiveValue<Option<T>>`. But I want `Option<T>` to be converted to `ActiveValue<T>` in my use case.

I think the associated code is here:
https://github.com/SeaQL/sea-orm/blob/f2a774573a962426d5c4ac9a5abb74d5b63313b9/src/entity/active_model.rs#L301-L308

I fix the bug and I will issue a PR.

PR #323, opened 2021-11-16
Fix `IntoActiveValue` for Option

YoshieraHuang:
Related to #322


tyt2y3:
@billy1624 thoughts?

billy1624:
Given the original model

```rust
#[sea_orm(table_name = "fruit")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub name: String,
    pub cake_id: Option<i32>,
}
```

Some of the possible mappings are...

1. Mapping `T` to `T`
    - Handled by
        ```rust
        impl IntoActiveValue<T> for T {
            fn into_active_value(self) -> ActiveValue<T> {
                Set(self)
            }
        }
        ```
    - Well handled!

2. Mapping `Option<T>` to `T`
    - Handled by
        ```rust
        impl IntoActiveValue<T> for Option<T> {
            fn into_active_value(self) -> ActiveValue<T> {
                match self {
                    Some(value) => Set(value),
                    None => Unset(None),
                }
            }
        }
        ```
    - Well handled!

3. Mapping `Option<Option<T>>` to `Option<T>`
    - Handled by
        ```rust
        impl IntoActiveValue<Option<T>> for Option<Option<T>> {
            fn into_active_value(self) -> ActiveValue<Option<T>> {
                match self {
                    Some(value) => Set(value),
                    None => Unset(None),
                }
            }
        }
        ```
    - Well handled!

4. Mapping `Option<T>` to `Option<T>`
    - Handled by
        ```rust
        impl IntoActiveValue<T> for Option<T> {
            fn into_active_value(self) -> ActiveValue<T> {
                match self {
                    Some(value) => Set(value),
                    None => Unset(None),
                }
            }
        }
        ```
    - **Wait! Seems wrong...**
        ```rust
        #[derive(DeriveIntoActiveModel)]
        pub struct NewFruit {
            pub cake_id: Option<i32>,
        }

        // This is fine
        assert_eq!(
            my_fruit::NewFruit {
                cake_id: Some(1),
            }
            .into_active_model(),
            fruit::ActiveModel {
                id: Unset(None),
                name: Unset(None),
                cake_id: Set(Some(1)),
            }
        );

        assert_eq!(
            my_fruit::NewFruit {
                cake_id: None,
            }
            .into_active_model(),
            fruit::ActiveModel {
                id: Unset(None),
                name: Unset(None),
                // Isn't this `Set(None)`?
                cake_id: Unset(None),
            }
        );
        ```

billy1624:
May I ask @Acidic9 for help? Since he is the original author of it

tyt2y3:
Thanks @billy1624 for the clarification.

This might not be a bug. Yeah so there are different possibilities,

1. `ActiveValue<Option<T>>` -> `Option<Option<T>>`, can be `Set(T)`, `Set(NULL)` or `Unset`
2. `ActiveValue<T>` -> `T`: always map to `Set`
3. `ActiveValue<T>` -> `Option<T>`: can be `Set` or `Unset`, but not `NULL`
4. T -> `Option<T>`: always `Set` and can be `NULL` (this does not currently exist)

So I think there is a semantic ambiguity, but probably not a bug.

Acidic9:
```rust
#[sea_orm(table_name = "fruit")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub foo: String,
    pub bar: Option<i32>,
}

#[derive(DeriveIntoActiveModel)]
pub struct UpdateFruit {
    pub name: String,
    pub price: f64, // Sets bar to `Set(Some(price))`
}

#[derive(DeriveIntoActiveModel)]
pub struct UpdateFruit {
    pub name: String,
    pub price: Option<f64>, // Sets bar to `Set(price)`
}

#[derive(DeriveIntoActiveModel)]
pub struct UpdateFruit2 {
    pub name: String,
    pub price: Option<Option<f64>>, // Sets bar to `Set(Some(price))` or `Set(None`
}
```
I think this is the ended behaviour of the API I had in mind when writing this.. they are not all working correctly?
I haven't had time to clone this PR and test it, but I thought I'd share what I was expecting my implementation to do.

YoshieraHuang:
Hi, thanks for your concerns. After reading comments, I agree that it is not a bug, it is just an ambiguity. I think the ambiguity originates from the **nullability** of fields in `Model`. For example:
```rust
#[sea_orm(table_name = "fruit")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub non_nullable: String,
    pub nullalbe: Option<String>
}
```

There are some possible desired mapping from `UpdateXXX` to `ActiveModel` for `non_nullable` field:
 - `String` -> `Set(String)`, Ok, works fine for now.
 - `Option::<String>::Some("string".to_string())` -> `Set("string".to_string())`.   Uhh, seems wrong. It is mapped to `Set(Some("string".to_string()))`. The type is wrong.
 - ``

YoshieraHuang:
Hi, thanks for your concerns. After reading comments, I agree that it is not a bug, it is just an ambiguity. I think the ambiguity originates from the **nullability** of fields in `Model`. For example:
```rust
#[sea_orm(table_name = "fruit")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub non_nullable: String,
    pub nullable: Option<String>
}
```
The current code is reasonable for `nullable` field:
1. `Some(Some(String))` -> `Set(Some(String))`
2. `Some(None)` -> `Set(None)`
3. `Some(String)` -> `Set(Some(String))`
4. `None` -> `Unset(None)`

It works fine now .1 and 2 is achieved by `impl IntoActiveValue<Option<T>> for Option<Option<T>>`, whiile 3 and 4 is achieved by `impl IntoActiveValue<Option<T>> for Option<T>`. 

There are some possible desired mapping from `UpdateXXX` to `ActiveModel` for `non_nullable` field:
 1.  `String` -> `Set(String)`, the string is required and set the string to field. Ok, works fine for now.
 2.  `None` -> `Unset(None)`.  the string is not required and remaining the field untouched. Ok, also fine. 
 3.  `Some(String)` -> `Set(String)`.  the string is not required but it is given, i have to set the field. Uhh, seems wrong. The type mismatches.
 
The 3rd of the mapping of `non-nullable` field contradicts with the 3rd of `nullable` field. However, now there is no tools to tell from each other.

tyt2y3:
Yes, it appears our type system do not currently capture the information of whether a field is nullable or not.

If, we have a `ActiveValue` and `ActiveValueNullable`, then there would not be such ambiguity. But it seems to require a major overhaul.

tyt2y3:
Oh, there is the `Deserialize` trait.

So, from a serde_json standpoint, can we actually enforce the schema of `required but nullable`?

Seems not! Because it maps both `undefined` and `null` to `None`.

billy1624:
I will revisit this after
- #340

tyt2y3:
Close it for now, we need to look for a better solution

kun1988:
> Close it for now, we need to look for a better solution

Have a better solution now? Thanks.

franklx:
I managed to implement partial updates using a custom simple Option-like type:
https://github.com/franklx/sea-orm/commit/25858941ebbf3d22e2e1d53ccee97af1ee77f106
After testing and refinements I'll open a PR.

LockedThread:
I wish there was a better solution for this

PR #331, opened 2021-11-22
Add TryFromU64 trait for `DateTime<FixedOffset>`.

kev0960:
The timestamp column (with time zone) generates the field with `chrono::DateTime<chrono::FixedOffset>`.
Sadly, `DeriveEntityModel` macro fails because `TryFromU64` is not implemented for this type.

I added the trivial implementation for `chrono::DateTime<chrono::FixedOffset>` since timestamp is not convertible to u64 anyway.


billy1624 (approved):
Thanks!! @kev0960

kev0960:
Hi can someone merge this into main branch?

PR #335, opened 2021-11-26
Migration

### LINKED ISSUE #333: Migration Supports
Feature provided by sea-schema
- SeaQL/sea-schema#12

billy1624:
## PR Info
- Resolve #333 
- Depends on SeaQL/sea-schema#26


src/database/transaction.rs
```diff
@@ -242,9 +244,7 @@ impl Drop for DatabaseTransaction {
 }
 
 #[async_trait::async_trait]
-impl<'a> ConnectionTrait<'a> for DatabaseTransaction {
-    type Stream = TransactionStream<'a>;
-
+impl ConnectionTrait for DatabaseTransaction {
```
tyt2y3, line 247: okay removing a lifetime is cool

src/database/transaction.rs
```diff
@@ -345,10 +349,17 @@ impl<'a> ConnectionTrait<'a> for DatabaseTransaction {
             }
             #[cfg(feature = "mock")]
             InnerConnection::Mock(conn) => return conn.query_all(stmt),
+            #[allow(unreachable_patterns)]
+            _ => unreachable!(),
```
tyt2y3, line 353: why is it needed here?

billy1624: `sea-schema` depends on `sea-orm` because it needs the symbol to compile. It's up to end-user to decide which backend they want to enable. So, it's defined as below, but this won't compile because it requires at least one backend / `mock` feature to be selected.

```toml
sea-orm = { git = "https://github.com/SeaQL/sea-orm.git", branch = "migration", default-features = false, features = ["macros"], optional = true }
```
https://github.com/SeaQL/sea-schema/blob/3309d13819ae8b2d58dcf4ebbd4cb1936a7f8f15/Cargo.toml#L40

tyt2y3: okay if we actually figured a way of building sea-orm without any backend, it's good! It could have resolved https://github.com/SeaQL/sea-orm/issues/438#issuecomment-1007968237 as a side effect?

PR #339, opened 2021-11-30
Insert & Update Return `Model`

### LINKED ISSUE #289: Why insert, update, etc return a ActiveModel instead of Model?
I find it curious, since you can easily go from a model to a activemodel, but not backwards (from what i've seen), and most of the time after a insert i want the Model which i can return from my api.

I've seen the impl of the insert model and it in fact does a query to find the inserted model and then it converts it to a activemodel, I think we should let the end user decide whether to convert it to a activemodel or not and just return a model.

billy1624:
## PR Info
- Resolve #289

## Breaking Changes
- `insert`, `update` and `save` return `Model` instead of `ActiveModel`
- `ActiveModelBehavior::after_save` takes `<Self::Entity as EntityTrait>::Model` instead of `ActiveModel`

### DIFF
--- a/tests/active_enum_tests.rs
+++ b/tests/active_enum_tests.rs
@@ -146,7 +146,7 @@ pub async fn insert_active_enum_child(db: &DatabaseConnection) -> Result<(), DbE
         category: Set(Some(Category::Big)),
         color: Set(Some(Color::Black)),
         tea: Set(Some(Tea::EverydayTea)),
-        ..am
+        ..am.into_active_model()
     }
     .save(db)
     .await?;
--- a/tests/active_enum_tests.rs
+++ b/tests/active_enum_tests.rs
@@ -21,25 +21,25 @@ async fn main() -> Result<(), DbErr> {
 pub async fn insert_active_enum(db: &DatabaseConnection) -> Result<(), DbErr> {
     use active_enum::*;
 
-    let am = ActiveModel {
-        category: Set(None),
-        color: Set(None),
-        tea: Set(None),
-        ..Default::default()
-    }
-    .insert(db)
-    .await?;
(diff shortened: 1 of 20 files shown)


tyt2y3:
Still drafting?

billy1624:
I planned to merge those for 0.4.1
- #348
- #347

Then, moving on to merge PRs for 0.5.0
- #340
- #339 (this PR)

(PR mentioned in tentative merge sequence)

tests/active_enum_tests.rs
```diff
@@ -89,7 +89,7 @@ pub async fn insert_active_enum(db: &DatabaseConnection) -> Result<(), DbErr> {
             .unwrap()
     );
 
-    let res = am.delete(db).await?;
+    let res = model.into_active_model().delete(db).await?;
```
tyt2y3, line 92: We should add a wrapper method to `Model` so it can be
`model.delete(db)`

tests/uuid_tests.rs
```diff
@@ -28,12 +28,11 @@ pub async fn insert_metadata(db: &DatabaseConnection) -> Result<(), DbErr> {
         bytes: vec![1, 2, 3],
         date: Some(Date::from_ymd(2021, 9, 27)),
         time: Some(Time::from_hms(11, 32, 55)),
-    }
-    .into_active_model();
+    };
 
-    let result = metadata.clone().insert(db).await?;
+    let result = metadata.clone().into_active_model().insert(db).await?;
```
tyt2y3, line 33: Same here

tyt2y3 (approved):

tyt2y3:
Please open a new PR to address the above suggestions.

PR #340, opened 2021-11-30
Rework `ActiveValue`

### LINKED ISSUE #321: Rework `ActiveValue`
Currently our ActiveValue looks like this:

```rust
pub struct ActiveValue<V>
where
    V: Into<Value>,
{
    value: Option<V>,
    state: ActiveValueState,
}
```

I am thinking to go back to a simple Enum:

```rust
pub enum ActiveValue<V>
where
    V: Into<Value>,
{
    Set(V),
    Unchanged(V),
    NotSet,
}
```

The overall API should remain the same, just that we should deprecate `Unset(None)` and favour `NotSet` instead.

billy1624:
## PR Info
- Resolve #321

## Breaking Changes
- Rename `sea_orm::unchanged_active_value_not_intended_for_public_use()` to `sea_orm::Unchanged()`
- Rename `ActiveValue::unset()` to `ActiveValue::not_set()`
- Rename `ActiveValue::is_unset()` to `ActiveValue::is_not_set()`
- `PartialEq` of `ActiveValue` will now check the equality of state instead just checking the value

## Deprecated
- `Unset()`

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -58,8 +58,9 @@ where
     ActiveValue::unset()
 }
 
-#[doc(hidden)]
-pub fn unchanged_active_value_not_intended_for_public_use<V>(value: V) -> ActiveValue<V>
+/// Defines an unchanged operation on an [ActiveValue]
+#[allow(non_snake_case)]
+pub fn Unchanged<V>(value: V) -> ActiveValue<V>
 where
     V: Into<Value>,
 {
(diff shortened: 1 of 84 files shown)


src/entity/active_model.rs
```diff
@@ -45,31 +55,18 @@ where
     ActiveValue::set(v)
 }
 
-/// Defines an unset operation on an [ActiveValue]
+/// Defines an not set operation on an [ActiveValue]
 #[allow(non_snake_case)]
-pub fn Unset<V>(_: Option<bool>) -> ActiveValue<V>
+pub fn NotSet<V>() -> ActiveValue<V>
```
tyt2y3, line 60: I think the intended use is to `NotSet` (not a function call). Which means we should 
```rust
pub use ActiveValue::NotSet;
```

The old `Unset` function should still be kept, just being deprecated.

billy1624: That make sense

billy1624: See d0c6b5e

tyt2y3 (approved):

PR #351, opened 2021-12-02
add offset and limit

lz1998:


src/query/helper.rs
```diff
@@ -237,6 +237,40 @@ pub trait QueryOrder: Sized {
...
+    ///         .limit(10)
+    ///         .build(DbBackend::MySql)
+    ///         .to_string(),
+    ///     "SELECT `cake`.`id`, `cake`.`name` FROM `cake` LIMIT 10"
+    /// );
+    /// ```
+    fn limit(mut self, limit: u64) -> Self {
+        self.query().limit(limit);
+        self
+    }
```
billy1624, line 273: The `offset` & `limit` were added to `QueryOrder`. I think it's better to add it to `QuerySelect` instead.

Any thoughts? @tyt2y3

billy1624 (changes requested):

src/query/helper.rs
lz1998: Now I have moved `offset` & `query` to `QuerySelect`.

billy1624 (approved):
Thanks!! @lz1998 for the contributions! this looks good to me :P

tyt2y3 (approved):

PR #360, opened 2021-12-04
Add feature to generate table Iden

### LINKED ISSUE #356: Add Table on the generated Column enum
This might seem weird, but it comes from a legit need to no change hundreds of lines of sea-query code.
Currently we have a lot of enums like:
```rust
#[derive(Iden, Clone)]
pub enum InterceptEntries {
    Table,
    Id,
    RequestId,
}
```
That we use like:
```rust
Query::select()
.column((InterceptEntries::Table, InterceptEntries::Id))
.column((InterceptEntries::Table, InterceptEntries::RequestId))
.from(InterceptEntries::Table)
// Wayyyy more code 
```

We want to use sea-orm for all the most simple use cases, but doing so would imply having to change all the `InterceptEntries::Table` to `intercept_entries::Entity` and `InterceptEntries::Id` to `intercept_entries::Column::Id`.
That is a lot of code for us.

A smoother transition would be to just change the import:
```rust
use entities::intercept_entries::Column as InterceptEntries;
```
But it would require the `Column` enum to have a `Table` field.

Maybe this could be a flag only for people that need it?

Sytten:
Fixes #356

Let me know where to write the documentation for that. I think a section to use sea-orm with sea-query could be appropriate.

Like I mentioned on discord, I think the strum crate needs a bit of love:
- The docs.rs links are not pointing at the right place
- The tests didn't run since they use the old crate name and the attribute could be replaced by `sea_strum`
- It should be rebased on the original repo (doable via contacting github support).


sea-orm-macros/src/derives/entity_model.rs
```diff
@@ -58,6 +59,18 @@ pub fn expand_derive_entity_model(data: Data, attrs: Vec<Attribute>) -> syn::Res
...
+    #[cfg(feature = "with-table-iden")]
+    if let Some(table_name) = table_name {
+        let table_field_name = Ident::new("Table", Span::call_site());
+        columns_enum.push(quote! {
+            #[sea_orm(table_name=#table_name)]
+            #[strum(disabled)]
+            #table_field_name
+        });
+        columns_trait
+            .push(quote! { Self::#table_field_name => panic!("Table cannot be used as a column") });
```
Sytten, line 72: I was not sure what to do with that, I figured it was best to abort early

billy1624: Make sense to me

billy1624:
Hey @Sytten, just now I want to add some test cases. But I don't have the permission to push commits. Can you tick that checkbox and give that permission to sea-orm maintainers. Check [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) for how to do so.

Sytten:
Ha since I forked under my organization instead of personal, I can't give access to maintainers :S
This is fairly annoying, I will give you repo access I guess?

billy1624:
> Ha since I forked under my organization instead of personal, I can't give access to maintainers :S This is fairly annoying, I will give you repo access I guess?

That would work :P

sea-orm-macros/Cargo.toml
```diff
@@ -25,3 +25,6 @@ proc-macro2 = "^1"
 [dev-dependencies]
 sea-orm = { path = "../", features = ["macros"] }
 serde = { version = "^1.0", features = ["derive"] }
+
+[features]
+with-table-iden = []
```
tyt2y3, line 30: I don't think it needs to be a feature in `sea-orm-macros`. It should be exposed as an additional attribute to the macro.

I could not find any proposed interface, but I imagine it could be:

```rust
#[sea_orm(table_name = "cake", with_table_iden)]
pub struct Model { ... }
```

billy1624: Same here? Don't want it to be feature gated?

Sytten: Ok will work on that!

billy1624:
Hey @Sytten, for the documentation you can fork and create a PR at https://github.com/SeaQL/seaql.github.io.

Then create a file with name `09-xxx-xxx.md` under https://github.com/SeaQL/seaql.github.io/tree/master/SeaORM/docs/08-advanced-query.

billy1624:
Thanks!!

Sytten:
@billy1624 @tyt2y3 I switched to an attribute, let me know what you think!

tyt2y3 (approved):
Pretty neat!

PR #361, opened 2021-12-07
Rename `sea-strum` lib back to `strum`

billy1624:
## PR Info
- Depends on SeaQL/strum#1


tyt2y3:
Well this confuses me. So the crate name is still `sea-strum` but the namespace is `strum`?

billy1624:
> Well this confuses me. So the crate name is still `sea-strum` but the namespace is `strum`?

Correct. I guess this would be much easier for us to fetch upstream changes?

PR #363, opened 2021-12-08
`Delete::many()` with `TableRef`

### LINKED ISSUE #362: Delete::many() doesn't work when schema_name is defined
I've a  PostgreSQL database where tables are located in their schema.
I need to delete multiple rows, so I'm using Delete::many(), but it seems that the schema name is not included in the query, so the database complains about it `relation \"valid_table_name\" does not exist`

billy1624:
## PR Info
- Resolve #362


tyt2y3 (approved):
Great

PR #373, opened 2021-12-09
First metric and tracing implementation

nappa85:
As discussed on https://github.com/SeaQL/sea-orm/discussions/372, I've done a first, minimal, implementation for tracing and metric support.
I'm actually using this version on my environment and it seems to works like a charm, especially the metric part.
For the Tracing part, I've instrumented a minimal part of crate methods, and I've decided to instrument it as level trace because I think a library crate should not be "too loud".
Maybe we could control that behavior with a feature if you prefer

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ async-trait = { version = "^0.1" }
 chrono = { version = "^0", optional = true }
 futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
-log = { version = "^0.4", optional = true }
+tracing = "0.1"
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.4.1", path = "sea-orm-macros", optional = true }
 sea-query = { version = "^0.19.1", features = ["thread-safe"] }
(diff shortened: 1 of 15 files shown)


tyt2y3:
Awesome! Thank you for the big contribution again.
Looks really good so far.

I think tracing has logger output support, so we might want to put that shim behind a 'use-log' feature flag.

nappa85:
> I think tracing has log output support, so we might want to put that shim behind a 'use-log' feature flag.

What do you mean by "log output support"?
To output log messages you need a subscriber, if you look the crate is in dev-dependencies in place of env-logger to allow test logging to work as before

tyt2y3:
> What do you mean by "log output support"?

I found it, I mean this https://docs.rs/tracing/0.1.29/tracing/index.html#log-compatibility
Seems like the user can enable that feature flag on their own, so we don't have to do anything for it )

tyt2y3 (approved):
Seem good to me. @billy1624 thoughts?

tyt2y3:
We will release this in `0.5.0`

PR #376, opened 2021-12-10
Fix related & linked with enum columns

### LINKED ISSUE #374: find_with_related panic
Hey, I am new to sea-orm, so was looking to see if I could get some help around this.

I have two entities with a one to many relation (that I built using the cli generation).
I can fetch them both individually, but it panics when I try to join them together.

They are quite big structs, each with the following properties
- id is a uuid
- have one Postgres enum
- with the only overlapping fields being id and their respective createdAt and updatedAt.
- foreign key is on the relevant uuid of the other table.
- fields are camelCase

I'm using the column_name attribute to map from snake case in the struct to camel case in the db.

here is a variant of the query in question, but just building the query

```
    let query = CustomerEmployeeRecord::find()
        .find_with_related(HistoricalCustomerEmployeeRecord)
        .build(DbBackend::Postgres)
        .to_string();

```

Result:
```
thread 'main' panicked at 'cannot apply alias for expr other than Column'
```

any help we be greatly appreciated!

billy1624:
## PR Info
- Resolve #374

### DIFF
--- a/src/entity/link.rs
+++ b/src/entity/link.rs
@@ -30,7 +30,7 @@ pub trait Linked {
 
             select.query().join_as(
                 JoinType::InnerJoin,
-                unpack_table_ref(&rel.from_tbl),
+                rel.from_tbl,
                 SeaRc::clone(&from_tbl),
                 join_tbl_on_condition(from_tbl, to_tbl, rel.from_col, rel.to_col),
             );
--- a/src/query/combine.rs
+++ b/src/query/combine.rs
@@ -1,9 +1,12 @@
 use crate::{
-    EntityTrait, IdenStatic, IntoSimpleExpr, Iterable, QueryTrait, Select, SelectTwo, SelectTwoMany,
+    ColumnTrait, EntityTrait, IdenStatic, IntoSimpleExpr, Iterable, QueryTrait, Select, SelectTwo,
+    SelectTwoMany,
 };
 use core::marker::PhantomData;
 pub use sea_query::JoinType;
-use sea_query::{Alias, ColumnRef, Iden, Order, SeaRc, SelectExpr, SelectStatement, SimpleExpr};
+use sea_query::{
+    Alias, ColumnRef, DynIden, Expr, Iden, Order, SeaRc, SelectExpr, SelectStatement, SimpleExpr,
+};
 
 macro_rules! select_def {
     ( $ident: ident, $str: expr ) => {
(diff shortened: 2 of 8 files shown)


billy1624:
Hey @MorganNesbitt, can you try this PR?

You can update your toml
```toml
sea-orm = { git = "https://github.com/SeaQL/sea-orm.git", branch = "issues/374", ... }
```

tyt2y3:
Can briefly explain what's the root cause?

MorganNesbitt:
hi @billy1624 that fixed the panic!

One thing I did notice now that I have a join working, is that it adds an order by id no matter what. which seemed odd, and when you add an order to the query it doesn't replace it just puts it right after.

A normal single query doesn't add an order, but the join does.

is this expected?

Below is what's being generated when I do a find_with_related but provide no order_by myself.
This is part of the debug that's being printed when I run the query
```
...
FROM
      "customer_employee_record"
      LEFT JOIN "historical_customer_employee_record" ON "customer_employee_record"."id" = "historical_customer_employee_record"."customerEmployeeRecordId"
    WHERE
      "customer_employee_record"."companyId" = $1
    ORDER BY
      "customer_employee_record"."id" ASC
```

src/query/combine.rs
```diff
@@ -42,10 +45,17 @@ where
...
+                        },
+                        SimpleExpr::AsEnum(_, simple_expr) => match simple_expr.as_ref() {
+                            SimpleExpr::Column(col_ref) => match &col_ref {
+                                ColumnRef::Column(col) | ColumnRef::TableColumn(_, col) => col,
+                            },
+                            _ => {
+                                panic!("cannot apply alias for AsEnum with expr other than Column")
+                            }
                         },
-                        _ => panic!("cannot apply alias for expr other than Column"),
```
billy1624, line 48: It panic at this line. It didn't catch `ColumnRef::AsEnum` and apply alias for it.

src/entity/link.rs
```diff
@@ -30,7 +30,7 @@ pub trait Linked {
 
             select.query().join_as(
                 JoinType::InnerJoin,
-                unpack_table_ref(&rel.from_tbl),
+                rel.from_tbl,
```
billy1624, line 33: Can I caught some other bug as well. This should pass the `TableRef` in directly, so that schema name will also be included if needed.

src/query/join.rs
```diff
@@ -79,21 +79,28 @@ where
 
             slf.query().join_as(
                 JoinType::LeftJoin,
-                unpack_table_ref(&rel.to_tbl),
+                rel.to_tbl,
```
billy1624, line 82: > Can I caught some other bug as well. This should pass the TableRef in directly, so that schema name will also be included if needed.

Same here

billy1624:
> hi @billy1624 that fixed the panic!
> 
> One thing I did notice now that I have a join working, is that it adds an order by id no matter what. which seemed odd, and when you add an order to the query it doesn't replace it just puts it right after.
> 
> A normal single query doesn't add an order, but the join does.
> 
> is this expected?
> 
> Below is what's being generated when I do a find_with_related but provide no order_by myself. This is part of the debug that's being printed when I run the query
> 
> ```
> ...
> FROM
>       "customer_employee_record"
>       LEFT JOIN "historical_customer_employee_record" ON "customer_employee_record"."id" = "historical_customer_employee_record"."customerEmployeeRecordId"
>     WHERE
>       "customer_employee_record"."companyId" = $1
>     ORDER BY
>       "customer_employee_record"."id" ASC
> ```

That additional order by is intended. If you look into the implementation of `SelectTwoMany`, you will notice that it consolidate single dimension query rows into nested result, `Vec<(E::Model, Vec<F::Model>)>`. Hence, the relative order of `E::Model` result and its related `F::Model` result is important.

https://github.com/SeaQL/sea-orm/blob/18ed68371238003cc39a13883f8c3a33256f46d9/src/executor/select.rs#L751-L785

_(My explanation might be confusing loll)_

src/query/combine.rs
billy1624: @tyt2y3

tyt2y3 (approved):
Great thanks

PR #378, opened 2021-12-10
Add docker create script for contributors to setup databases locally

billy1624:


tyt2y3:
Actually, we can use docker-compose

tyt2y3 (changes requested):
We can use docker-compose yaml

billy1624:
What I intended to do here is just setting up all the databases at once. Then, develop can fire it up when they want to. Instead of running all at once.

image

billy1624:
> We can use docker-compose yaml

But... yes, this would be the conventional way of doing so. See
- https://github.com/launchbadge/sqlx/blob/f2b3cc8dc3035279243f3a636379f7f20838488a/.github/workflows/sqlx.yml#L217
- https://github.com/launchbadge/sqlx/blob/master/tests/docker-compose.yml

PR #384, opened 2021-12-14
Fix rocket version

Gabriel-Paulucci:
Fixing the rocket version to no longer use the master git breanch, and use the create version `0.5.0-rc.1`

Using the version directly from git ends up creating conflicts

And if possible I think it would be interesting to add this library to the crate


tyt2y3 (approved):
Thank you!

PR #399, opened 2021-12-23
Log with tracing-subscriber

billy1624:
## PR Info
- All examples, issues and cli should use `tracing-subscriber` to log debug messages instead of `env_logger`, due to #373


tyt2y3:
That's a nice catch

tyt2y3 (approved):

PR #400, opened 2021-12-24
Model with Generics

### LINKED ISSUE #402: Model with Generics
I am working on a crate called [`async-graphql-relay`](https://github.com/oscartbeaumont/async-graphql-relay) and have recently redone the public API to make it less cumbersome to use (you can checkout the example folder in the repo). When doing this I introduced a type `RelayNodeID` which is generic on the Model so it is declared as `RelayNodeID<Self>`. This is done so that among other things all ID's returned from GraphQL are globally unique and contain information about the specific type they are. This type is a wrapper around a standard Uuidv4 and from a database perspective should be treated this way.

After chatting with @billy1624 he pointed me towards [this resource](https://github.com/SeaQL/sea-orm/blob/master/issues/324/src/model.rs) which shows how to implement `sea-orm` support for custom types. The issue I found with this is that my type is generic and this does not currently seem to be supported by `sea-orm`. When I run my code the `DeriveEntityModel` macro panics (refer to the code linked below).

I put together some code [here](https://gist.github.com/oscartbeaumont/94fc0412668609255df88f17a3de0cfa) which should help show the sort of syntax I am going for without all the complexity and macros of the `async-graphql-relay` codebase.

Thanks, @billy1624 for creating PR #400 which is going to be the fix for this.

billy1624:
## PR Info
- Resolve #402

### DIFF
--- a/issues/400/Cargo.toml
+++ b/issues/400/Cargo.toml
@@ -1,0 +1,11 @@
+[workspace]
+# A separate workspace
+
+[package]
+name = "sea-orm-issues-400"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+sea-orm = { path = "../../", features = [ "sqlx-mysql", "runtime-async-std-native-tls" ]}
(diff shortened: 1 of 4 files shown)


tyt2y3:
Looks great

Issue #425, opened 2022-01-05
Clone is not implemented for `sea_orm::DatabaseConnection` in rocket_example


Rudi3:
I tried the `rocket_example` and I'm getting an error in `pool.rs`.

https://github.com/SeaQL/sea-orm/blob/82bad4a376efb875ac32ab601ac2cd9cce191e3f/examples/rocket_example/src/pool.rs#L12

```
the trait bound `sea_orm::DatabaseConnection: std::clone::Clone` is not satisfied

the trait `std::clone::Clone` is not implemented for `sea_orm::DatabaseConnection`rustc(E0277)
pool.rs(12, 5): the trait `std::clone::Clone` is not implemented for `sea_orm::DatabaseConnection`
```

Removing `#[derive(Clone)]` from `SeaOrmPool` fixes the error, but is it intended to have the `Clone` trait?

Features I was using:
```
sea-orm = { version = "0.5.0", features = ["sqlx-postgres", "macros", "runtime-tokio-rustls"]}
sea-orm-rocket = { git = "https://github.com/SeaQL/sea-orm" }
```

billy1624:
Hi @Rudi3, I believe you can safely remove the derive `Clone` here.

https://github.com/SeaQL/sea-orm/blob/82bad4a376efb875ac32ab601ac2cd9cce191e3f/examples/rocket_example/src/pool.rs#L10-L13

tyt2y3:
You missed the `default-features = false`

```toml
sea-orm = { version = "0.5.0", default-features = false, features = ["sqlx-postgres", "macros", "runtime-tokio-rustls"]}
```

PR #428, opened 2022-01-05
#425

kkroy22:
added #[derive(clone)] for DatabaseConnection


tyt2y3:
As explained in the other thread, we need a better solution

PR #429, opened 2022-01-06
Support the use of chrono::DateTime<Utc> in sea-orm

charleschege:
Support sea-orm derive macros for using Sea-Query `DateTimeUtc`.

An update to sea-query dependency from depending on the fork will be made once sea-query PR supporting `DateTimeUtc` is approved


charleschege:
@billy1624  @tyt2y3  Thoughts on this also welcome

billy1624:
We need some testing against all three databases

charleschege:
> 

@billy1624 Can you clarify this or give an example. Currently, the example I can reference is at `tests/common/features/applog.rs`

src/entity/prelude.rs
```diff
@@ -28,6 +28,50 @@ pub use chrono::NaiveDateTime as DateTime;
...
+/// #[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
+/// pub enum Column {
+///     Id,
+///     Name,
+///     CakeId,
+///     Timer,
+/// }
+/// ```
+#[cfg(feature = "with-chrono")]
+pub type DateTimeUtc = chrono::DateTime<chrono::Utc>;
```
billy1624, line 73: The example would be inserting and updating `ActiveModel` with `DateTimeUtc` attribute.

e.g.

```rust
#[sea_orm(table_name = "applog")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub action: String,
    pub json: Json,
    pub deleted_at: DateTimeUtc,
}
```

charleschege: @billy1624 Some help on understanding the results of github action failures. They seem to build fine and store the values in the databases, but retrieval is an issue.

For example, asserting on mysql
```
2022-01-07T14:12:10.675164Z  INFO sqlx::query: /* SQLx ping */; rows: 0, elapsed: 5.909ms  
Error: Query("error returned from database: column \"launch_date\" is of type timestamp with time zone but expression is of type text")
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/test/src/lib.rs:195:5

```

billy1624: The problem you encountered is rooted in https://github.com/SeaQL/sea-query/pull/222#discussion_r780303186

charleschege:
@tyt2y3 @billy1624 Anyone who has an idea on how to fix the errors from the CI running mysql 5.7 ? 
Running the program panics with
```
 "CREATE TABLE IF NOT EXISTS `satellites` ( `id` int NOT NULL AUTO_INCREMENT PRIMARY KEY, `satellite_name` varchar(255) NOT NULL, `launch_date` timestamp NOT NULL, `deployment_date` timestamp NOT NULL )"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Database(MySqlDatabaseError { code: Some("42000"), number: 1067, message: "Invalid default value for 'deployment_date'" })', src/main.rs:57:18
```

This is an issue with older versions of mysql, in this case version 5.7. Solving the problem on my end involved changing myql 5.7 configuration as specified in this article https://ixnfo.com/en/the-solution-of-error-error-1067-42000-at-line-211-invalid-default-value-for-blablabla.html

So what I am asking, is this a configuration issue, if not how do I solve the issue from sea-orm ? I am out of ideas

charleschege:
In case you need more details, re-running the tasks can provide this or check my fork https://github.com/charleschege/sea-orm/runs/4748288791?check_suite_focus=true

tests/common/features/datetimeutc.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
+#[sea_orm(table_name = "satellites")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub satellite_name: String,
+    pub launch_date: DateTimeUtc,
+    pub deployment_date: DateTimeUtc,
+}
```
billy1624, line 11: Hey @charleschege, you could set one of the timestamp column to nullable. This will solve the problem, let say changing the type of `deployment_date` attribute to `Option<DateTimeUtc>`

charleschege: Thanks @billy1624

charleschege: I think I am a little bit confused about some results, does using the derive macro yield different results than the `Table::create` statement?

charleschege: For example, on run https://github.com/charleschege/sea-orm/runs/4770827851?check_suite_focus=true

using https://github.com/SeaQL/sea-orm/pull/429/files/1227ad8252815c72b19c4a6aa27e998d3f1e5330#diff-db71d1612ed300a9a9a38e99d0096890446cc04f150507455638b35110d81bf0 results in that error

charleschege: But using the manual method
```
sea_query::Table::create()
                .table(Entity)
                .col(
                    ColumnDef::new(Column::Id)
                        .integer()
                        .not_null()
                        .auto_increment()
                        .primary_key(),
                )
                .col(ColumnDef::new(Column::SatelliteName).string().not_null())
                .col(
                    ColumnDef::new(Column::LaunchDate)
                        .timestamp_with_time_zone()
                        .not_null(),
                )
                .col(
                    ColumnDef::new(Column::DeploymentDate)
                        .timestamp_with_time_zone()
                        .not_null(),
                )
                .if_not_exists()
                .to_owned();
```

Assertion doesn't fail, @billy1624 is there a way to ensure fields are not null using the derive macro ? The module docs in https://docs.rs/sea-orm/latest/sea_orm/ dont show how to do this if possible

billy1624: If `launch_date` is nullable then the table create statement will be like...

```rust
.col(
    ColumnDef::new(Column:: LaunchDate)
        .timestamp_with_time_zone(),
)
```

tyt2y3:
What we are missing is a default expression like `DEFAULT CURRENT_TIMESTAMP`.
We might need to add that in sea-query first though.

PR #431, opened 2022-01-06
Rocket example db pool without deriving Clone

### LINKED ISSUE #425: Clone is not implemented for `sea_orm::DatabaseConnection` in rocket_example

billy1624:
## PR Info
- Resolve #425

### DIFF
--- a/examples/rocket_example/src/pool.rs
+++ b/examples/rocket_example/src/pool.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 #[database("sea_orm")]
 pub struct Db(SeaOrmPool);
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct SeaOrmPool {
     pub conn: sea_orm::DatabaseConnection,
 }


05storm26:
Instead of this, can you make `MockDatabaseConnection` clonable? 

If we could do that, then by default (having the default features enabled) `sea_orm::DatabaseConnection` could be `Clone` by default. Some frameworks like actix_web require app global datas to be clonable. Currently one can only disable the mock feature to fix this. It would be great if  `sea_orm::DatabaseConnection` could be clonable in all cases.

examples/rocket_example/Cargo.toml
```diff
@@ -23,7 +23,7 @@ serde_json = { version = "^1" }
 [dependencies.sea-orm]
 path = "../../" # remove this line in your own project
 version = "^0.5.0"
-features = ["macros", "runtime-tokio-native-tls"]
+features = ["macros", "mock", "runtime-tokio-native-tls"]
```
tyt2y3, line 26: We dont actually have to add `mock` here right?

billy1624: Not needed. I will revert it

billy1624: Reverted

tyt2y3:
I think we don't need this. Better keep it as is.

PR #432, opened 2022-01-06
Insert Default - Inserting `ActiveModel` with all attributes `NotSet`

### LINKED ISSUE #420: Failed to insert row with only default values
Invalid SQL query is generated when inserting into a table that has default values for all columns.

e.g.

When inserting into below table
```
CREATE TABLE topics (
    id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
);
```
with this rust code
```
let topic = topics::ActiveModel { ..Default::default() };
topic.insert(db).await?;
```
returns this error
`DbErr(Query("error returned from database: syntax error at or near \")\""))`

because below query is generated
`INSERT INTO "topics" () VALUES () RETURNING "id"`

billy1624:
## PR Info
- Resolve #420
- Depends on
  - SeaQL/sea-query#223
  - #430


tyt2y3:
Oh I forgot to merge the sea-query PR

tyt2y3:
I don't think this feature is ready. I reverted the commit for now. Please revisit later.

Issue #398, opened 2021-12-22
Support creating a `sea_orm::Database` from any `sqlx::Executor` for example from an `sqlx::any::AnyPool`.


05storm26:
Currently sqlx allows you to create a connection pool with custom pool options based on e.g.: the kind of database (extracted from the url) or anything else.

e.g.:

```rust
    let mut connect_options = sqlx::any::AnyConnectOptions::from_str(&database_url).expect("Incorrect database url!");

    let mut pool_options = sqlx::any::AnyPoolOptions::new();

    if std::matches!(connect_options.kind(), AnyKind::Sqlite) {
        let sqlite_connect_options = SqliteConnectOptions::from_str(&database_url).unwrap()
            .create_if_missing(true);
        connect_options = sqlite_connect_options.into();
        pool_options = pool_options.max_connections(1);
    }

    let kind = connect_options.kind();
    log::info!("Connecting to the database...");
    let mut pool = pool_options.connect_with(connect_options).await.expect("Unable to connect to the database!");
```

(I know that the sqlite `create_if_missing` option could be configured from the database url as well). 

The user should have the ability to create and setup their own connection using sqlx. SeaQL should allow creating `sea_orm::Database` not just from a database url string but also from any `sqlx::Executor`.

I understand that SeaQL needs to know the kind of database to know which flavor of SQL to generate, but 
  1. This could be something that we could allow the user to specify along with giving an `sqlx::Executor`.
  2. In the next release `sqlx::AnyPool` will have an `any_kind` method for getting the database kind, so that will be usefull and can be used to get the underlying database type.

billy1624:
Hey @05storm26, I wonder what is the use case for you to convert any `sqlx::Executor` into `sea_orm::Database`?

billy1624:
Seems `Pool::any_kind()` have been released today.
https://github.com/launchbadge/sqlx/releases/tag/v0.5.10

tyt2y3:
SeaORM will not be able to support AnyPool

PR #434, opened 2022-01-07
Allow creating a sea_orm::Databse from sqlx::ConnectOptions

05storm26:
Seems like doing what is requested in #398 is too difficult. 

However allowing users to create a `sea_orm::Database` from `sqlx::ConnectOptions` implementations like `MySqlConnectOptions` seems to be doable. 

For me this is good enough. 

Also I have added a new feature options `sqlx-any`. If this is enabled the sqlx is `any` feature is turned on and you can create an `sea_orm::Database` from `AnyConnectOptions` as well.

~However this depends on this PR https://github.com/launchbadge/sqlx/pull/1610 (not yet merged!)~ **MERGED** and [released](https://github.com/launchbadge/sqlx/pull/1715)


05storm26:
@billy1624 Can you approve the run now?

billy1624:
Sure!

05storm26:
I think the rocket example that failed has beed fixed. 

Also I have went through and added Hash, PartialEq, Clone derives on all types where possible, in the second commit.

billy1624 (commented):
I'm curious about why we need to derive `Hash`?

src/database/mod.rs
```diff
@@ -23,11 +32,55 @@ use crate::DbErr;
...
 /// Defines the configuration options of a database
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ConnectOptions {
-    /// The URI of the database
-    pub(crate) url: String,
+    /// The database sqlx::ConnectOptions used to connect to the database.
+    pub(crate) connect_options: SqlxConnectOptions,
+    /// The URI of the database, if this struct was created from an URI string, otherwise None
+    pub(crate) url: Option<String>,
```
billy1624, line 83: We should make this an enum with two variants
- `SqlxConnectOptions`: wrapper enum of all supported databases' connection option
- `ConnectionOptions`: raw url with some other connection option included in the old `ConnectOptions` struct

05storm26: > This PR has changed the url() in a way that it will only return something if it was created from an url because otherwise we currently cannot get the url from the connect_options.

Which is pretty confusing IMO... :\

05storm26:
> I'm curious about why we need to derive Hash?

I am not sure. Maybe someone needs it. If you want to use ConnectOptions in a hashmap as keys maybe. I don't know why you would do that just an idea.

I am generally on the side that we should implement for all types as much from the traits: Clone, Clone, PartialEq, Hash as possible.

src/database/mod.rs
05storm26: What do you think we should do with the url? Should we keep the url() method? Should that only return the url if the sea_orm::ConenctOptions was created from an url string?

billy1624: I think we can keep the `ConnectOptions` untouched. And introduce a new trait (e.g. `ConnectionOptionTrait`) that convert any kinds of DatabaseOptions into `Database`. Then, `Database::connect()` method takes any `ConnectionOptionTrait`.

Make sense? CC @tyt2y3

05storm26: @billy1624 No this doesn't answer the question about what should we return from the `DatabaseConnection` when get_url is called if the `DatabaseConnection` was created from a `sqlx::ConnectOptions` implementation like `SqliteConnectOptions?

About `ConnectionOptionTrait` we don't need that to be able to implement creating DatbaseConnections from both urls and both sqlx::ConnectOptions implementations. You can just do what I already did in this PR implementing with `TryFrom`.

05storm26: @billy1624 any update about this? Specifically about what should `get_url` return if the `DatabaseConnection` was not created from an url string but from an `sqlx::ConnectOptions` e.g.: `SqliteConnectOptions`?

05storm26: @billy1624 @tyt2y3 Anyone?

billy1624: Hey @05storm26, sorry for the delay. I will get back to you ASAP

billy1624:
Will launchbadge/sqlx#1610 be in SQLx 0.6? And when will 0.6 lands?

05storm26:
Due to [changes](https://github.com/launchbadge/sqlx/compare/9890bdb61a57c8b67f38a151ef8fc60d3e0c3711..80962656944d26c1a270ad42b3c5d58c0941cce1) [requested by sqlx owner](https://github.com/launchbadge/sqlx/pull/1610#pullrequestreview-885145801) I had to make a small change in this pr as well.

src/database/mod.rs
05storm26: Hi @billy1624 the required sqlx commits have been merged so this could be merged after the review is completed. 

Could you take a look at this conversation again?

billy1624:
Oh, I wonder why the tests failed

05storm26:
> the trait `std::hash::Hash` is not implemented for `sea_query::Value`

Seems like maybe there was a `sea_query` change that removed `Hash` from `Value`. No idea why. I think it is useful to have these basic traits implemented like `Debug`, `Clone`, `PartialEq`, `Hash` for all possible types.

Currently there is two commits in this PR the actual change and also a second one which adds these useful trait derives on multiple types. Probably that is the one breaking. 

Should I just remove that? Or should we fix this in `sea_query`?

05storm26:
But I am still not sure what to do with the `get_url` method on `DatabaseConnection`, when the connection was not created from an url string but a `sqlx::ConnectOptions`....

billy1624:
> Should I just remove that? Or should we fix this in `sea_query`?

I think we can remove the `Hash` derive

billy1624:
> However allowing users to create a `sea_orm::Database` from `sqlx::ConnectOptions` implementations like `MySqlConnectOptions` seems to be doable.

Now that I re-think this. Shouldn't we just add a method to create `DatabaseConnection` from `sqlx::ConnectOptions`? Similar to `from_sqlx_*_pool` methods.

https://github.com/SeaQL/sea-orm/blob/483de17da885431749962bc8e3d807a020bdb1e7/src/driver/sqlx_mysql.rs#L66-L72

05storm26:
@billy1624 I have rebased this on top of master. Please run the tests for it.

05storm26:
> Now that I re-think this. Shouldn't we just add a method to create `DatabaseConnection` from `sqlx::ConnectOptions`? Similar to `from_sqlx_*_pool` methods.

For me it looks like that should be constructed by calling `Database::connect`. So I changed that to accept an `sqlx::ConnectOptions` as well.

05storm26:
Is `clippy` supposed to run without warning? Because currently `master` `HEAD` has some `clippy` warnings:

```
$ cargo clippy --features "sqlx-all macros with-json with-chrono with-uuid sqlx-all runtime-async-std-native-tls"  -- -D warnings
    Checking sea-orm v0.10.3 (/home/adam/sea-orm)
error: redundant closure
   --> src/database/transaction.rs:278:25
    |
278 |             err.map_err(|e| sqlx_error_to_query_err(e))
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `sqlx_error_to_query_err`
    |
    = note: `-D clippy::redundant-closure` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure

error: passing a unit value to a function
   --> src/driver/sqlx_mysql.rs:216:9
    |
216 |         Ok(self.pool.close().await)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::unit-arg` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
    |
216 ~         self.pool.close().await;
217 +         Ok(())
    |

error: passing a unit value to a function
   --> src/driver/sqlx_postgres.rs:231:9
    |
231 |         Ok(self.pool.close().await)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
    |
231 ~         self.pool.close().await;
232 +         Ok(())
    |

error: passing a unit value to a function
   --> src/driver/sqlx_sqlite.rs:219:9
    |
219 |         Ok(self.pool.close().await)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
    |
219 ~         self.pool.close().await;
220 +         Ok(())
    |


```

05storm26:
I don't think this is ever going to be merged so I am just gonna let it go.

billy1624:
Hey @05storm26, sorry and thanks again for the effort!

Issue #426, opened 2022-01-05
Custom ColumnTypes


Rudi3:
Is it currently possible to use other types such as (Pg) Interval?

I found the `Custom` Variant in the [Docs](https://docs.rs/sea-orm/0.5.0/sea_orm/entity/enum.ColumnType.html), but is this meant to map back to existing `sea_orm::Value`s or can it be used to access other types than the predefined ones?

Is there currently a way to use other types than the [listed ones](https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-structure) or would they have to be implemented in `sea_orm` first?

billy1624:
Hi @Rudi3, regarding the use of custom types you can check the following discussion & PR
- #139
- #400

Btw... how would you defined the interval attribute in the Model?

```rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "mytable")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    #[sea_orm(column_type = "Custom(\"interval\".to_owned())")]
    pub duration: chrono::Duration, # or std::time::Duration?
}
```

Rudi3:
Hi @billy1624, sorry - it seems that I missed those issues.
I'd probably use `chrono` as the other Date/Time types already rely on `chrono` (as far as I've seen) and it should be higher level (and can be converted back to `std::time::Duration` if needed).

But if I try this, `DeriveEntityModel` fails.
I'll try if implementing the traits in https://github.com/SeaQL/sea-orm/pull/400/files and using a wrapper like
`
pub struct Interval<T>(chrono::Duration, std::marker::PhantomData<T>);`

will make my example work.

```
the trait bound `std::time::Duration: sea_orm::TryGetable` is not satisfied

the trait `sea_orm::TryGetable` is not implemented for `std::time::Duration`rustc(E0277)

the trait bound `sea_orm::Value: std::convert::From<std::time::Duration>` is not satisfied

the trait `std::convert::From<std::time::Duration>` is not implemented for `sea_orm::Value`

help: the following implementations were found:
        <sea_orm::Value as std::convert::From<&'a [u8]>>
        <sea_orm::Value as std::convert::From<&'a str>>
        <sea_orm::Value as std::convert::From<bool>>
        <sea_orm::Value as std::convert::From<chrono::DateTime<Tz>>>
      and 19 others
[...]

PR #456, opened 2022-01-14
[DONTMERGE] Interval Support

Rudi3:
Should fix https://github.com/SeaQL/sea-orm/issues/426

I'm not sure what `entitiy_model.rs` does, so I'm not sure if i got that right.

(i got an error in `combine.rs` and just added the `todo!`s for now...)

This depends on changes in sea-query, so merging doesn't make sense atm.


tyt2y3:
I mean it looks great, although I am not sure how should we expose this Postgres-only feature. The `entity_model.rs` is a derive macro, and the code you added converts Rust type (from the Entity attribute) into Column type

PR #471, opened 2022-01-19
Add Send bound to streams.

sebpuetz:
I had to dig a bit into the different `QueryStream` implementations and the only reason for `!Send` I could find was the builder function passed to the builder in the `ouroboros` context. The `self_referencing` macro generates a signature on the current release of `ouroboros` that allows `!Send` constructors:

https://github.com/joshua-maros/ouroboros/blob/main/ouroboros_macro/src/info_structures.rs#L265

~~I forked `ouroboros` and added the `Send` bound on the constructor future and replaced the crates.io dependency with one pointing at my branch. After fixing up the `Send` requirements on the `TransactionStream`, a few other places in this crate needed `+ Send` bounds.~~

~~This PR is not meant to be merged, it's mostly for documentation / demonstration of what would be required to get streams working in `tokio::spawn` context. If you're interested in getting this PR into better shape, we'll have to get into touch with the `ouroboros` maintainers to figure out if the missing `Send` bound is intentional.~~

I just recognized that `TransactionStream::build` can use a synchronous builder since the builder closure doesn't await any futures. That means the change on `ouroboros` is not even required to enable `Send` streams.

The first commit here is a solution that keeps the async constructor, the second one just removes the unused future on build


billy1624 (commented):
Wow, looks good! Thanks @sebpuetz

tyt2y3 (approved):
Thank you! it looks extremely useful!

tyt2y3:
@billy1624 it seems to conflict with the changes you made to `ConnectionTrait`. can you help resolve?

billy1624:
Let me check

billy1624:
Hey @sebpuetz, please check and cherrypick https://github.com/billy1624/sea-orm/commit/4bf35b435c4ef072554878fb6bf39ba439cea34c

PR #489, opened 2022-01-26
Support `DateTime<Utc>` & `DateTime<Local>`

### LINKED ISSUE #381: Support `DateTime<Utc>` & `DateTime<Local>`
I think we should support `DateTime<Utc>` natively in SeaORM soon

_Originally posted by @tyt2y3 in https://github.com/SeaQL/sea-orm/issues/344#issuecomment-991500315_

May be we can create a new type alias `DateTimeUtc` on `chrono::DateTime<Utc>`

billy1624:
## PR Info
- Continue of #429
- Resolve #381
- Depends on SeaQL/sea-query#249

### Added
- supports saving & querying `DateTimeUtc`
- supports saving & querying `DateTimeLocal`
- set default value of `ColumnDef`
- set default value of each column when generating create table statement with `Schema::create_table_from_entity` method


tyt2y3:
Seems Charles's code is included. Does that mean we can close 429?

billy1624:
> Seems Charles's code is included. Does that mean we can close 429?

Correct!

PR #492, opened 2022-01-27
Update ActiveModel by JSON

### LINKED ISSUE #346: how to update dynamicly from json value
Hello, 
Is there any way to support dynamic update from json value?
like update from user input , and only allow update special fields
if the input json field is null, need set None value for activemodel,
so it need get None value from column:

``` rust
struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub name: String,
    pub addr: Option<String>,
    pub other: Option<String>,
    ...
}
...
let allow_change_fields = vec![Column::Name, Column::Addr];

let mut actives = table::ActiveModel::new();
actives.set(Column::Id, 1.into());

// json is from user input, dynamic
let mut json = json!{addr: "new address"}; 

for col in allow_change_fileds{
   if let Some(json_val) =  json.get(col.to_string()) {
     let new_val = if json_val.is_null() {
       col.null_value() // Value::from(None::<i32>) or Value::from(None::<String>)
     } else {
       json_value_to_sea_value(json_val)
     };
     actives.set(col, new_val);
   }
}

actives.update(db).await?;
...
```
Any better suggestion?

https://github.com/qyihua/sea-orm/commit/0fa0f33ac1ebfa1444e8b4de1411a11e59bd4533

billy1624:
## PR Info
- Resolve #346

### Added
- Set the corresponding attributes in the ActiveModel from a JSON value with `ActiveModel::set_from_json` method
- New error variant `DbErr::Json`, representing error of converting JSON value into Model

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -940,6 +940,17 @@ mod tests {
             }
         );
 
+        assert_eq!(
+            fruit::ActiveModel::from_json(json!({
+                "name": "Apple",
+            }))?,
+            fruit::ActiveModel {
+                id: ActiveValue::NotSet,
+                name: ActiveValue::Set("Apple".to_owned()),
+                cake_id: ActiveValue::NotSet,
+            }
+        );
+
         fruit.set_from_json(json!({
             "name": "Apple",
             "cake_id": null,
@@ -976,6 +987,25 @@ mod tests {
             fruit,
             fruit::ActiveModel {
                 id: ActiveValue::NotSet,
+                name: ActiveValue::Set("Apple".to_owned()),
+                cake_id: ActiveValue::Set(Some(1)),
+            }
+        );
+
+        let mut fruit = fruit::ActiveModel {
+            id: ActiveValue::Set(1),
+            name: ActiveValue::NotSet,
+            cake_id: ActiveValue::NotSet,
+        };
+        fruit.set_from_json(json!({
+            "id": 8,
+            "name": "Apple",
+            "cake_id": 1,
+        }))?;
+        assert_eq!(
+            fruit,
+            fruit::ActiveModel {
+                id: ActiveValue::Set(1),
                 name: ActiveValue::Set("Apple".to_owned()),
                 cake_id: ActiveValue::Set(Some(1)),
             }
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -483,6 +483,53 @@ pub trait ActiveModelTrait: Clone + Debug {
         ActiveModelBehavior::after_delete(am_clone)?;
         Ok(delete_res)
     }
+
+    /// Set the corresponding attributes in the ActiveModel from a JSON value
+    ///
+    /// Note that this method will not alter the primary key values in ActiveModel.
+    #[cfg(feature = "with-json")]
+    fn set_from_json(&mut self, json: serde_json::Value) -> Result<(), DbErr>
+    where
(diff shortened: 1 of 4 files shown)


billy1624:
But instead of setting columns one by one, why don't we convert a custom deserializable Model into ActiveModel?

https://www.sea-ql.org/SeaORM/docs/advanced-query/custom-active-model

qyihua:
> But instead of setting columns one by one, why don't we convert a custom deserializable Model into ActiveModel? https://www.sea-ql.org/SeaORM/docs/advanced-query/custom-active-model

Hi @billy1624  Thanks for your suggestion. And Sorry for the delay
The `Custom Active Model` is very helpful for most case of mine.
But in some use case, the keys in json is dynamic at runtime, `Custom Active Model`  can not satisfy.
Do you have any thought of this?

billy1624:
Hey @qyihua, if you wanna do it dynamically. I guess you need to do it via SeaQuery.

```rust
let mut stmt = Query::insert();

stmt.into_table(entity::Entity)
    .columns(cols); // List of columns

for node_json in node_json_batch.into_iter() { // Iterate array of JSON object
    let mut vals = vec![];
    for attribute in attributes.iter() { // Iterate attributes in JSON object
        let val = node_json.get(attribute).unwrap();
        vals.push(val);
    }
    stmt.values_panic(vals);
}

let builder = db.get_database_backend();
let mut stmt = builder.build(&stmt);
db.execute(stmt).await?;
```

qyihua:
@billy1624 Thank you
 
``` rust
stmt.into_table(entity::Entity)
    .columns(cols); // List of columns
```
But `sea_query` need provide special columns in rust code.
So I write a helper trait for my use case:
```rust
pub trait SetFromJson
where
    Self: ActiveModelTrait,
{
    fn set_from_json(&mut self, json: &mut JsonValue) -> Result<(), ConvertError>;
}

impl<T> SetFromJson for T
where
    T: ActiveModelTrait,
{
    fn set_from_json(&mut self, json: &mut JsonValue) -> Result<(), ConvertError> {
        for col in <<Self::Entity as EntityTrait>::Column>::iter() {
            if let Some(json_val) = json.get_mut(col.as_str()) {
                self.set_col_from_json(col, json_val.take())
                    .map_err(|err| {
                        ConvertError::Other(format!(
                            "col: `{}`, err: {}",
                            col.as_str(),
                            err.to_string()
                        ))
                    })?;
            }
        }
        Ok(())
    }
}
```

src/entity/active_model.rs
```diff
@@ -843,4 +908,175 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    #[cfg(feature = "with-json")]
+    #[should_panic(
+        expected = r#"called `Result::unwrap()` on an `Err` value: Json("missing field `id`")"#
```
tyt2y3, line 915: I don't understand why this should panic where `test_active_model_set_from_json_1` does not

billy1624: `test_active_model_set_from_json_1` will fail because `cake::Model` doesn't specify `#[serde(skip_deserializing)]`.

https://github.com/SeaQL/sea-orm/blob/b32e0227168b2455aa78ab172d2c93fb9fa41f3e/src/tests_cfg/cake.rs#L7-L15

While `fruit::Model` specify it and prevent the error in `test_active_model_set_from_json_2`.

https://github.com/SeaQL/sea-orm/blob/b32e0227168b2455aa78ab172d2c93fb9fa41f3e/src/tests_cfg/fruit.rs#L7-L16

tyt2y3: got it

PR #494, opened 2022-01-29
Remove `r#` prefix when deriving `FromQueryResult`

smrtrfszm:
When deriving `FromQueryResult` on a struct that has a field with a raw identifier, then in the expanded code the `r#` prefix isn't stripped.
This is inconsistent with the behaviour of the `DeriveEntityModel` because that macro removes it.

Example:
```rust
#[derive(Debug, FromQueryResult)]
pub struct Example {
    pub r#type: i32,
}
```
Expands to:
```rust
#[automatically_derived]
impl sea_orm::FromQueryResult for Example {
    fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> Result<Self, sea_orm::DbErr> {
        Ok(Self {
            r#type: row.try_get(pre, "r#type")?,
        })
    }
}
```
But the column name in the `try_get` should be `type` without the `r#` prefix.

### DIFF
--- a/sea-orm-macros/src/derives/from_query_result.rs
+++ b/sea-orm-macros/src/derives/from_query_result.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote, quote_spanned};
-use syn::{Data, DataStruct, Field, Fields};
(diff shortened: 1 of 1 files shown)


tyt2y3:
thanks for the catch! @billy1624 I imagine we can also add some test cases using the MockInterface?

PR #495, opened 2022-01-30
Fix FromQueryResult when Result is redefined

tasn:
Macros should only ever use absolute import paths.
Shorthands can clash (as they have in my case).

### DIFF
--- a/sea-orm-macros/src/derives/active_enum.rs
+++ b/sea-orm-macros/src/derives/active_enum.rs
@@ -211,7 +211,7 @@ impl ActiveEnum {
                     .to_owned()
                 }
 
-                fn try_from_value(v: &Self::Value) -> Result<Self, sea_orm::DbErr> {
+                fn try_from_value(v: &Self::Value) -> std::result::Result<Self, sea_orm::DbErr> {
                     match #val {
                         #( #variant_values => Ok(Self::#variant_idents), )*
                         _ => Err(sea_orm::DbErr::Type(format!(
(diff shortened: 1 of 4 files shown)


tasn:
Still not a complete fix (seems to still be failing for me) with the same reason:
```
error[E0107]: this type alias takes 1 generic argument but 2 generic arguments were supplied
   --> svix-server/src/v1/endpoints/endpoint.rs:144:10
    |
144 | #[derive(FromQueryResult)]
    |          ^^^^^^^^^^^^^^^
    |          |
    |          expected 1 generic argument
    |          help: remove this generic argument
    |
note: type alias defined here, with 1 generic parameter: `T`
   --> svix-server/src/error.rs:16:10
    |
16  | pub type Result<T> = std::result::Result<T, Error>;
    |          ^^^^^^ -
    = note: this error originates in the derive macro `FromQueryResult` (in Nightly builds, run with -Z macro-backtrace for more info)
```

billy1624:
Hey @tasn, thank you for fixing this. We should have fully quantify it, I have go through all procedural macros and fixed some more of it. Please check https://github.com/tasn/sea-orm/compare/patch-1...billy1624:pr/495 and cherrypick it :P

billy1624:
We could merge and release this in 0.6.0
CC @tyt2y3

billy1624 (approved):
Thanks!! @tasn

tyt2y3 (approved):

PR #508, opened 2022-02-07
time support

cemoktra:
Adding support for time crate. Note that sea-query dependency goes to my branch and i currently defaulted to time crate for CI


billy1624:
Why this was closed?

cemoktra:
Its not that easy as i thaught and needs more work, ltree has my prio now at the moment

billy1624:
Okay! We could revisit this later :P

PR #523, opened 2022-02-13
Operations could take connection with unknown size

### LINKED ISSUE #522: How do you seed data in migrations using ActiveModels?
### Discussed in https://github.com/SeaQL/sea-orm/discussions/519

<div type='discussions-op-text'>

<sup>Originally posted by **Drakota** February 13, 2022</sup>
I'm trying to seed data directly inside a migration, but I'm unable to get a DatabaseConnection handle to be able to do it.
```rust

let board = board::ActiveModel {
    name: Set("Test Board".to_owned()),
    ..Default::default()
};

board.save(db).await?;
```
I tried something like this, but I'm getting:
`the size for values of type 'dyn sea_schema::migration::sea_orm::ConnectionTrait' cannot be known at compilation time the trait 'std::marker::Sized' is not implemented for 'dyn sea_schema::migration::sea_orm::ConnectionTrait'`. </div>

billy1624:
## PR Info

- Closes #522

- Dependencies:
  - SeaQL/sea-schema#45

## Fixes

- Methods that take in `ConnectionTrait` is implied to be with known sized, aka `ConnectionTrait + Sized`. But this assumption can be removed.

## Breaking Changes

- Methods that take in `ConnectionTrait`, now take in `ConnectionTrait + ?Sized`

### DIFF
--- a/examples/rocket_example/entity/Cargo.toml
+++ b/examples/rocket_example/entity/Cargo.toml
@@ -15,6 +15,8 @@ rocket = { version = "0.5.0-rc.1", features = [
 
 [dependencies.sea-orm]
 # path = "../../../" # remove this line in your own project
+git = "https://github.com/SeaQL/sea-orm.git"
+branch = "conn-relax-sized-bound"
 version = "^0.6.0"
 features = [
   "macros",
--- a/examples/rocket_example/migration/Cargo.toml
+++ b/examples/rocket_example/migration/Cargo.toml
@@ -9,6 +9,6 @@ name = "migration"
 path = "src/lib.rs"
 
 [dependencies]
-sea-schema = { version = "^0.5.0", default-features = false, features = [ "migration", "debug-print" ] }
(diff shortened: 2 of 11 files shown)


tyt2y3:
I don't think it is an appropriate fix.
Should really try to fix it on SeaSchema side.
The root of the problem is the use of `dyn` vs use of `enum` for polymorphism, and SeaORM chooses not to use `dyn`

billy1624:
I think we should keep the SeaSchema migrator flexible and allow it to take anything that implemented `ConnectionTrait`. This way user can migrate with an ordinary connection or a transaction.

The SeaSchema Migration Manager: https://github.com/SeaQL/sea-schema/blob/db674ed28daff9bbfa3be51519f98ebaebf591b4/src/migration/manager.rs#L13

Also, in this PR, I relaxed the trait bound inside SeaORM, and I think the impact is minimum.

tyt2y3:
I only think that SeaSchema can use generics and enums instead of a dyn, unless there is clear reasons we can't

tyt2y3:
It definitely is something we will look into - on breaking down SeaORM into core, driver & frontend crates.

PR #531, opened 2022-02-15
Testing inserts on CockroachDB

### LINKED ISSUE #507: CockroachDB support
Sqlx has a known issue with transations and MARS
https://github.com/launchbadge/sqlx/issues/933

Does sea orm has a workaround to do multiple inserts inside single transaction with CockroachDB?

tokio-postgres has no problems, so it might be fixed on client side somehow without waiting for https://github.com/cockroachdb/cockroach/issues/40195

billy1624:
## PR Info

- Closes #507

## Summary

Try SeaORM inserts on CockroachDB.

Tracking issue: cockroachdb/cockroach#40195

Could workaround it as suggested on launchbadge/sqlx#933, we can use `fetch_all` instead of `fetch_one` as a temporary solution. However, this impose additional performance penalty for other databases that don't require this workaround.

### DIFF
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::*, ActiveModelTrait, ColumnTrait, ConnectionTrait, EntityTrait, Insert, IntoActiveModel,
-    Iterable, PrimaryKeyTrait, QuerySelect, SelectModel, SelectorRaw, Statement, TryFromU64,
+    Iterable, PrimaryKeyTrait, SelectModel, SelectorRaw, Statement, TryFromU64,
 };
 use sea_query::{
     Alias, Expr, FromValueTuple, Iden, InsertStatement, IntoColumnRef, Query, ValueTuple,
@@ -118,8 +118,15 @@ where
             let cols = PrimaryKey::<A>::iter()
                 .map(|col| col.to_string())
                 .collect::<Vec<_>>();
+
+            #[cfg(not(feature = "cockroachdb"))]
+            let res = db.query_one(statement).await?.unwrap();
+
+            #[cfg(feature = "cockroachdb")]
             let vec = db.query_all(statement).await?;
+            #[cfg(feature = "cockroachdb")]
             let res = vec.get(0).unwrap();
+
             res.try_get_many("", cols.as_ref()).ok()
         }
         false => {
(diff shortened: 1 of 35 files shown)


src/executor/insert.rs
```diff
@@ -118,7 +118,8 @@ where
             let cols = PrimaryKey::<A>::iter()
                 .map(|col| col.to_string())
                 .collect::<Vec<_>>();
-            let res = db.query_one(statement).await?.unwrap();
+            let vec = db.query_all(statement).await?;
+            let res = vec.get(0).unwrap();
```
billy1624, line 122: Avoid `fetch_one`

src/executor/insert.rs
```diff
@@ -163,19 +164,20 @@ where
...
+            .all(db)
             .await?
         }
         false => {
             let insert_res =
                 exec_insert::<A, _>(primary_key, db_backend.build(&insert_statement), db).await?;
             <A::Entity as EntityTrait>::find_by_id(insert_res.last_insert_id)
-                .one(db)
+                .limit(1)
+                .all(db)
```
billy1624, line 175: And, here

billy1624 (commented):
Hey @sergeyshaykhullin, I get it work by changing these lines in SeaORM core.

sergeyshaykhullin:
Does this fixes can be a part of 0.7 release? Should it be a separate feature like "postgres", but "cockroachdb"?

billy1624:
> Does this fixes can be a part of 0.7 release? Should it be a separate feature like "postgres", but "cockroachdb"?

I think we could have release it in 0.7.0. However, we need to write CI tests to make sure it works on a live CockroachDB.

@tyt2y3 thoughts?

tyt2y3:
Yes. I think we definitely need automated tests setup before releasing.
I think the best approach in the long term is to treat CockroachDB as a separate backend from Postgres.

billy1624:
Added CockroachDB tests and `cockroachdb` feature

CC @tyt2y3

ducc:
Hey @billy1624 I've just tested this by crudely copy pasting the src/executor/insert.rs changes into my own fork (https://github.com/duccltd/sea-orm/commit/23f81f462c6574614099f1736602baa9bb4b0870).

My problem was that I was unable to insert a row, however these changes fix it for me!

billy1624:
Hey @ducc, thanks for testing it!

PR #545, opened 2022-02-18
sqlx ltree/lquery support

cemoktra:
## PR Info

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/264
  - https://github.com/SeaQL/sea-schema/pull/46

## Adds

sqlx ltree support


cemoktra:
@billy1624 can you help identify why this unit test fails? i think the runtime is specified

PR #558, opened 2022-02-26
`sea` as an alternative bin name to `sea-orm-cli`

ZhangHanDong:
I feel that the command `sea-orm-cli` is too long and inconvenient to use, so it should be more convenient to simplify it to `sea` now. 

What do you think?

## Changes

The `sea` command is now an option, and a new bin file has been added to achieve this.

The only downside: `bin/sea.rs` is a duplicate of `bin/main.rs`, but it has now been modified to have only one main function, so if there are changes, they are simply copied to `sea.rs`. It won't be too much trouble to maintain.

Now `cargo run` still defaults to `sea-orm-cli`

Install and Usage: 

```sh
> cargo install sea-orm-cli 
> sea-orm-cli help
```

Or: 

```sh
> cargo install --bin sea
> sea help
```

### DIFF
--- a/sea-orm-cli/Cargo.toml
+++ b/sea-orm-cli/Cargo.toml
@@ -12,10 +12,20 @@ documentation = "https://docs.rs/sea-orm"
 repository = "https://github.com/SeaQL/sea-orm"
 categories = [ "database" ]
 keywords = ["async", "orm", "mysql", "postgres", "sqlite"]
+default-run = "sea-orm-cli"
+
+
+[lib]
+name = "sea_orm_cli"
+path = "src/lib.rs"
+
+[[bin]]
+name = "sea-orm-cli"
+path = "src/bin/main.rs"
 
 [[bin]]
 name = "sea"
-path = "src/main.rs"
+path = "src/bin/sea.rs"
 
 [dependencies]
 clap = { version = "^2.33.3" }
--- a/sea-orm-cli/Cargo.toml
+++ b/sea-orm-cli/Cargo.toml
@@ -14,7 +14,7 @@ categories = [ "database" ]
 keywords = ["async", "orm", "mysql", "postgres", "sqlite"]
 
 [[bin]]
-name = "sea-orm-cli"
+name = "sea"
 path = "src/main.rs"
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


tyt2y3:
I think this is a reasonable thought. `sea-orm-cli` is more canonical, but `sea` is a perfectly fine shorthand. Is it possible to make it an alias?

tyt2y3:
That's interesting! Certainly better than a breaking name change. I also think we need to set the default binary, other `cargo run` will confuse

ZhangHanDong:
> That's interesting! Certainly better than a breaking name change. I also think we need to set the default binary, other `cargo run` will confuse
This has the minor drawback of reporting a compile warning at compile time: 
```rust
warning: /Projects/sea-orm/sea-orm-cli/Cargo.toml: file found to be present in multiple build targets
```

ZhangHanDong:
@tyt2y3 

The `sea` command is now an option, and a new bin file has been added to achieve this.

The only downside: `bin/sea.rs` is a duplicate of `bin/main.rs`, but it has now been modified to have only one main function, so if there are changes, they are simply copied to `sea.rs`. It won't be too much trouble to maintain.

Now `cargo run` still defaults to `sea-orm-cli`

Install and Usage: 

```sh
> cargo install sea-orm-cli 
> sea-orm-cli help
```

Or: 

```sh
> cargo install --bin sea
> sea help
```

ZhangHanDong:
@tyt2y3  what do you think?

tyt2y3:
Awesome

ZhangHanDong:
@tyt2y3  Fixed CI compile error

ZhangHanDong:
@tyt2y3  need approve workflow

ZhangHanDong:
@tyt2y3 

Can't it be merged yet? Is there anything that needs to be changed?

tyt2y3 (approved):
LGTM

ZhangHanDong:
@tyt2y3 

P.S Can't seaorm automatically generate migration files now?

billy1624:
> @tyt2y3
> 
> P.S Can't seaorm automatically generate migration files now? Except for the `sea migrate init`.

Not supported at the moment. I think you want something like `sea migrate new name_of_new_migration_file` which create a new migration file?

ZhangHanDong:
> 

Yes. Like the `rails` migrate command

PR #562, opened 2022-03-01
Exclude `mock` from default features

### LINKED ISSUE #438: Make `DatabaseConnection` `Clone` with the default features enabled
Some frameworks like actix_web require app global datas to be clonable. Currently one can only disable the `mock` feature to fix this. It would be great if `sea_orm::DatabaseConnection` could be clonable in all cases.

Currently the `mock` feature is by default enabled and that causes the `sea_orm::DatabaseConnection` to not implement `Clone` by default, because `MockDatabaseConnection` (which is a variant of `DatabaseConnection`) is not clonable.

### LINKED ISSUE #517: sea_orm::DatabaseConnection should implement `Clone` by default
Because `MockConnection` is under an `Arc` - `DatabaseConnection` can implement `Clone` by default
https://github.com/SeaQL/sea-orm/blob/f418c4e58069cb7ae42b79c721a44c8bb8915f1a/src/database/db_connection.rs#L18
https://github.com/SeaQL/sea-orm/blob/f418c4e58069cb7ae42b79c721a44c8bb8915f1a/src/database/db_connection.rs#L31

billy1624:
## PR Info

- Closes #517
- Closes #438

## Breaking Changes

- Remove `mock` from default feature
  - This means user who are using `mock` utilities has to opt-in for it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["database"]
 keywords = ["async", "orm", "mysql", "postgres", "sqlite"]
 
 [package.metadata.docs.rs]
-features = ["default", "sqlx-all", "runtime-async-std-native-tls"]
+features = ["default", "sqlx-all", "mock", "runtime-async-std-native-tls"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [lib]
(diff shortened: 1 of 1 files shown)


tyt2y3 (approved):
Nice

Issue #403, opened 2021-12-25
Who's using SeaORM?


tyt2y3:
If you have built an app using SeaORM and want to showcase it, feel free to open a PR and add it to the list below!

https://github.com/SeaQL/sea-orm/blob/master/COMMUNITY.md#built-with-seaorm

tyt2y3:
@Sytten Added Caido 😄

PR #569, opened 2022-03-02
Add mediarepo to showcase

Trivernis:
## PR Info

#403 

## Changes

- Added mediarepo to showcase


billy1624 (approved):
Thanks!! @Trivernis

Issue #552, opened 2022-02-21
Update / Delete by id


Ilqjx:
Is there something like `find_by_id(1).one(&db)` for `update` and `delete`?

Look like `update_by_id(1).one(&db)` and `delete_by_id(1).one(&db)`.

billy1624:
Hey @Ilqjx, interesting idea!

For delete, I think it's make sense to delete by id. I support adding a new API for it.

For update, I would say it's natural to select the ActiveModel from db then alter the row then save it back to db.

```rust
let orange: Option<fruit::Model> = Fruit::find_by_id(30).one(db).await?;
let orange: fruit::Model = orange.unwrap();

let res: DeleteResult = orange.delete(db).await?;
assert_eq!(res.rows_affected, 1);
```

Or, update any rows exists that match specific conditions.

```rust
// Bulk set attributes using ActiveModel
let update_result: UpdateResult = Fruit::update_many()
    .set(pear)
    .filter(fruit::Column::Id.eq(1))
    .exec(db)
    .await?;

// UPDATE `fruit` SET `cake_id` = NULL WHERE `fruit`.`name` LIKE '%Apple%'
Fruit::update_many()
    .col_expr(fruit::Column::CakeId, Expr::value(Value::Null))
    .filter(fruit::Column::Name.contains("Apple"))
    .exec(db)
    .await?;
```

I'm not sure how `update_by_id(1).one(&db)` would work? Can you explain more with it?

Ilqjx:
Indeed, delete are useful and update are unnecessary.

billy1624:
I will open this issue for contributions :)

AbhijithGanesh:
I'd like to work on this @billy1624

billy1624:
Hey @AbhijithGanesh, thanks for the interest!!

The `delete_by_id` API is under development by @ShouvikGhosh2048.

I saw your message on Discord, you want to work on the `update_by_id` API. However, as discussed above. Seems that it's not necessary to have the API. I wonder if you got any other thoughts on it?

AbhijithGanesh:
I agree with llqjx

PR #590, opened 2022-03-10
Added delete_by_id

ShouvikGhosh2048:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/552

## Adds

Adds a function delete_by_id to delete by primary key.

## Fixes

None
## Breaking Changes

None

## Changes

Added delete_by_id + a unit test.

### DIFF
--- a/src/entity/base_entity.rs
+++ b/src/entity/base_entity.rs
@@ -745,10 +745,77 @@ pub trait EntityTrait: EntityName {
     fn delete_many() -> DeleteMany<Self> {
         Delete::many(Self::default())
     }
+
+    /// Delete a model based on primary key
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use sea_orm::{error::*, tests_cfg::*, *};
+    /// #
+    /// # #[smol_potat::main]
+    /// # #[cfg(feature = "mock")]
+    /// # pub async fn main() -> Result<(), DbErr> {
+    /// #
+    /// # let db = MockDatabase::new(DbBackend::Postgres)
+    /// #     .append_exec_results(vec![
+    /// #         MockExecResult {
+    /// #             last_insert_id: 0,
+    /// #             rows_affected: 1,
+    /// #         },
+    /// #     ])
+    /// #     .into_connection();
+    /// #
+    /// use sea_orm::{entity::*, query::*, tests_cfg::fruit};
+    ///
+    /// let delete_result = fruit::Entity::delete_by_id(1).exec(&db).await?;
+    ///
+    /// assert_eq!(delete_result.rows_affected, 1);
+    ///
+    /// assert_eq!(
+    ///     db.into_transaction_log(),
+    ///     vec![Transaction::from_sql_and_values(
+    ///         DbBackend::Postgres,
+    ///         r#"DELETE FROM "fruit" WHERE "fruit"."id" = $1"#,
+    ///         vec![1i32.into()]
+    ///     )]
+    /// );
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn delete_by_id(values: <Self::PrimaryKey as PrimaryKeyTrait>::ValueType) -> DeleteMany<Self> {
+        let mut delete = Self::delete_many();
+        let mut keys = Self::PrimaryKey::iter();
+        for v in values.into_value_tuple() {
+            if let Some(key) = keys.next() {
+                let col = key.into_column();
+                delete = delete.filter(col.eq(v));
+            } else {
+                panic!("primary key arity mismatch");
+            }
+        }
+        if keys.next().is_some() {
+            panic!("primary key arity mismatch");
+        }
+        delete
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    #[test]
+    fn test_delete_by_id() {
+        use crate::tests_cfg::cake;
+        use crate::{entity::*, query::*, DbBackend};
+        assert_eq!(
+            cake::Entity::delete_by_id(1)
+                .build(DbBackend::Sqlite)
+                .to_string(),
+            r#"DELETE FROM "cake" WHERE "cake"."id" = 1"#,
+        );
+    }
+    
     #[test]
     #[cfg(feature = "macros")]
     fn entity_model_1() {
--- a/src/entity/base_entity.rs
+++ b/src/entity/base_entity.rs
@@ -840,7 +840,7 @@ pub trait EntityTrait: EntityName {
 #[cfg(test)]
 mod tests {
     #[test]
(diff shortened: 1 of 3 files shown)


src/entity/base_entity.rs
```diff
@@ -745,10 +745,77 @@ pub trait EntityTrait: EntityName {
...
+        use crate::tests_cfg::cake;
+        use crate::{entity::*, query::*, DbBackend};
+        assert_eq!(
+            cake::Entity::delete_by_id(1)
+                .build(DbBackend::Sqlite)
+                .to_string(),
+            r#"DELETE FROM "cake" WHERE "cake"."id" = 1"#,
+        );
+    }
+
```
AbhijithGanesh, line 818: Is there a reason as to why `"cake"."id"` is written separately in quotation marks?

src/entity/base_entity.rs
```diff
@@ -745,10 +745,77 @@ pub trait EntityTrait: EntityName {
...
+        for v in values.into_value_tuple() {
+            if let Some(key) = keys.next() {
+                let col = key.into_column();
+                delete = delete.filter(col.eq(v));
+            } else {
+                panic!("primary key arity mismatch");
+            }
+        }
+        if keys.next().is_some() {
+            panic!("primary key arity mismatch");
```
AbhijithGanesh, line 799: Is it supposed to be parity or arity ?

src/entity/base_entity.rs
```diff
@@ -745,10 +745,77 @@ pub trait EntityTrait: EntityName {
...
+    /// ```
+    fn delete_by_id(values: <Self::PrimaryKey as PrimaryKeyTrait>::ValueType) -> DeleteMany<Self> {
+        let mut delete = Self::delete_many();
+        let mut keys = Self::PrimaryKey::iter();
+        for v in values.into_value_tuple() {
+            if let Some(key) = keys.next() {
+                let col = key.into_column();
+                delete = delete.filter(col.eq(v));
+            } else {
+                panic!("primary key arity mismatch");
```
AbhijithGanesh, line 795: Will this actually delete the code, I understand this might be inspired from `find_by_id` but I am not sure if this deletes the case

src/entity/base_entity.rs
ShouvikGhosh2048: I think [arity](https://en.wikipedia.org/wiki/Arity) is fine.

src/entity/base_entity.rs
ShouvikGhosh2048: I used [this](https://github.com/SeaQL/sea-orm/blob/0a26dee674a00cdda4cf3568f3d89a76fb5c8f57/src/query/delete.rs#L182-L219) as a reference.

src/entity/base_entity.rs
ShouvikGhosh2048: The example code and the unit test work, but I'm not entirely sure.

src/entity/base_entity.rs
AbhijithGanesh: Arity is fine. The conversation can be resolved.

src/entity/base_entity.rs
AbhijithGanesh: I'll tell why, find by ID  might be able to pull the data but delete is to remove data right? That kind of makes thing skeptical. I 'd also like to know if you can make it `delete_one` instead of `delete_many`

AbhijithGanesh: Please use delete_one @ShouvikGhosh2048

ShouvikGhosh2048: DeleteOne needs an a type that implements ActiveModelTrait. We're only passed an id.

src/entity/base_entity.rs
billy1624: Quotation marks were added around identifier, `"cake"` is a identifier as it's a table name and `"id"` is a identifier as well as it's a column name.
See https://www.sqlite.org/lang_keywords.html for more

AbhijithGanesh: I am aware about the identifier patterns, I don't think its a common pattern with Postgres/MySQL thats why I wasn't expecting this

tests/delete_by_id_tests.rs
```diff
@@ -0,0 +1,60 @@
...
+    assert_eq!(delete_res.rows_affected, 1);
+
+    let find_res = Applog::find_by_id(1).all(db).await?;
+    assert_eq!(find_res, vec![log1]);
+
+    let find_res = Applog::find_by_id(2).all(db).await?;
+    assert_eq!(find_res, vec![]);
+
+    let delete_res = Applog::delete_by_id(3).exec(db).await?;
+    assert_eq!(delete_res.rows_affected, 0);
```
AbhijithGanesh, line 57: Looks good to me.

AbhijithGanesh (approved):

tests/delete_by_id_tests.rs
```diff
@@ -0,0 +1,60 @@
...
+        json: Json::String("HELLO".to_owned()),
+        created_at: "2022-09-17T17:50:20+08:00".parse().unwrap(),
+    };
+
+    Applog::insert(log2.clone().into_active_model())
+        .exec(db)
+        .await?;
+
+    let find_res = Applog::find_by_id(2).all(db).await?;
+    assert_eq!(find_res, vec![log2]);
```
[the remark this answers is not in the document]
ShouvikGhosh2048: I'll remove it.

AbhijithGanesh:
LGTM👍

tests/delete_by_id_tests.rs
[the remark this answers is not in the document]
AbhijithGanesh: Resolved.

src/entity/base_entity.rs
```diff
@@ -745,10 +745,112 @@ pub trait EntityTrait: EntityName {
...
+            panic!("primary key arity mismatch");
+        }
+        delete
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    #[test]
+    fn test_delete_by_id() {
```
billy1624, line 843: Thanks!! @ShouvikGhosh2048
Can we have a unit test for entity with composite primary key? You can use [`cake_filling_price`](https://github.com/SeaQL/sea-orm/blob/master/src/tests_cfg/cake_filling_price.rs) entity

ShouvikGhosh2048: Added

billy1624: Thanks!!

billy1624 (approved):
Hey @ShouvikGhosh2048, thanks for contributing! This looks good!

tyt2y3 (approved):

Issue #554, opened 2022-02-23
`#[sea_orm(indexed)]` only works for MySQL


grtlr:
## Description

Using the `#[sea_orm(indexed)]` property only works for the MySQL backend and crashes when using sqlite.

## Steps to Reproduce

Put the `#[sea_orm(indexed)]` property on any field of a struct that derives `DeriveEntityModel` while using a sqlite backend.

### Expected Behavior

Ideally it would create the index, but at least it shouldn't crash, emitting an error at compile time.

### Actual Behavior

It crashes.

### Reproduces How Often

Everytime 😉.

## Versions

I'm running `sea-orm v0.6.0`.

## Additional Information

It might be helpful to emphasize this in the appropriate sections of the documentation.

nickb937:
I'd opened a similar issue, incorrectly against sea-query, too : https://github.com/SeaQL/sea-query/issues/232

billy1624:
Thanks!! @grtlr

> It might be helpful to emphasize this in the appropriate sections of the documentation.

As discussed on Discord the first thing I would do was this.

tyt2y3:
We should separate the logic from create_table_from_entity and extract to a new function create_index_from_entity

PR #593, opened 2022-03-11
Allow for creation of indexes for PostgeSQL and SQLite

nickb937:
PostgreSQL and SQLite do not allow creation of general indexes within a `CREATE TABLE` statement, so a method is required to generate `CREATE INDEX` statements for these.

`create_table_from_entity` avoids creating invalid statements for non-MySQL backends,
forcing uses to explicitly run `create_index_from_entity`.  Ideally creating indexes would be removed from `create_table_from_entity` entirely, but this would introduce a breaking change for MySQL use.

## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/554

### DIFF
--- a/src/tests_cfg/cake_filling_price.rs
+++ b/src/tests_cfg/cake_filling_price.rs
@@ -22,6 +22,8 @@ pub struct Model {
     pub price: Decimal,
     #[sea_orm(ignore)]
     pub ignored_attr: i32,
+    pub indexed_attr: i32,
+    pub unique_indexed_attr: i32,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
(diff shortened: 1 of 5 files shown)


tyt2y3:
> Ideally creating indexes would be removed from create_table_from_entity entirely, but this would introduce a breaking change for MySQL use.

I agree with this change

src/tests_cfg/cake_filling_price.rs
```diff
@@ -22,13 +22,17 @@ pub struct Model {
     pub price: Decimal,
     #[sea_orm(ignore)]
     pub ignored_attr: i32,
+    pub indexed_attr: i32,
```
tyt2y3, line 25: I would very much appreciate that we do not add columns to this Entity.
May be we can test it by having multiple test cases, with each test a different behaviour

tyt2y3 (changes requested):

tyt2y3:
Thank you for the contribution all-in-all

src/tests_cfg/cake_filling_price.rs
nickb937: In the update, I have created a new standalone Entity for the purpose of testing index creation

tyt2y3 (approved):

PR #602, opened 2022-03-13
Supports `time` crate

### LINKED ISSUE #499: Support the time crate in addition (instead of?) chrono
It looks like chrono has security issues that can't be fixed. It depends on an old version of the `time` library which is good on its own anyway. Should we add support for the time crate?

Is there a way for me to do it in our own code before support hits sea-orm?

More info: https://github.com/svix/rust-ksuid/issues/1

billy1624:
## PR Info

- Closes #499
- Dependencies:
  - SeaQL/sea-query#267

## Adds

- Supports `time` crate


src/entity/prelude.rs
```diff
@@ -36,6 +36,39 @@ pub type DateTimeUtc = chrono::DateTime<chrono::Utc>;
 #[cfg(feature = "with-chrono")]
 pub type DateTimeLocal = chrono::DateTime<chrono::Local>;
 
+#[cfg(feature = "with-chrono")]
```
AbhijithGanesh, line 39: is there an abstract way which would do this faster rather than importing the `#[cfg()]` macro always.

tests/time_crate_tests.rs
```diff
@@ -0,0 +1,46 @@
...
+pub async fn create_transaction_log(db: &DatabaseConnection) -> Result<(), DbErr> {
+    let transaction_log = transaction_log::Model {
+        id: 1,
+        date: date!(2022 - 03 - 13),
+        time: time!(16:24:00),
+        date_time: date!(2022 - 03 - 13).with_time(time!(16:24:00)),
+        date_time_tz: date!(2022 - 03 - 13)
+            .with_time(time!(16:24:00))
+            .assume_utc(),
+    };
```
AbhijithGanesh, line 33: Cool, changes. Good work👍

tasn:
Yay, thanks a lot!

Issue #591, opened 2022-03-10
[ADD]: Pre-commit hook to keep the project clean


AbhijithGanesh:
## Summary

One can run a script to verify whether their changes are working, do they pass the automated tests etc?

## Motivation

Integration tests

billy1624:
Hey @AbhijithGanesh, we do have GitHub Actions for running unit tests and integration tests.
https://github.com/SeaQL/sea-orm/actions

AbhijithGanesh:
@billy1624 I meant, local pre-commit hooks for a smoother dev flow

AbhijithGanesh:
Apologies, closed by mistake.

billy1624:
> Apologies, closed by mistake.

Hahaa nvm!

billy1624:
I wonder what will be checked inside the pre-commit hooks? For example, running `cargo fmt`?

maltevesper:
Unfortunately I have not seen #611 before amking this.
I think there is a slight conflict now, i will touch on this under the actual PR.

billy1624:
Hey @AbhijithGanesh, sorry for the conflict. I think similar code has been merged
- https://github.com/SeaQL/sea-orm/pull/840
- https://github.com/SeaQL/sea-orm/pull/919

Please check if you have anything to add.

AbhijithGanesh:
Thanks @billy1624

billy1624:
Hey @AbhijithGanesh, please close this issue and PR (https://github.com/SeaQL/sea-orm/pull/611) if you think the merged PR is good enough and you have nothing to add.

Once again, thank you so much for the contributions :)

AbhijithGanesh:
I think this [PR](https://github.com/SeaQL/sea-orm/pull/611) is still open which can be merged. It's a pre commit check script which I think was good for developers to run locally.

PR #611, opened 2022-03-15
Addition of DCO Workflow

### LINKED ISSUE #610: [ADD]: Addition of DCO
## Summary

It'd be good if we implement DCO (Developer certificate of Origin) for more accountability.

## Motivation

https://github.com/apps/dco

## Additional Information
N/A

AbhijithGanesh:
Signed-off-by: Abhijith Ganesh <67182544+AbhijithGanesh@users.noreply.github.com>

## PR Info

- Closes #610 

- Dependencies:
  - 

- Dependents:
  - 

## Fixes

None

## Breaking Changes

None

## Changes
Adds DCO Workflow to the project

### DIFF
--- a/.github/workflows/dco.yaml
+++ b/.github/workflows/dco.yaml
@@ -1,0 +1,17 @@
+name: Sanity check
+on: [pull_request]
+
+jobs:
+  commits_check_job:
+    runs-on: ubuntu-latest
+    name: Commits Check
+    steps:
+    - name: Get PR Commits
+      id: 'get-pr-commits'
+      uses: tim-actions/get-pr-commits@master
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+    - name: DCO Check
+      uses: tim-actions/dco@master
+      with:
+        commits: ${{ steps.get-pr-commits.outputs.commits }}


billy1624 (commented):
Hey @AbhijithGanesh, as I said on the Discord. I would suggest not to enforce DCO with GitHub Action. Maybe we can just add a pre-commit hook to add the sign `-s` flag when running `git commit`.

AbhijithGanesh:
Is there a difference between DCO and `-s`? I thought they were the same thing. We agreed not to enforce GPG based commit signing

AbhijithGanesh:
According to the [docs](https://github.com/apps/dco) , I think -s and DCO are the same thing?

kakarot-dev:
im guessing what billy means is that instead of making the user add a `-s` flag to the commit, use a pre commit hook to add it automatically

pre commit hook is something like [husky](https://github.com/typicode/husky)

billy1624:
Thanks!! @kakarot-dev That's what I meant :P

kakarot-dev:
Well first off, Husky is a nodejs module, so i cant see how ur gonna use it in seaorm
Secondly, pre commit hooks can also me configured by “git config core.hooksPath <dir>“

My suggestion is to make a .sh doing the precommit, then add that using the above command during the initialization of the project

billy1624:
So, the pre-commit hook should be run on commit and without the need to config / enable it?

kakarot-dev:
> So, the pre-commit hook should be run on commit and without the need to config / enable it?

It wont run rn if you dont config it with `git config core.hooksPath`
But then again, that command works for a dir, @AbhijithGanesh can u tell how r u planning to make this work?

AbhijithGanesh:
@billy1624 any suggestions?

billy1624:
I think it's better to add the extra `-s` when performing git commit and leave testing for GitHub Actions to handle it

Btw... @tyt2y3 do you think we should enforce Developer Certificate of Origin (DCO)? https://github.com/apps/dco

AbhijithGanesh:
The script does add `-s`, in addition, the script runs couple of tests, thats it. If you want the tests removed, let me know!

kakarot-dev:
Tests r not optional tho…
Also how bout moving the scripts to a folder and adding some lines that specify how to set it up with git

AbhijithGanesh:
The tests are not optional because they are being run in the script. Do you want them optional?

AbhijithGanesh:
@billy1624 can you look at this please? Thanks

maltevesper:
> Hey @AbhijithGanesh, as I said on the Discord. I would suggest not to enforce DCO with GitHub Action. Maybe we can just add a pre-commit hook to add the sign `-s` flag when running `git commit`.

I would suggest that you check everything you want enforced in a pipeline, after all `git commit --no-verify` skips the hooks, or contributors might forget to install the hooks on a new machine.

Also `git config --global commit.gpgsign true` (not sure) might be an option, which goes in a similar direction like a hook. (Downside you need an ssh key, upside it is actually signed). Unfortunately I ahve not found a plain signoff  option that could be set.

That said, the reason I am writing here is #591 and the resulting #757 which was created before I was aware of this PR.

PR #632, opened 2022-03-24
add example for integrate with jsonrpsee

hunjixin:
just add example for integrate with jsonrpsee library


tyt2y3:
Thank you. This is extremely helpful!
@billy1624 can add that to our CI?

hunjixin:
@billy1624 updated

PR #633, opened 2022-03-24
Codegen PostgreSQL timestamp column as `DateTime` type

### LINKED ISSUE #586: Postgres timestamp column type
## Steps to Reproduce

1. Create table with column with type `timestamp`
2. Use cli to generate struct
3. Insert a value

### Behavior

```
Query Error: error occurred while decoding column "created_at": mismatched types; Rust type `core::option::Option<chrono::datetime::DateTime<chrono::offset::utc::Utc>>` (as SQL type `TIMESTAMPTZ`) is not compatible with SQL type `TIMESTAMP`
```

### Reproduces How Often

Always

## Versions

Postgres 14.1

```
├── sea-orm v0.6.0
│   ├── sea-orm-macros v0.6.0 (proc-macro)
│   ├── sea-query v0.21.0
│   │   ├── sea-query-derive v0.2.0 (proc-macro)
│   ├── sea-strum v0.23.0
│   │   └── sea-strum_macros v0.23.0 (proc-macro)
```

billy1624:
## PR Info

- Closes #586
- Closes SeaQL/sea-schema#50

## Fixes

- Consider database when deciding what is the corresponding Rust type for a db column
  - Maps PostgreSQL timestamp column as `DateTime` in Rust

### DIFF
--- a/sea-orm-cli/src/main.rs
+++ b/sea-orm-cli/src/main.rs
@@ -1,6 +1,6 @@
 use clap::ArgMatches;
 use dotenv::dotenv;
-use sea_orm_codegen::{EntityTransformer, OutputFile, WithSerde};
+use sea_orm_codegen::{DbBackend, EntityTransformer, OutputFile, WithSerde};
 use std::{error::Error, fmt::Display, fs, io::Write, path::Path, process::Command, str::FromStr};
 use url::Url;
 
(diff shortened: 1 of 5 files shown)


tyt2y3:
It's awkward because NaiveDateTime has the exact same layout as DateTime<Utc>, it's just a type system thing

billy1624:
> It's awkward because NaiveDateTime has the exact same byte format as `DateTime<Utc>`, it's just a type system thing

Perhaps we can consult SQLx team on this?

tyt2y3:
besides, I think passing in `db_backend` during codegen completely broke the abstraction

billy1624:
> besides, I think passing in `db_backend` during codegen completely broke the abstraction

But then how can we solve this db backend dependent type mapping?

Issue #618, opened 2022-03-17
sea_orm_cli: Enums could implement Copy


junbl:
## Summary

The enums generated by `sea_orm_cli` (those placed in `sea_orm_active_enums`, referencing databse enums) can derive copy, since they never have struct variants. 

## Motivation

The [language docs](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy) have this to say:
>Generally speaking, if your type can implement Copy, it should. Keep in mind, though, that implementing Copy is part of the public API of your type. If the type might become non-Copy in the future, it could be prudent to omit the Copy implementation now, to avoid a breaking API change.

If the databases were to support more complex enum variants, then this may no longer be true. However, this seems unlikely.

## Additional Information

To be safe, it could be toggled with an optional flag (cf. `--with-serde`).

billy1624:
Hey @e-rhodes, I agree with what you have suggested!

> The enums generated by `sea_orm_cli` (those placed in `sea_orm_active_enums`, referencing databse enums) can derive copy, since they never have struct variants.

Yes, this is true for ordinary user but we have to cater the need of advance users.

> To be safe, it could be toggled with an optional flag (cf. `--with-serde`).

I would say we can safely derive `Copy` for all generated Enums. For advance users, they simply have to update the Enums in `sea_orm_active_enums` file. And perhaps remove some of the `Copy` derive if they want to, it would be a easy "replace all" for them with the help of code editor.

Sounds good?

CC @tyt2y3

tyt2y3:
> If the type might become non-Copy in the future, it could be prudent to omit the Copy implementation now, to avoid a breaking API change.

Yes, I am still considering whether I'd extend the API by using tuple variants in enums

PR #634, opened 2022-03-25
Add Copy to the derived traits for active enums

e-rhodes:
## PR Info

- Closes [#618](https://github.com/SeaQL/sea-orm/issues/618)

## Adds

Added command line flag `--with-copy-enums` to `sea-orm-cli`. If provided, the automatically generated enums will derive the `Copy` trait. Otherwise, the behavior is unchanged.

## Breaking Changes

No changes to public API methods, CLI behavior is backwards compatible.

## Changes

Added `with_copy_enums` parameter in several places to get the behavior from the CLI to `sea_orm_codegen` (basically added the new flag anywhere the `with-serde` flag got passed).


tyt2y3:
Although as for now there is no reason those enums can't be `Copy` , I wanted them to feel `heavyweight` so they are only `Clone` but not copy.

tyt2y3:
I think the behaviour of adding `Copy` should be opt-in instead of a `default`

billy1624 (approved):
Hey @e-rhodes, sorry for the delay! Thanks for contributions!!

The opt-in `--with-copy-enums` flag look great!

@tyt2y3 has any comments on the naming?

tyt2y3:
Thank you. We may turn that on as default in the future!

billy1624:
> Thank you. We may turn that on as default in the future!

Hey @tyt2y3, I'll make `--with-copy-enums` enabled by default

PR #666, opened 2022-04-08
Dump SeaORM Dependency from SeaSchema's Migrator

billy1624:
## PR Info

- Closes SeaQL/sea-schema#57
- Dependencies:
  - SeaQL/sea-schema#59

## Adds

- All migration traits and structs are moved to SeaOMR except SchemaManager

## Breaking Changes

- `sea_schema::migration::prelude` should be replaced by `sea_orm::migration::prelude`


tyt2y3:
This changeset is too big and it kind of broke the API for no good reason. Is there another approach?

billy1624:
What you mean by "breaking the API"?

billy1624:
Hey @tyt2y3, this is ready for review!

examples/graphql_example/migration/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-pub use sea_schema::migration::*;
+pub use entity::sea_orm::migration::prelude::*;
```
tyt2y3, line 1: why is this `entity::sea_orm` instead of `sea_orm`?

Cargo.toml
```diff
@@ -57,6 +62,7 @@ pretty_assertions = { version = "^0.7" }
 debug-print = []
 default = [
     "macros",
+    "migration",
```
tyt2y3, line 65: I don't think we should add this as default feature

Cargo.toml
```diff
@@ -65,6 +71,7 @@ default = [
 ]
 macros = ["sea-orm-macros"]
 mock = []
+migration = ["sea-schema", "macros", "clap", "tracing-subscriber", "dotenv", "async-std"]
```
tyt2y3, line 74: It seems that what we need is one more crate sea-orm-migration

Cargo.toml
```diff
@@ -40,11 +40,16 @@ uuid = { version = "0.8", features = ["serde", "v4"], optional = true }
 ouroboros = "0.15"
 url = "^2.2"
 once_cell = "1.8"
+sea-schema = { git = "https://github.com/SeaQL/sea-schema", branch = "dump-sea-orm-dep", features = [ "migration" ], default-features = false, optional = true }
+async-std = { version = "^1", features = [ "attributes", "tokio1" ], optional = true }
+tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = true }
+dotenv = { version = "^0.15", optional = true }
+clap = { version = "^2.33", optional = true }
```
tyt2y3, line 47: These dependencies definitely do not belong here :(

tyt2y3 (changes requested):

examples/graphql_example/migration/src/lib.rs
tyt2y3: we should make it `sea_orm_cli::migration` or `sea_orm_migration`

tyt2y3: `entity` is a pure logic crate it does not make sense for `migration` be imported via `entity`

Cargo.toml
tyt2y3: actually we should remove this from the main `sea-orm` crate altogether

tyt2y3:
Architecturally, the dependency chain should be `sea-query` (AST) -> `sea-schema` (logic) -> `sea-orm-xxx` (API) -> `sea-orm-cli` (executable)

tyt2y3:
Let's take the time to make it right once and for all

tyt2y3:
> I would refer splitting it into `sea-orm-migration`, as I don't want `sea-orm-cli` depends on `sea-orm`. It'll makes things complicated.

Agree

tyt2y3 (approved):

tyt2y3:
Hell yeah

PR #667, opened 2022-04-09
Fix lints and warnings

fairingrey:
## PR Info

This PR doesn't change any logic but it does clean up a bunch of rustc and clippy lints in both source code and tests. I've suppressed warnings where I thought they were appropriate.


tyt2y3 (approved):
Thank you so much. Definitely appreciated!

PR #671, opened 2022-04-13
Support on conflict

SebastienGllmt:
## PR Info

This PR exposes the `ON CONFLICT` support that was added in [sea-query](https://github.com/SeaQL/sea-query/pull/279)

## Adds

Option for a new `insert_or` that allows you to specify an on-conflict behavior

## Public API Rationale

I think there were 3 ways this could be exposed from a public API point of view:
1. Modify the existing `insert` function to add a new parameter
2. Keep the `insert` function as-is, but introduce some state in the ActiveModel to handle conflicts
3. Create a new function called `insert_or`

- I didn't pick (1) because this feels like too big of a breaking change.
- I didn't pick (2) because to be, handling conflict is something you do in a per-query basis and not in the model. For example, in some cases you may want to do nothing on a conflict, some cases you may want to error, some cases you may want to override.
- (3) seemed like the best solution

## Handling of return value

One of the conflict options is `ON CONFLICT DO NOTHING`. This doesn't work with the current API where the `insert` function is assumed to always have a return value (do nothing returns an empty SQL result if a conflict is found)

There are two ways this can be solved:
1. When using `ON CONFLICT DO NOTHING`, check if nothing was returned and then do a `find` to find the entry that caused the conflict and return that as the result of the insert
2. Make that when using `insert_or`, we have an optional return instead of a mandatory return

I picked (2) seems (1) probably has performance implications

## TODO

- [ ] When upgrading to `sea-query` 0.24.0 (which introduces the on-conflict support), I had to add `window: None` to a few functions. Is this correct? (see https://github.com/SeaQL/sea-query/pull/271)
- [ ] When using `do nothing`, we can't call `after_save` since nothing was saved. Is this okay?
- [ ] I'm not sure what to do about testing since it seems the insert functions are not directly tested either. Should it be done as an integration test in `tests/relational_tests`?
- [ ] Add documentation for `insert_or`

### DIFF
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -54,14 +54,15 @@ where
     /// Execute an insert operation and return the inserted model (use `RETURNING` syntax if database supported)
     pub fn exec_with_returning<'a, C>(
         self,
+        enforce_return: bool,
         db: &'a C,
-    ) -> impl Future<Output = Result<<A::Entity as EntityTrait>::Model, DbErr>> + '_
+    ) -> impl Future<Output = Result<Option<<A::Entity as EntityTrait>::Model>, DbErr>> + '_
     where
         <A::Entity as EntityTrait>::Model: IntoActiveModel<A>,
         C: ConnectionTrait,
         A: 'a,
     {
-        Inserter::<A>::new(self.primary_key, self.query).exec_with_returning(db)
+        Inserter::<A>::new(self.primary_key, self.query).exec_with_returning(enforce_return, db)
     }
 }
 
@@ -91,14 +92,15 @@ where
     /// Execute an insert operation and return the inserted model (use `RETURNING` syntax if database supported)
     pub fn exec_with_returning<'a, C>(
         self,
+        enforce_return: bool,
         db: &'a C,
-    ) -> impl Future<Output = Result<<A::Entity as EntityTrait>::Model, DbErr>> + '_
+    ) -> impl Future<Output = Result<Option<<A::Entity as EntityTrait>::Model>, DbErr>> + '_
     where
         <A::Entity as EntityTrait>::Model: IntoActiveModel<A>,
         C: ConnectionTrait,
         A: 'a,
     {
-        exec_insert_with_returning::<A, _>(self.primary_key, self.query, db)
+        exec_insert_with_returning::<A, _>(self.primary_key, self.query, enforce_return, db)
     }
 }
 
@@ -139,8 +141,9 @@ where
 async fn exec_insert_with_returning<A, C>(
     primary_key: Option<ValueTuple>,
     mut insert_statement: InsertStatement,
+    enforce_return: bool,
     db: &C,
-) -> Result<<A::Entity as EntityTrait>::Model, DbErr>
+) -> Result<Option<<A::Entity as EntityTrait>::Model>, DbErr>
 where
     <A::Entity as EntityTrait>::Model: IntoActiveModel<A>,
     C: ConnectionTrait,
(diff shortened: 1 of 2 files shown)


src/executor/insert.rs
```diff
@@ -139,8 +141,9 @@ where
 async fn exec_insert_with_returning<A, C>(
     primary_key: Option<ValueTuple>,
     mut insert_statement: InsertStatement,
+    enforce_return: bool,
```
SebastienGllmt, line 144: Maybe instead of doing this, `insert_or` should always call `exec` instead of `exec_with_returning`?

It would force you to always have to do two queries if you care about getting the return value of your insert though

SebastienGllmt:
I rebased on top of #673 to fix the bug with the 0.24.0 upgrade

src/executor/insert.rs
billy1624: Executing two queries even when the db supports returning isn't ideal

SebastienGllmt: Maybe instead of adding this extra `enforce_return` parameter, we should instead return an iterator/array

The reason I think this might be better is that I think there are 4 cases we care about:
1) No return because either insert failed
2) No return because a DO NOTHING conflict was triggered
3) Single return (from an `insert` call
4) Multiple return (from an `insert_many` call

So instead of introducing a `enforce_return`, we could handle the cases as follows
1) Return a DbErr just like now
2) Return an empty iterator
3) Return an iterator with a single value (but this is maybe kind of ugly user-side)
4) Return an iterator with all values

I bring this up because after this PR is merged, I want to make a PR that allows `exec_insert_with_returning` for multiple return values so I don't want to make incompatible breaking changes (but I also don't want to over-complicate this PR which leads to it never getting merged)

Another option would be to make `insert`/`insert_or` and  `insert_many` return different implementations of `Insert` with different functions.

billy1624 (commented):
Hey @SebastienGllmt, I would like to take a step back and look at the overall API.

- `insert_or` should be aligned with `insert` method which return `Result<Model, DbErr>`
- from the `OnConflict` statement provided by the user, we will know which columns are the conflict target (the `target` field inside `OnConflict` struct). We can execute the insert statement. If it failed, then we can construct a select statement conditioning on the conflict target. Finally, query the select statement and return the existing row from the db

SebastienGllmt:
@billy1624 I'm concerned that the performance implications of doing this probably aren't desired for cases where `ON CONFLICT DO NOTHING` is used (the only case where nothing is returned)

# Rationale

`ON CONFLICT` can only be used on unique constraints. Although there are some tables with multiple unique constraints, most of the time your unique constraint is the primary key (possibly a composite key)

Example: A database with 3 tables:
- Persons table
- Jobs table
- Some table that maps the many-many relation of <PersonId, JobId>

If you want to insert a new `<PersonId, JobId>` pair, you can just insert with an `ON CONFLICT DO NOTHING` because if row already exists, you already know all the data in the row so you don't care about the return value. Having sea-orm do a 2nd query to get the row provides no additional information and double the number of database roundtrips required

billy1624:
> If you want to insert a new `<PersonId, JobId>` pair, you can just insert with an `ON CONFLICT DO NOTHING` because if row already exists, you already know all the data in the row so you don't care about the return value. Having sea-orm do a 2nd query to get the row provides no additional information and double the number of database roundtrips required

Make sense. I think the return type of `insert_or` is confusing, `Result<Option<Model>, ...>`, as you have said the return values doesn't matter, we could just return `Result<(), ...>`. However, one can't access the inserted `ActiveModel` anymore as it has been moved because the signature of `insert_or<'a, C>(self, ...)`.

```rs
async fn insert_or<'a, C>(self, db: &'a C, on_conflict: &OnConflict) -> Result<Option<<Self::Entity as EntityTrait>::Model>, DbErr>
```

billy1624:
Hey @SebastienGllmt, sorry for the delay.

The ON CONFLICT features has been added to SeaORM on another PR:
- https://github.com/SeaQL/sea-orm/pull/791

Thank you so much for the contributions. And do let us know if you have idea on improving it further :)

PR #673, opened 2022-04-14
Update sea-query dependency to 0.24.0

### LINKED ISSUE #672: Lower function missing
## Description

The `lower` function was added to sea-query "0.24.0", but sea-orm 0.7.1 depends on sea-query "0.23.0".

### Background

We're migrating a project from Diesel to sea-orm, and need the lower function to perform a case-insensitive compare. 
Updating the dependency will allow us to write something like:

```rust
let like = format!("%{}%", default.to_lowercase());
Expr::expr(Func::lower(Expr::col(column))).like(&like)
```

which can be used to produce something like:
```sql
SELECT LOWER("first_name") LIKE '%bob%' FROM "character"
```

## Versions

sea-orm v0.7.1

## Additional Information

I'll submit a pull request to update the dependency and link it with this issue.

mibes404:
## PR Info

- Closes #672 

## Adds

The use of the `lower` function to sea-orm 0.7; amongst other features that sea-query 0.24.0 brings.

## Changes

- `Cargo.toml`; update sea-query to version = "^0.24.0"
- Use `window: None` on `SelectExpr` constructs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ futures-util = { version = "^0.3" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.7.0", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.23.0", features = ["thread-safe"] }
+sea-query = { version = "^0.24.0", features = ["thread-safe"] }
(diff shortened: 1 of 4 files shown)


tyt2y3:
We seemed to have broken something with Postgres enum

billy1624:
Parameter binding part in SeaQuery should be at fault.

Related: SeaQL/sea-query#303

PR #677, opened 2022-04-19
Derive extra error traits

SebastienGllmt:
Is there any reason these traits weren't derived for DbErr? I think every error in the Rust std lib derives these 4 traits so I'm not sure why this error only derives two of them.

## Adds

- Derive extra traits to DbErr


billy1624 (approved):
Hey @SebastienGllmt, no reason not to derive these two. Thanks!!

PR #679, opened 2022-04-21
Generate module names in the singular

negezor:
First of all, I would like to thank you for this wonderful library. The examples used a module named `cake`, I was expecting when generating the module name entity in singular (one entity = one row). However, unfortunately for me, I got `cakes`, which confused me a bit. For that reason, this PR is here.
P.S. Perhaps you should think about using the model name instead of a table, to simplify, for example, the singularization 🤔 

## Adds

- Adds a new argument to sea-orm-cli `--singularize`, which generates singular module names.

## Breaking Changes

This would probably break the `sea-orm-codegen` calls from the code, but I'm not sure if it was used that way or not.

- Adds `singularize` property for `Entity`, `Relation`, `ConjunctRelation` in  `sea-orm-codegen`.
- Requires one more argument in `EntityTransformer::transform`


billy1624 (commented):
Hey @negezor, thanks for the contributions! Basically the generated models and relations are named according to the table name in snake case.

tyt2y3 (changes requested):
Thank you for the big PR. I like the idea actually and I think it might be reasonable to be the default behaviour indeed.
However, I don't like the implementation where `singularize` is everywhere. For better maintainability, we should have one method called `resolve_module_name` that actually handles ALL conversions. It also opens opportunity to further customize the behaviour.

negezor:
Closing in favor of https://github.com/SeaQL/sea-orm/pull/680. But I'm still not sure it's the best option.

Issue #661, opened 2022-04-04
Add flag to sea-orm-cli to generate code for time crate


cemoktra:
## Summary

Since the `time` crate is supported as alternative to `chrono` it should be possible to generate code for `time` crate

## Motivation

Currently you need to write the ORM code yourself if you want to use the `time` crate, which might be annoying

nahuakang:
I'm looking for a `good first issue` :) Just wondering how one is generating code for `chrono` right now?

billy1624:
Hey @nahuakang, welcome! You can take a look at
- https://github.com/SeaQL/sea-orm/blob/86e7e808b37179315a1cc5c6c852764830c04661/sea-orm-codegen/src/entity/column.rs#L47-L51

nahuakang:
@billy1624 Thanks! I'm reading the docs as well as trying to understand the workflow with `sea-orm` now. Are there any examples/tests that demonstrate how to use `sea-orm-cli` to generate code for entities involving `chrono`? I haven't seen a `chrono` flag in `sea-orm-cli` so I wonder if the solution here is to add a flag for `time` crate or something else (like handling `time` types)?

billy1624:
Hey @nahuakang, you can simply create a database table with timestamp / datetime columns. Then, follow the steps on https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli, to generate  entity files with `sea-orm-cli`.

We don't have a `chrono` flag for `sea-orm-cli`, since it's the default crate to represent datetime crate as of now. I think we can add an option, `--date-time-crate`, to `sea-orm-cli generate entity`. It will take values such as `chrono` / `time` with `chrono` being the default. This way it's backward compatible and users can opt-in to it.

nahuakang:
Thanks! I'll give this a try :) Please assign the issue to me.

PR #680, opened 2022-04-22
Adds support custom name resolver in codegen

negezor:
I added the `NameResolver` trait for better maintainability. I haven't touched columns and other parts that don't depend on module names yet, and I haven't deleted the getter code either.

## Adds

- Adds a new argument to sea-orm-cli `--singularize`, which generates singular module names.
- Create your own name transformations in `sea-orm-codegen`

## Breaking Changes

- Requires one more argument in `EntityTransformer::transform`


sea-orm-codegen/src/entity/name_resolver.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn resolve_entity_name(&self, name: &str) -> String {
+        singular(name.to_camel_case())
+    }
+
+    fn resolve_relation_name(&self, name: &str) -> String {
+        singular(name.to_camel_case())
+    }
+}
```
billy1624, line 57: Having a name resolver is a good idea :)
Can we supply the output of `NameResolver`'s default implementation to `singular` method?

negezor: @billy1624 yes of course

tyt2y3 (commented):
Actually, we don't need a trait here. Everything can be done by a struct `NameResolver` and it only needs to have a state `singularize: bool` which can only be configured in the constructor `NameResolver::new()`

This is better that the last PR indeed

sea-orm-cli/src/commands.rs
```diff
@@ -150,7 +151,9 @@ pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dy
                 _ => unimplemented!("{} is not supported", url.scheme()),
             };
 
-            let output = EntityTransformer::transform(table_stmts)?
+            let name_resolver = NameResolver::new(singularize);
+
+            let output = EntityTransformer::transform(table_stmts, name_resolver)?
                 .generate(expanded_format, WithSerde::from_str(with_serde).unwrap());
```
billy1624, line 157: Hey @negezor, thanks for the updates! I think the semantic isn't seems right. The transformer, `EntityTransformer::transform(table_stmts, name_resolver)`, don't need name resolver. Instead, we could have introduce a new struct call `EntityWriterContext` which contains three things...

```rs
pub struct EntityWriterContext {
  pub(crate) expanded_format: bool,
  pub(crate) with_serde: WithSerde,
  pub(crate) name_resolver: NameResolver,
}
```

Then, `EntityWriter::generate` method would take an `EntityWriterContext`. Thoughts?

sea-orm-codegen/src/entity/base_entity.rs
```diff
@@ -125,6 +159,27 @@ impl Entity {
...
+            .collect()
+    }
+
+    pub fn resolve_conjunct_relations_to_relation_name(&self, name_resolver: &NameResolver) -> Vec<Ident> {
+        self.conjunct_relations
+            .iter()
+            .map(|con_rel| con_rel.resolve_to_relation_name(name_resolver))
+            .collect()
+    }
+
```
billy1624, line 182: All these `resolve_*` methods (here, and in other files except those in `name_resolver.rs`) seems just a duplicate of the original methods few line below, e.g. `resolve_conjunct_relations_via_module_name` method is just `get_conjunct_relations_via_snake_case` method with name resolve applied. We should change the logic in the original method, i.e. applying name resolver in it.

sea-orm-cli/src/commands.rs
tyt2y3: I agree indeed

nahuakang: @billy1624 In the case where we discussed about saving a context info on which datetime crate the user chooses in [the `--date-time-crate` flag for `sea-orm-cli`](https://github.com/SeaQL/sea-orm/issues/661), we'd basically write:
```rust
pub struct EntityWriterContext {
  pub(crate) expanded_format: bool,
  pub(crate) with_serde: WithSerde,
  pub(crate) name_resolver: NameResolver,
  pub(crate) date_time_crate: &str,
}
```
and pass this context all the way down to `Column.get_rs_type()` for resolving the Rust types based on `chrono` vs `time`? :)

billy1624:
Hey @negezor, any updates? And do you need any help? :)

billy1624:
> Also, since in both of our PRs we're using `EntityWriterContext`, maybe we could merge this PR and [PR#724](https://github.com/SeaQL/sea-orm/pull/724) sequentially? I'm happy to help with the context update, too.

Good suggestions! We could consolidate all flags into `EntityWriterContext`.

negezor:
Sorry for the long answer. Now this PR depends on #724.

sea-orm-codegen/src/entity/base_entity.rs
negezor: I didn't touch them originally, as that would have required more changes. Also, they are now more than just getters. Perhaps this one could be changed by later refactoring. 🤔

billy1624 (commented):
Hey @negezor, sorry for the delay. Since https://github.com/SeaQL/sea-orm/pull/724 has been merged. Would you like to rebase?

negezor:
I think I did the rebase wrong somehow 😅

sea-orm-codegen/src/entity/name_resolver.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub fn resolve_relation_name(&self, name: &str) -> String {
+        let name = name.to_camel_case();
+
+        if self.singularize {
+            singular(name)
+        } else {
+            name
+        }
+    }
```
billy1624, line 44: Hey @negezor, thanks for the hard rebase (it's hard as this PR is far behind from master).

I wonder why we need three different methods? I think we can merge it and have two methods? One for snake_case and the other for camel_case.

negezor: A fair remark, but I proceeded from the clear goals of the method. I can make changes if necessary.

PR #683, opened 2022-04-25
add is_changed to ActiveModelTrait

### LINKED ISSUE #674: is_changed on active_model
## Summary

As of now, we can check if field of active model is changed, but we can't check if anything in whole active model has changed.
It would be great to just check if pear.is_changed() to make updated

## Motivation

Now when i have active model and code like that:
```
async fn update_fruit(id: i64, color: Option<String>) -> Result<(), Error> {
    let apple: Option<fruit::Model> = Fruit::find_by_id(id).one(db).await?;
    let mut apple: Fruit::ActiveModel = apple.unwrap().into();

    if color.is_some() {
        apple.color = Set(color.unwrap());
    }

    let apple: fruit::Model = apple.update(db).await?;

    Ok(())
}
```
this code will produce error, cos it try to make Update query like that:
`update fruit Set Where`

With is_changed() method we could make:
```
if apple.is_changed() {
    apple.update(db).await?;
}
```
so, no errors will be produced

kirawi:
I will work on writing tests, but I wanted to submit a draft PR.

## PR Info

- Closes #674

## Adds

`is_changed` will check if any column in the model was changed.

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -548,6 +548,11 @@ pub trait ActiveModelTrait: Clone + Debug {
 
         Ok(am)
     }
+
+    /// Returns `true` if any columns were changed.
+    fn is_changed(&self) -> bool {
+        <Self::Entity as EntityTrait>::Column::iter().any(|col| !self.get(col).is_unchanged())
+    }
 }
 
 /// A Trait for overriding the ActiveModel behavior
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -263,7 +263,7 @@
 //! ## Who's using SeaORM?
 //!
 //! SeaORM is the foundation of [StarfishQL](https://github.com/SeaQL/starfish-ql), an experimental graph database and query engine.
-//! 
+//!
 //! For more projects, see [Built with SeaORM](https://github.com/SeaQL/sea-orm/blob/master/COMMUNITY.md#built-with-seaorm).
 //!
 //! ## License


src/lib.rs
```diff
@@ -263,7 +263,7 @@
 //! ## Who's using SeaORM?
 //!
 //! SeaORM is the foundation of [StarfishQL](https://github.com/SeaQL/starfish-ql), an experimental graph database and query engine.
-//! 
+//!
```
kirawi, line 266: This was changed after running `cargo fmt`

src/entity/active_model.rs
```diff
@@ -548,6 +548,11 @@ pub trait ActiveModelTrait: Clone + Debug {
 
         Ok(am)
     }
+
+    /// Returns `true` if any columns were changed.
+    fn is_changed(&self) -> bool {
+        <Self::Entity as EntityTrait>::Column::iter().any(|col| !self.get(col).is_unchanged())
```
billy1624, line 554: Hey @kirawi, thanks for contributing! Shouldn't this be...?

```suggestion
        <Self::Entity as EntityTrait>::Column::iter().any(|col| self.get(col).is_unchanged())
```

billy1624 (changes requested):

src/lib.rs
billy1624: Nice!

src/entity/active_model.rs
kudlatyamroth: i think it is good with !, as on column we can only check if it is not changed and on active model we will check if it is changed

kirawi: If symmetry is desired, I could change it to be `is_unchanged`

kudlatyamroth: for me `is_changed` is way better then `is_unchanged`

billy1624: Oppps... sorry, please ignore this. I thought it's `is_changed` but in fact it's `is_unchanged`

tyt2y3: 😄

kirawi:
I only added one test, is that fine?

tyt2y3 (approved):
Awesome

PR #696, opened 2022-05-04
(de)serialize custom JSON types

### LINKED ISSUE #676: Unable to (de)serialize nested JSON Blobs using serde Traits
## Description

We have a database schema where we save (from the db pov) untyped json blobs into the database + some metadata columns. We have a serde compatible struct for describing the blobs on the rust side and expected to be able to use the `#[sea_orm(db_type = "Json")]` as annotation to be able to deserialize the JSON col into the rust struct on select.

#### Rust

```rust
use some_external_crate::priding::our::type::SomeStruct;

#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "table_name")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: String,
    #[sea_orm(db_type = "Json")]
    pub details: SomeStruct,
    pub refreshed_at: DateTime,
}

pub enum Relation {}

...
```

#### SQL

```sql
CREATE TABLE table_name (
    id varchar(64) PRIMARY KEY,
    refreshed_at   timestamptz NOT NULL,
    details        jsonb NOT NULL
);
```

### Expected Behavior

The details field of the Model should be deserialized from `json` / `jsonb` columns into the `SomeStruct` type in rust.

### Actual Behavior

```
"the trait bound `sea_orm::Value: From<SomeStruct>` is not satisfied"
```

### Reproduces How Often

Every time.

## Versions

0.7.1

## Additional Information

None

billy1624:
## PR Info

- Closes #676

- Dependencies:
  - SeaQL/sea-query#316

## Adds

- [x] Implement `TryGetable` for `sea_query::JsonValue`
- [x] Re-export `sea_query::JsonValue` in prelude

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,7 @@ futures-util = { version = "^0.3" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.7.0", path = "sea-orm-macros", optional = true }
-# sea-query = { version = "^0.24.4", features = ["thread-safe"] }
-sea-query = { path = "../sea-query", features = ["thread-safe"] }
+sea-query = { version = "^0.24.4", git = "https://github.com/SeaQL/sea-query", branch = "185-json-wrapper-struct", features = ["thread-safe"] }
 sea-strum = { version = "^0.23", features = ["derive", "sea-orm"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,8 @@ futures-util = { version = "^0.3" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.7.0", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.24.4", features = ["thread-safe"] }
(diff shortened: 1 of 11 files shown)


schulke-214:
Hey! Thank you so much for digging into the details of this issue :)

One thing i noticed is still the missing from trait implementations mentioned in my issue:

https://github.com/SeaQL/sea-orm/issues/676#issuecomment-1106562045

The absence of those prevent this code from compiling:

```rust
...

#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "contract_cache")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: String,
    #[sea_orm(db_type = "Json")]
    pub details: JsonStruct,
    pub refreshed_at: DateTime<Utc>,
}

...
```

I would strongly advice to keep the sea orm type `JsonValue<T>` out of the models since it makes working with them outside of sea orm a lot nastier than needed.

What do you think about this?

Have a nice day :)

tyt2y3:
Perhaps a PR would better explain what you meant?

tyt2y3:
Okay so while I think that introducing a new Value type is unnecessary, introducing a new Trait `TryGetableFromJson` is necessary for SeaORM to provide blanket implementation.

```rust
pub trait TryGetableFromJson: Deserialize { ... }

impl<T: TryGetableFromJson> TryGetable for T {}
```

So the user shall only:

```rust
impl TryGetableFromJson for MyJson {}
``` 

And such we don't have to use a wrapper type.

PR #698, opened 2022-05-04
Adding test for actix-web

dsullivan7:
I'd like to add a test for actix-web, however, I found that the `into_transaction_log` function can't be called on the `MockDatabase` because `AppState`, more specifically `conn`, has already been consumed. I'm looking for guidance on how to avoid this problem, let me know what I can do.

### DIFF
--- a/examples/actix4_example/entity/Cargo.toml
+++ b/examples/actix4_example/entity/Cargo.toml
@@ -19,6 +19,7 @@ features = [
   "debug-print",
   "runtime-actix-native-tls",
   "sqlx-mysql",
+  "mock",
   # "sqlx-postgres",
   # "sqlx-sqlite",
 ]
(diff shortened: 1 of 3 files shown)


dsullivan7:
@tyt2y3 No problem, good to know for the future.

billy1624 (commented):
Hey @dsullivan7, sorry for the delay and thanks for the contributions!!

Recently, we're updating the examples to include mock testing for SeaORM's CRUD operations. Please check the PR at https://github.com/SeaQL/sea-orm/pull/890.

Basically, we decouple SeaORM CRUD operations from the web API. Then, we mock test the CRUD operations. This avoid the pitfall that most web API requires the app state to be cloneable but SeaORM connection isn't cloneable when `mock` feature is enabled.

I think what you did is interesting and it's a good demonstration on testing the point of integration between Actix, SeaORM and Tera. And it fill the missing gap that we didn't cover in https://github.com/SeaQL/sea-orm/pull/890 - testing the handlers of web API.

dsullivan7:
@billy1624 Thanks very much for the response. I'm going to close this PR. Glad to see more development on this, it's a great library.

billy1624:
Thanks again for contributing!! Let me know if you spot any room for improvements :)

PR #723, opened 2022-05-13
[examples] entity & application crate depends on SeaORM with minimum required features enabled

billy1624:
## Changes

- [x] entity crate in each example should depends on SeaORM with only `macros` and `with-*` features enabled, i.e. without any runtime or backend features enabled. As it's up to application crate (root crate) to make the decision.
- [x] application crate in each example would depends on SeaORM with the target runtime and backend features enabled

The philosophy behind it is to allow the sharing of entity crate for other purposes not only for the usage in backend API but also frontend or even for unit testing in a standalone crate. In addition, this makes compiling entity crate alone much faster as it doesn't depends on any runtime library nor db backend.

### DIFF
--- a/examples/actix3_example/Cargo.toml
+++ b/examples/actix3_example/Cargo.toml
@@ -22,3 +22,14 @@ serde = "1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 entity = { path = "entity" }
 migration = { path = "migration" }
+
+[dependencies.sea-orm]
+path = "../../" # remove this line in your own project
+version = "^0.8.0"
+features = [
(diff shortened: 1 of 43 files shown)


nahuakang:
If this new crate dependency for workspace structure should be recommended to users, maybe we could also update the [entity crate](https://github.com/SeaQL/seaql.github.io/blob/master/SeaORM/docs/05-migration/01-setting-up-migration.md#entity-crate) doc?

tyt2y3:
Yeah sure thank you for the suggestion

tyt2y3 (approved):
Exactly how it was designed

Issue #722, opened 2022-05-13
[CLI] Show progress when generating entity files


billy1624:
## Motivation

The `sea-orm-cli generate entity` command doesn't show its progress such as which entity (db table) it's generating. Some time when it fail, user have no idea which db table is at fault and making it hard to debug.

## Proposed Solutions

- Display the progress when running the command and show info such as the table name and the specification of table columns that it's currently generating

## Additional Information

- Discussed and suggested on #702

kyoto7250:
@billy1624 

Hello, I am interested in this issue.
I am new comer to contributing to this repository,  Could you tell me about it?

billy1624:
Hey @kyoto7250, thanks for the interest!

I think there are two things we need to code.

First, log the table name and its column info of the generating entity.

- We could add the `tracing::info!` macro to https://github.com/SeaQL/sea-orm/blob/e00562695ebb99eca444acbce41aafd0c5f3d539/sea-orm-codegen/src/entity/writer.rs#L93-L111
- For example, in `sea-orm-migration` we have done something similar logging https://github.com/SeaQL/sea-orm/blob/e00562695ebb99eca444acbce41aafd0c5f3d539/sea-orm-migration/src/migrator.rs#L154-L157

Second, print log of log level `info` via `tracing_subscriber`.

- We could initialize the `tracing_subscriber` at the `else` part of https://github.com/SeaQL/sea-orm/blob/e00562695ebb99eca444acbce41aafd0c5f3d539/sea-orm-cli/src/commands.rs#L19-L24
- For example, https://github.com/SeaQL/sea-orm/blob/e00562695ebb99eca444acbce41aafd0c5f3d539/sea-orm-migration/src/cli.rs#L45-L54

PR #735, opened 2022-05-16
feat(sea-orm-cli): output log about  generated file name.

kyoto7250:
## PR Info

- Closes 
close : #722 

## Adds
- [x] Add logging during code generation for sea-orm-cli.

## Breaking Changes
- [x] No Breaking Changes

## example output
```
❯ ../sea-orm/sea-orm-cli/target/debug/sea-orm-cli generate entity --database-url mysql://root@localhost/world --output-dir out/
Generating city.rs
    > Column `ID`: i32, auto_increment, not_null
    > Column `Name`: String, not_null
    > Column `CountryCode`: String, not_null
    > Column `District`: String, not_null
    > Column `Population`: i32, not_null
    > Column `created_at`: DateTimeUtc, not_null
Generating country.rs
    > Column `Code`: String, not_null
    > Column `Name`: String, not_null
    > Column `Continent`: Continent, not_null
    > Column `Region`: String, not_null
    > Column `SurfaceArea`: Decimal, not_null
    > Column `IndepYear`: Option<i16>
    > Column `Population`: i32, not_null
    > Column `LifeExpectancy`: Option<Decimal>
    > Column `GNP`: Option<Decimal>
    > Column `GNPOld`: Option<Decimal>
    > Column `LocalName`: String, not_null
    > Column `GovernmentForm`: String, not_null
    > Column `HeadOfState`: Option<String>
    > Column `Capital`: Option<i32>
```

---

It is my first PR in this repo.
Thank you in advance :)

### DIFF
--- a/sea-orm-cli/src/commands.rs
+++ b/sea-orm-cli/src/commands.rs
@@ -29,10 +29,10 @@ pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dy
                     .with_level(false)
                     .without_time();
 
-                tracing_subscriber::registry()
+                let _ = tracing_subscriber::registry()
                     .with(filter_layer)
                     .with(fmt_layer)
-                    .init()
+                    .try_init();
             }
 
             let max_connections = args
--- a/sea-orm-cli/src/commands.rs
+++ b/sea-orm-cli/src/commands.rs
@@ -4,6 +4,7 @@ use regex::Regex;
 use sea_orm_codegen::{EntityTransformer, OutputFile, WithSerde};
 use std::{error::Error, fmt::Display, fs, io::Write, path::Path, process::Command, str::FromStr};
 use url::Url;
+use tracing_subscriber::{prelude::*, EnvFilter};
 
 pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dyn Error>> {
     match matches.subcommand() {
@@ -21,6 +22,17 @@ pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dy
                     .with_max_level(tracing::Level::DEBUG)
                     .with_test_writer()
                     .try_init();
+            } else {
+                let filter_layer = EnvFilter::try_new("sea_orm_codegen=info").unwrap();
+                let fmt_layer = tracing_subscriber::fmt::layer()
+                    .with_target(false)
+                    .with_level(false)
+                    .without_time();
+
+                tracing_subscriber::registry()
+                    .with(filter_layer)
+                    .with(fmt_layer)
+                    .init()
             }
 
             let max_connections = args
--- a/sea-orm-cli/src/commands.rs
+++ b/sea-orm-cli/src/commands.rs
@@ -3,8 +3,8 @@ use clap::ArgMatches;
 use regex::Regex;
 use sea_orm_codegen::{EntityTransformer, OutputFile, WithSerde};
 use std::{error::Error, fmt::Display, fs, io::Write, path::Path, process::Command, str::FromStr};
-use url::Url;
 use tracing_subscriber::{prelude::*, EnvFilter};
+use url::Url;
 
 pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dyn Error>> {
     match matches.subcommand() {
--- a/sea-orm-codegen/src/entity/column.rs
+++ b/sea-orm-codegen/src/entity/column.rs
@@ -1,9 +1,9 @@
-use sea_query::Write;
 use crate::util::escape_rust_keyword;
 use heck::{CamelCase, SnakeCase};
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
 use sea_query::{ColumnDef, ColumnSpec, ColumnType};
+use std::fmt::Write as FmtWrite;
 
 #[derive(Clone, Debug)]
 pub struct Column {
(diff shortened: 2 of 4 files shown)


sea-orm-codegen/src/entity/writer.rs
```diff
@@ -94,6 +95,12 @@ impl EntityWriter {
         self.entities
             .iter()
             .map(|entity| {
+                let entity_file = format!("{}.rs", entity.get_table_name_snake_case());
+                info!(
+                    "Generating {}",
+                    entity_file
+                );
```
billy1624, line 102: Is it possible to print the info of all table columns, one line per column.
https://github.com/SeaQL/sea-orm/blob/e00562695ebb99eca444acbce41aafd0c5f3d539/sea-orm-codegen/src/entity/base_entity.rs#L9

billy1624 (commented):
Hey @kyoto7250, thanks for the contributions! Looking good!

sea-orm-codegen/src/entity/writer.rs
kyoto7250: I fixed in [this commit](https://github.com/SeaQL/sea-orm/pull/735/commits/b893a3c81356a0bfdc9678cc2ffb9e50a4c0290e).

If there is a cooler way to display, I will respond.

kyoto7250:
@billy1624 
Could I request another review?

kyoto7250:
I missed CI failed notification.
I will check it.

sea-orm-cli/src/commands.rs
```diff
@@ -21,6 +22,17 @@ pub async fn run_generate_command(matches: &ArgMatches<'_>) -> Result<(), Box<dy
...
+                let filter_layer = EnvFilter::try_new("sea_orm_codegen=info").unwrap();
+                let fmt_layer = tracing_subscriber::fmt::layer()
+                    .with_target(false)
+                    .with_level(false)
+                    .without_time();
+
+                tracing_subscriber::registry()
+                    .with(filter_layer)
+                    .with(fmt_layer)
+                    .init()
```
billy1624, line 35: Can you try below? It will fix the errors thrown on tests

```suggestion
                let _ = tracing_subscriber::registry()
                    .with(filter_layer)
                    .with(fmt_layer)
                    .try_init();
```

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -94,6 +95,14 @@ impl EntityWriter {
...
             .iter()
             .map(|entity| {
+                let entity_file = format!("{}.rs", entity.get_table_name_snake_case());
+                let column_names = entity
+                    .columns
+                    .iter()
+                    .map(|column| column.name.clone())
+                    .collect::<Vec<String>>();
+                info!("Generating {} {:?}", entity_file, column_names);
+
```
billy1624, line 105: I think we can show a bit more details of each column, e.g. https://github.com/kyoto7250/sea-orm/compare/issue_702...billy1624:pr/735
What's your thoughts?

billy1624: This will print something like...

```
Generating metadata.rs
	> Column `uuid`: Binary(Some(16)), not_null
	> Column `type`: String(Some(255)), not_null
	> Column `key`: String(Some(255)), not_null
	> Column `value`: String(Some(255)), not_null
	> Column `bytes`: Custom(BLOB), not_null
	> Column `date`: Date
	> Column `time`: Time(None)
Generating applog.rs
	> Column `id`: Integer(None), auto_increment, not_null
	> Column `action`: String(Some(255)), not_null
	> Column `json`: Json, not_null
	> Column `created_at`: Timestamp(None), not_null
```

sea-orm-cli/src/commands.rs
kyoto7250: It seems to good.

Thank you!

sea-orm-codegen/src/entity/column.rs
```diff
@@ -143,6 +143,30 @@ impl Column {
...
+            vec.push("auto_increment")
+        }
+        if self.not_null {
+            vec.push("not_null")
+        }
+        if self.unique {
+            vec.push("unique")
+        }
+        return vec.join(", ");
+    }
```
billy1624, line 169: Just a nitpick, I think we and refactor the code and use `write!` macro instead of `join` on an `Vec`?
- https://doc.rust-lang.org/std/macro.write.html#examples

kyoto7250: I refactored in [this commit](https://github.com/SeaQL/sea-orm/pull/735/commits/b57e17269a1a5416cc13ba04691ac1911a91a555)

billy1624 (commented):
Hey @kyoto7250, sorry for being so picky. We want to keep the code as clean as possible. I still think there is room for refactoring, please review the commit below and feel free to cherry pick it
- https://github.com/kyoto7250/sea-orm/compare/issue_702...billy1624:pr/735

kyoto7250:
@billy1624 
Of course. I'll take care of it.
Thank you for giving me a better way to write!

https://github.com/SeaQL/sea-orm/pull/735#pullrequestreview-976309024

billy1624 (approved):
Thanks! @kyoto7250

Irykson:
Hey there, I’m still quite new to rust in general so more of question here.
When would you use std::fmt::Display trait instead? Or in other terms, why not to use it here?

billy1624:
Hey @Irykson, welcome and good questions! I think generally speaking `std::fmt::Display` is for printing debug info and shows all the attribute inside a data structure. But here, we're printing column info with in custom format, hence, I don't think it's ideal to implement `std::fmt::Display` for this purpose.

kyoto7250:
I aggre @billy1624 .

Irykson:
Thanks for the reply and your warm welcome @billy1624!
I really don't want to hijack this PR to start a discussion or a Q&A, so feel free to shut it here :) 
I just found this [guideline](https://github.com/rust-lang/rfcs/blob/master/text/0565-show-string-guidelines.md) which says, `Display` is intended for user-facing output which was also my understanding of this trait. So I actually thought, the reasoning behind not using it here, has more to do with the fact, that `get_info` could be thought of a somewhat niche case and you want another default output for `Display` in the future.

tyt2y3:
@Irykson yeah usually what would happen is when we impl Display get_info will be called inside the function

tyt2y3:
Thank you all for making a better developer experience for all

PR #737, opened 2022-05-16
removing async_trait from non async trait in connection.rs

sjud:
removed 1 async trait macro over trait that is not async


billy1624 (approved):
Thanks!! @sjud Nice observation!

tyt2y3:
Thank you

PR #754, opened 2022-05-20
Cursor Pagination

### LINKED ISSUE #300: Efficient Pagination
It would be fantastic if SeaORM could support pagination in an efficient manner, namely by using a combination of `first`/`after` or `last`/`before` count/cursor combinations.

I see that the current implementation uses `OFFSET` and `LIMIT` to do pagination, which will work just fine for small tables, but quickly becomes an issue as the size of the table grows. Imagine that an `OFFSET` grows very large, then this becomes necessarily a lot of work for the database as the entire ordered result set containing `OFFSET` + `LIMIT` rows needs to be computed then `OFFSET` rows skipped over.

This topic has been written about [many](http://mysql.rjweb.org/doc.php/pagination), [many](https://mariadb.com/kb/en/pagination-optimization/) [times](https://ivopereira.net/efficient-pagination-dont-use-offset-limit).

A more efficient way of tackling this problem would be to use a cursor based pagination where the database can leverage its indexes to produce the ordered result set starting from the cursor. e.g. `SELECT id, name FROM table WHERE table.id >= $after LIMIT $first;`.

This feels like it might be a necessary approach for users wanting to use this ORM at anything resembling scale and would greatly assist implementation for systems that de facto use cursor based pagination (i.e., [graphQL](https://relay.dev/graphql/connections.htm)).

It's unlikely I'll be able to contribute a PR to implement this unfortunately, but figured I'd raise the issue for anyone else.

billy1624:
## PR Info

- Closes #300
- Dependencies
  - https://github.com/SeaQL/sea-query/commit/f92a08fb254bda4b923595785f71e0046cbc3d58
  - https://github.com/SeaQL/sea-schema/commit/4e28a6fb4f00ee7be2e42e13bf767844d51b0974

## Adds

- [x] Add `Cursor` & `CursorTrait`, implementing cursor pagination for SeaORM


Sytten:
Nice API! My biggest critic is that people will want to paginate on models that dont have a natural order via the PK. And supporting that will require a breaking change.

src/executor/cursor.rs
```diff
@@ -0,0 +1,264 @@
...
+        self.limit = Some(CursorLimit::Last(num_rows));
+        self
+    }
+
+    /// Fetch the rows
+    pub async fn all<C>(&self, db: &C) -> Result<Vec<S::Item>, DbErr>
+    where
+        C: ConnectionTrait,
+    {
+        let mut query = self.query.clone();
```
Sytten, line 78: That clone is somewhat annoying since most of the time self is not reused.

src/executor/cursor.rs
```diff
@@ -0,0 +1,264 @@
...
+    PrimaryKeyToColumn, PrimaryKeyTrait, Select, SelectModel, SelectorTrait,
+};
+use sea_query::{Condition, IntoValueTuple, Order, SelectStatement, SimpleExpr, Value};
+use std::marker::PhantomData;
+
+/// Limit number of rows to be fetched from the cursor
+#[derive(Debug)]
+pub enum CursorLimit {
+    /// Fetch first N rows in ascending order of primary key
+    First(u32),
```
ikrivosheev, line 12: Maybe we can expand to `u64`?

tyt2y3:
I am a little confused that why there isn't any raw SQL unit tests to illustrate the behaviour

src/executor/cursor.rs
```diff
@@ -0,0 +1,264 @@
...
+        if let Some(limit) = &self.limit {
+            let (order, limit) = match limit {
+                CursorLimit::First(limit) => (Order::Asc, limit),
+                CursorLimit::Last(limit) => (Order::Desc, limit),
+            };
+            for key in E::PrimaryKey::iter() {
+                query.order_by(key, order.clone());
+            }
+            query.limit(*limit as u64);
+        }
```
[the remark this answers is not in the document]
billy1624: > To ensure proper ordering of the paginated result, the select statement must have order by expression.

I have updated the logic and ordering are assumed to be set in the select statement, even the limit is not set.

src/executor/cursor.rs
billy1624: Managed to avoid the clone :)

src/executor/cursor.rs
billy1624: Sure!

billy1624:
> Nice API! My biggest critic is that people will want to paginate on models that dont have a natural order via the PK. And supporting that will require a breaking change.

I have updated the API to... which user have full control over the target to filter

```rs
Entity::find()
    .order_by_asc(Column::Id)
    .cursor()
    .after(Column::Id, 10)
    .last(2)
    .all(&db)
    .await?,
```

billy1624:
Hey @tyt2y3, unit tests were being updated and now it asserts on plain SQL string

negezor:
What about order by non-unique keys? 🤔

Sytten:
You always need a column that is totally orderable for cursor pagination, even if you order by a partially orderable one.

src/executor/cursor.rs
```diff
@@ -0,0 +1,315 @@
...
+            },
+        ];
+
+        let db = MockDatabase::new(DbBackend::Postgres)
+            .append_query_results(vec![models.clone()])
+            .into_connection();
+
+        assert_eq!(
+            Entity::find()
+                .cursor()
```
tyt2y3, line 163: I feel like the order by column should be part of the constructor

src/executor/cursor.rs
```diff
@@ -0,0 +1,315 @@
...
+/// Cursor pagination
+///
+/// To ensure proper ordering of the paginated result, the select statement must have order by expression.
+#[derive(Debug, Clone)]
+pub struct Cursor<S>
+where
+    S: SelectorTrait,
+{
+    pub(crate) query: SelectStatement,
+    pub(crate) last: bool,
```
tyt2y3, line 17: I don't know why this state is necessary

src/executor/cursor.rs
```diff
@@ -0,0 +1,315 @@
...
+            self.reverse_ordering();
+        }
+        self.last = false;
+        self
+    }
+
+    /// Limit result set to only last N rows in ascending order of the paginated query
+    pub fn last(&mut self, num_rows: u64) -> &mut Self {
+        self.query.limit(num_rows);
+        if !self.last {
```
tyt2y3, line 73: why are we doing this?

src/executor/cursor.rs
```diff
@@ -0,0 +1,315 @@
...
+                    id: 26,
+                    name: "Blueberry".into(),
+                    cake_id: Some(1),
+                },
+            ]])
+            .into_connection();
+
+        assert_eq!(
+            Entity::find()
+                .order_by_asc(Column::Id)
```
tyt2y3, line 274: Is it an assumption that the pre-existing order by clause has to be in ASC ?

src/executor/cursor.rs
```diff
@@ -0,0 +1,315 @@
...
+
+impl<S> Cursor<S>
+where
+    S: SelectorTrait,
+{
+    /// Initialize a cursor
+    pub fn new(query: SelectStatement) -> Self {
+        Self {
+            query,
+            last: false,
```
tyt2y3, line 29: um... it seems to be

tyt2y3:
Thank you so far. Great work!

I think there are 2 things I think we need to address:

Instead of "inheriting" the order by clause, we should override it by the cursor column the user passes in via the `cursor()` method, and so we don't assume it is ordered already.

And thus, the Cursor itself should hold which column to cursor on. And so, the API can be simplified into:

```rust
Entity::find()
    .cursor(Column::Id)
    .after(10)
    .last(2)
    .all(&db)
    .await?
```

tyt2y3 (changes requested):
As stated

Sytten:
@tyt2y3 like I commented above, you need to support two column if you want to support partially orderable columns. If not then you can use one column.

tyt2y3 (approved):

tyt2y3:
oh I didn't merge the sea-query side yet

tyt2y3:
I reverted this merge. We need to upgrade sea-query to 0.25 first.
As stated before, `git = XXX` cannot be merged into master to avoid maintenance confusion.

PR #755, opened 2022-05-20
Allow old pending migration to be applied

### LINKED ISSUE #751: sea-orm-migrations relaxed file ordering
## Motivation

Currently the file ordering is used to get if a migration has been applied:
https://github.com/SeaQL/sea-orm/blob/3f90c094079501211e88ea2111ce8f2473493bdf/sea-orm-migration/src/migrator.rs#L70

This is fine in basic cases but it becomes a pain when you work with a larger team where concurrent PR can introduce new migrations. Since those migrations might not be related, there is often no need to force the migrations to be at the top of the stack.

## Proposed Solutions

Now there is somewhat of a debate on this topic in our community. One camp is the flywaydb (https://flywaydb.org/documentation/concepts/migrations#naming) where historically you could only do versioned migrations.
In this case the naming should force a version in the name (meaning the trait `MigrationName` should probably change).
Generally this migration type also has more guarantees (similar to sqlx migrations) with a hash.

The other camp (which I generally refer) is to approach it in a more relaxed manner and is adopted by a lot of migrations tools (diesel is an example for rust). In this camp, the tool will just apply the migrations in the order in which they appear, no matter if a "more recent" migration was already applied.

billy1624:
## PR Info

- Closes #751

## Breaking Changes

- [x] SeaORM migrator will search and apply all pending migration, without enforcing pending migrations must be in front of migrations that has been applied. See below for the in depth explanation

For example...

In the database, there are 2 migrations already been applied:
- m20220120_000001_create_post_table
- m20220520_000001_create_order_table

On the other hand, we have 3 migration files:
- m20220120_000001_create_post_table.rs [Applied]
- m20220519_000001_alter_post_table.rs [Pending]
- m20220520_000001_create_order_table.rs [Applied]

So, we have a old (pending) migration, `m20220519_000001_alter_post_table`, waiting to be applied. And I have relaxed the constraint, now migrator will only search for pending migrations instead of enforcing all pending migrations to be in front of migrations that has been applied.

### DIFF
--- a/sea-orm-migration/src/migrator.rs
+++ b/sea-orm-migration/src/migrator.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::fmt::Display;
 use std::time::SystemTime;
 use tracing::info;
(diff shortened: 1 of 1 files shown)


Sytten (approved):
Logic looks good!

tyt2y3:
Thank you all

PR #758, opened 2022-05-23
Migration without depending on async-std

### LINKED ISSUE #748: sea-orm-migration should not depend on async-std
## Description

The migrations should not depend on any async implementation.
What should instead be done is changing the template created on init to have the dependency and then remove it from the crate.

billy1624:
## PR Info

- Closes #748

## Breaking Changes

- [ ] Drop `async-std` dependency from `sea-orm-migration`
- [ ] `async-std` isn't re-exported from `sea_orm_migration` & `sea_orm_migration::prelude`

### DIFF
--- a/examples/actix3_example/migration/Cargo.toml
+++ b/examples/actix3_example/migration/Cargo.toml
@@ -10,7 +10,13 @@ path = "src/lib.rs"
 
 [dependencies]
 entity = { path = "../entity" }
+async-std = { version = "^1", features = ["attributes", "tokio1"] }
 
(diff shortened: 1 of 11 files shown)


ikrivosheev (approved):
@billy1624 thank you! LGTM!

Sytten (approved):

tyt2y3:
Thank you everyone. Btw should this be released in 0.8.x?

billy1624:
I think it should be shipped in 0.9.0. To avoid breaking changes in minor version

PR #767, opened 2022-05-26
`LinkDef`  with custom join condition

### LINKED ISSUE #160: `LinkDef` with custom join condition
Right now, `LinkDef` is simply an alias of `RelationDef`.
We can design a richer API for `LinkDef` that impl `From<RelationDef>`:

```rust
pub struct CakeToFilling;

impl Linked for CakeToFilling {
    type FromEntity = Entity;

    type ToEntity = super::filling::Entity;

    fn link(&self) -> Vec<LinkDef> {
        vec![
            super::cake_filling::Relation::Cake.def().rev().into()
                .and_on_condition(Expr::cust("a = b")),
            super::cake_filling::Relation::Filling.def().into(),
        ]
    }
}
```

This will open more flexibilities. Unfortunately it is a breaking change, thus has to wait until `0.3.0`

billy1624:
## PR Info

- Closes #160

## Adds

- [ ] `LinkDef` struct

## Breaking Changes

- [ ] `LinkDef` no longer is an alias of `RelationDef`

## Problems

- We need to consider the case where the RHS (right hand side) table in join expression is being renamed during runtime when user construct the query with `find_linked` method.

```sql
SELECT
    `filling`.`id`,
    `filling`.`name`,
    `filling`.`vendor_id`
FROM
    `filling`
    INNER JOIN `cake_filling` AS `r0` ON `r0`.`filling_id` = `filling`.`id`
    INNER JOIN `cake` AS `r1` ON `cake`.`id` > 2 /* This shoule be "`r1`.`id` > 2" */
    AND `r1`.`id` = `r0`.`cake_id`
WHERE
    `r1`.`id` = 12
```

## Solution

- `and_on_condition` and `and_on_condition` methods could take a lambda function which has table `DynIden` of LHS and RHS provided.

i.e.

```rs
pub fn and_on_condition<C>(mut self, func: F) -> Self
where
    F: Fn(DynIden, DynIden) -> SimpleExpr,
{
    ...
}
```

### DIFF
--- a/src/entity/link.rs
+++ b/src/entity/link.rs
@@ -1,10 +1,81 @@
 use crate::{
-    join_tbl_on_condition, unpack_table_ref, EntityTrait, QuerySelect, RelationDef, Select,
+    join_tbl_on_condition, unpack_table_ref, EntityTrait, Identity, QuerySelect, RelationDef,
+    Select,
 };
-use sea_query::{Alias, IntoIden, JoinType, SeaRc};
+use sea_query::{Alias, Condition, ConditionExpression, IntoIden, JoinType, SeaRc, TableRef};
 
-/// Same as [RelationDef]
-pub type LinkDef = RelationDef;
+/// Defines a link between tables
+#[derive(Debug)]
+pub struct LinkDef {
+    /// Reference from another Entity
+    pub from_tbl: TableRef,
+    /// Reference to another ENtity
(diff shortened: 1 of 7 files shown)


billy1624:
@tyt2y3 thoughts?

billy1624:
Closing this because of redesign

PR #768, opened 2022-05-26
Create a `num_items_and_pages` on the `paginator`

fistons:
This method allows us to get both number of items and pages of a
paginator with only one database query.

## Adds

- [ ] Adds a new method on the paginator allowing us to get bot number of pages and items with one request

### DIFF
--- a/src/executor/paginator.rs
+++ b/src/executor/paginator.rs
@@ -80,8 +80,20 @@ where
     /// Get the total number of pages
     pub async fn num_pages(&self) -> Result<usize, DbErr> {
         let num_items = self.num_items().await?;
-        let num_pages = (num_items / self.page_size) + (num_items % self.page_size > 0) as usize;
+        let num_pages = self.compute_pages_number(num_items);
         Ok(num_pages)
+    }
+
+    /// Get the total number of items and pages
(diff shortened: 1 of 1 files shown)


src/executor/paginator.rs
```diff
@@ -80,10 +80,22 @@ where
     /// Get the total number of pages
     pub async fn num_pages(&self) -> Result<usize, DbErr> {
         let num_items = self.num_items().await?;
-        let num_pages = (num_items / self.page_size) + (num_items % self.page_size > 0) as usize;
+        let num_pages = self.compute_pages_number(num_items);
         Ok(num_pages)
     }
 
+    /// Get the total number of items and pages in a tuple as (number_of_items, number_of_pages)
+    pub async fn num_items_and_pages(&self) -> Result<(usize, usize), DbErr> {
```
billy1624, line 88: Hey @fistons, thanks for the contributions! I'm just thinking would it be better to return a struct with named fields: `number_of_items` and `number_of_pages`?

fistons: Yes, sure things!

tyt2y3:
Thank you all

PR #769, opened 2022-05-26
Parse enum from repr[X] enums with values

dragonnn:
## PR Info

- Dependencies:
  - 

- Dependents:
  - 

## Adds

- [x] this PR allows the usage of enums with repr[X] and integer values for every variant if no `string_value` or `num_value`. This can be useful for auto-generated enums from some third party library, this is exactly my use case as I have enums generated from a protobuf file by tonic-build create, it only allows to add on generated enums attributes for the whole enum, so I add EnumIter and DeriveActiveEnum to it. DeriveActiveEnum still needs  `string_value` or `num_value` even that the generated enum looks like this:
```rust
#[derive(serde::Deserialize, serde::Serialize)]
#[derive(strum_macros::EnumString, strum_macros::IntoStaticStr, strum_macros::Display, sqlx::Type, sea_orm::entity::prelude::EnumIter, sea_orm::entity::prelude::DeriveActiveEnum)]
#[sqlx(type_name = "text")]
#[sea_orm(rs_type = "i32", db_type = "Integer")]
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
#[repr(i32)]
pub enum Kind {
    PreInsulation = 0,
    Hub = 1,
    ImpulsMeter = 2,
    GenericDevice = 3,
    HeatMeter = 4,
    AccessControl = 5,
    SolarInverter = 6,
    VibroAccelerometers = 7,
    VibroDetector = 8,
}
```
With that PR as a fallback DeriveActiveEnum parses all the enum integers directly from them self.

## Changes

- [x] ActiveEnum macro was extended to parse enum values if no `string_value` or `num_value` is present

### DIFF
--- a/sea-orm-macros/src/derives/active_enum.rs
+++ b/sea-orm-macros/src/derives/active_enum.rs
@@ -1,7 +1,7 @@
 use heck::CamelCase;
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned};
-use syn::{punctuated::Punctuated, token::Comma, Lit, LitInt, LitStr, Meta};
+use syn::{punctuated::Punctuated, token::Comma, Expr, Lit, LitInt, LitStr, Meta};
 
 enum Error {
     InputNotEnum,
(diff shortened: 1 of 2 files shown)


billy1624 (commented):
Hey @dragonnn, thanks for proposing this on Discord and making this as PR :)

Would you mind adding some test cases for it? Asserting `to_value` & `try_from_value` on ActiveEnum defined in repr[x]
https://github.com/SeaQL/sea-orm/blob/183639dc8c817fbd3211a87c67bcb983db4050ed/src/entity/active_enum.rs#L233-L271

dragonnn:
That was a good idea because I found out that negative values don't parse 😅. Add a workaround for that use case in [f5d9b2f](https://github.com/SeaQL/sea-orm/pull/769/commits/f5d9b2ff34fdbf48f39a1da64a58c7efd56c3daf)
and add tests for all the fallback cases and all types :)

billy1624 (approved):
Thanks again for the great contributions! @dragonnn Glad we catch the edge case of negative value :D

tyt2y3:
This is brilliant! And I love the tests too. We'd want to include this in our docs too!

tyt2y3 (approved):
well done

PR #778, opened 2022-05-31
Stream metrics

nappa85:
This PR improves metrics for streams.
Actual code meters only query execution in streams, not data transfer/encoding, therefore things seems to be really fast.
E.g.: the same query, using `Selector::all()` is metered at 0.005s, using `Selector::stream()` is metered at 0.000003s. Things are fast but not that fast.

My PR measures also every `Stream::next()` duration and sums all together to give a more realistic value, taking the metric on Stream drop.

To be honest, there is an async part of the stream that can't be measured, e.g. when `Stream::next()` is called and data is already ready, but that's one of the advantages of Streams.

More, this PR removes calls to `SystemTime::now()` when not needed, removing the overhead of a system call for who doesn't uses metrics.

### DIFF
--- a/src/database/stream/metric.rs
+++ b/src/database/stream/metric.rs
@@ -1,0 +1,56 @@
+use std::{time::Duration, pin::Pin, task::Poll};
+
+use futures::Stream;
+
+use crate::{QueryResult, DbErr, Statement};
+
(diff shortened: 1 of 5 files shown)


tyt2y3:
Interesting... so it's 5ms vs 3us? oh

Issue #576, opened 2022-03-04
Be able to support Array in database


hf29h8sh321:
## Summary

Be able to directly support Array in database

Such as `text[]`

## Additional Information

Right now it is represented by a custom `column_type = "Custom(\"array\".to_owned())"` and the type in struct is `String`, not `Vec<T>`

billy1624:
Hey @hf29h8sh321, I assumed you mean supporting Array of some permitives for PostgreSQL?

hf29h8sh321:
@billy1624 For example, `Vec<String>`

billy1624:
Okay, this has been discussed before
- #151

EtienneProthon:
Could we have an example of the workaround using Json type for the moment ? I need to store and array of enum and i can't find any example on array storage as json.

billy1624:
Hey @EtienneProthon, please check #783

negezor:
This is the only thing that prevents me from using sea-orm right now, because I need to work with the postgres array. I can't change the array to JSON in my tables. Is there a workaround to use an array without JSON? I'm interested in an int8 array.

kudlatyamroth:
and array support was merged in sea-query: https://github.com/SeaQL/sea-query/pull/467 :)

ikrivosheev:
@kudlatyamroth and released)

kudlatyamroth:
@ikrivosheev in release 0.27 info there is no mention about it :)
so i assume that it was not included in 0.27, but if it is, its great news :)

ikrivosheev:
@kudlatyamroth thank! I will edit Changelog!

emmiegit:
Exciting! What is needed for sea-orm to support it now that sea-query has it?

ikrivosheev:
@emmiegit you can look here: https://github.com/SeaQL/sea-orm/pull/985

PR #783, opened 2022-06-02
[Demo] vector of string save as serialized format in db

billy1624:
## PR Info

- Demo for https://github.com/SeaQL/sea-orm/issues/576#issuecomment-1144475251

## Adds

- [x] Add demo for `Vec<T>` save as JSON string in the database


tyt2y3:
Thank you. It looks like quite some boilerplate. May be we can throw a macro for this somewhen

PR #785, opened 2022-06-02
Migration shouldn't depends on symbol provided by entity crate

billy1624:
## PR Info

- Dependents:
  - https://github.com/SeaQL/sea-orm-tutorial/pull/9

## Breaking Changes

- [x] Drop dependency on `entity` crate inside all example `migration` crate: making all migration independent from `entity` crate

## Changes

- [x] `sea-orm-cli`: add `async-std` dependency; default migration file with sample

CC @shpun817

### DIFF
--- a/examples/actix3_example/migration/Cargo.toml
+++ b/examples/actix3_example/migration/Cargo.toml
@@ -9,7 +9,6 @@ name = "migration"
 path = "src/lib.rs"
 
 [dependencies]
-entity = { path = "../entity" }
 async-std = { version = "^1", features = ["attributes", "tokio1"] }
 
 [dependencies.sea-orm-migration]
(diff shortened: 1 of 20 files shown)


RipleyTom:
I think the migration part could do with some link to sea-query documentation for Iden trait ( https://docs.rs/sea-query/latest/sea_query/#iden ) somewhere in there to give more context to the syntax.

It wasn't instantly clear to me that Post was the name of the table and Table was a special value giving the enum name.

billy1624:
Hey @RipleyTom, nice suggestions! I have added extra docs for the use of `Iden`.

tyt2y3:
Actually Post::Table seems like a SeaQuery relic now that is slight inconsistent with SeaORM's convention
But yeah good to have them separated

PR #790, opened 2022-06-09
Soft Delete - 2

### LINKED ISSUE #220: [Feature Request] Soft Delete

billy1624:
## PR Info

- Closes #220

### DIFF
--- a/sea-orm-macros/src/derives/column.rs
+++ b/sea-orm-macros/src/derives/column.rs
@@ -23,10 +23,12 @@ pub fn impl_default_as_str(ident: &Ident, data: &Data) -> syn::Result<TokenStrea
         })
         .collect();
 
+    let mut soft_delete_column = quote! { None };
     let name: Vec<TokenStream> = variants
         .iter()
         .map(|v| {
-            let mut column_name = v.ident.to_string().to_snake_case();
+            let variant_ident = &v.ident;
+            let mut column_name = variant_ident.to_string().to_snake_case();
             for attr in v.attrs.iter() {
                 if let Some(ident) = attr.path.get_ident() {
                     if ident != "sea_orm" {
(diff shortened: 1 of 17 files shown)


tyt2y3:
This will probably not land in 0.9

billy1624:
Too much for 0.9?

rakshith-ravi:
Any chance this will make it through to 0.12?

tyt2y3:
There is some chance, we plan to rethink/rework this PR.

But before that we'd have to clear the backlog for 0.12 first.

May be it will land in 0.12.1?

amorfc:
Hello, how things going on about soft-delete feature ?

wvffle:
Hi, 0.12.15 landed a few weeks ago, is there any active work done on this PR?

boringplay:
Hello, how things going on about soft-delete feature ?

PR #792, opened 2022-06-13
✨ feat(sqlite3): support set sqlcipher key

cache-missing:
## Adds

- [ ] support sqlcipher 

## Breaking Changes

no breaking changes


tyt2y3:
Awesome thanks!

PR #800, opened 2022-06-17
`sqlx_logging_level`, a wrapper around `sqlx::ConnectOptions::log_statements`

SandaruKasa:
## PR Info

- Closes: None

## Adds

- An ability to change level at which sqlx logs statements. 
- New methods:
```rust
impl ConnectOptions {
    ...
    
    pub fn sqlx_logging_level(&mut self, level: log::LevelFilter) -> &mut Self { ... }

    pub fn get_sqlx_logging_level(&self) -> log::LevelFilter { ... }
}
```

## Changes
- `ConnectOptions`' size has changed, I don't think it's a breaking change
- `log` is now a direct dependency
- The fact that `log::LevelFilter::Off` exsits makes it possible to deprecate 
```rust
    pub fn sqlx_logging(&mut self, value: bool) -> &mut Self { ... }
```
in future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ time = { version = "^0.2", optional = true }
 futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
 tracing = { version = "0.1", features = ["log"] }
-log = "^0.4"
+log = { version = "^0"}
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.8.0", path = "sea-orm-macros", optional = true }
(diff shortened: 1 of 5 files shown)


tyt2y3 (approved):
Only nitpick might be to use the = { version = } syntax

SandaruKasa:
> Only nitpick might be to use the = { version = } syntax

Done 31bc9e937c592dcda54ea132c44f00e49769f1af

tyt2y3:
Actually, I consider 800 to be a good number :)

PR #807, opened 2022-06-21
Select one-to-one mandatory relationship

billy1624:
## Adds

- [x] `SelectTwo`: select models linked by mandatory relationship. The resulting models will be `(M, N)` instead of `(M, Option<N>)`

### DIFF
--- a/src/database/mock.rs
+++ b/src/database/mock.rs
@@ -707,4 +707,50 @@ mod tests {
 
         Ok(())
     }
+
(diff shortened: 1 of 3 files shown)


tyt2y3:
Um... interesting idea. Although I would say it's just a one liner to unwrap the second element of the tuple.
And actually there is no way to guarantee it won't panic, unless there is a not null constraint.

Issue #809, opened 2022-06-22
Table name not match in some examples


swaitw:
## Description

try to run [axum-example](https://github.com/SeaQL/sea-orm/tree/master/examples/axum_example) which not work for me

fixed by change(https://github.com/SeaQL/sea-orm/blob/master/examples/axum_example/migration/src/m20220120_000001_create_post_table.rs):
`
#[derive(Iden)]
enum Post {
    Table,
    Id,
    Title,
    Text,
}
`
to 
`
#[derive(Iden)]
enum Posts {
    Table,
    Id,
    Title,
    Text,
}
`

or modify (https://github.com/SeaQL/sea-orm/blob/master/examples/axum_example/entity/src/post.rs) from
`
#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Serialize, Deserialize)]
#[sea_orm(table_name = "posts")]
pub struct Model {
    #[sea_orm(primary_key)]
    #[serde(skip_deserializing)]
    pub id: i32,
    pub title: String,
    #[sea_orm(column_type = "Text")]
    pub text: String,
}

`

to 

`
#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Serialize, Deserialize)]
#[sea_orm(table_name = "post")]
pub struct Model {
    #[sea_orm(primary_key)]
    #[serde(skip_deserializing)]
    pub id: i32,
    pub title: String,
    #[sea_orm(column_type = "Text")]
    pub text: String,
}

`

billy1624:
Hey @swaitw, thanks for catching that. I will fix it in a bit!

PR #812, opened 2022-06-22
[examples] Fix table identifier

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/809

## Fixes

- [x] Example table name `posts` is in plural; Update identifier enum to match the table name in plural form


tyt2y3:
Cool thanks

Issue #486, opened 2022-01-23
Allow to fetch entity and all related ones


Sytten:
Consider the following example (where a release has multiple release_links):
```rust
let result = releases::Entity::find()
                .find_with_related(release_links::Entity)
                .order_by_desc(releases::Column::ReleasedAt)
                .all(conn)
                .await?;
```
This returns a `Vec<(Model, Vec<Model>)>`. Changing `.all` to .`one` yields a `Option<(Model, Option<Model>)>`.
For a One-to-Many relation this is really weird, what is expected is `Option<(Model, Vec<Model>)>`.

billy1624:
Hey @Sytten, basically, calling one returns the first pair of related model i.e. `Option<(Model, Option<Model>)>`. And, calling all returns all of the pairs, `Vec<(Model, Vec<Model>)>`.

For a one-to-many relation you can expect the behaviour stated above happened.
But for a one-to-one relation you can expect both one and all will return zero or one pair of related model.

Sytten:
I guess then there should be a `first` which would return a `Option<(Model, Vec<Model>)>`? Basically the same query as all but with a `LIMIT 1` on the parent for performance reasons.

isgj:
That will complicate the query builder as there is no easy way to limit on one of the join tables. It can be achieved with a subquery like:

```sql
select releases.id, other_col
from releases
join release_links on ....
where releases.id in (
    select releases.id 
    from releases 
    where ... 
    limit 1
)
``` 

and if the conditions involve both the tables the join should happen also in the subquery.

I guess you can do:

```rs
let result = releases::Entity::find()
                .filter(
                              Condition::any().add(
                                  releases::Column::Id.in_subquery(
                                      Query::select()
                                          .column(releases::Column::Id)
                                          .from(releases::Entity) // if required you can add other conditions here in the subquery
                                          .limit(1)
                                          .to_owned()
                                  )
                              )
                  )
                .find_with_related(release_links::Entity)
                .order_by_desc(releases::Column::ReleasedAt)
                .all(conn)
                .await?;
```
The result will be of type `Vec<(Model, Vec<Model>)>` and the `len` at most `1`

tyt2y3:
This is a very interesting sugguestion. First of all, if you only need `one` and `all` it's related models, then `find_related` on the model itself is the recommended way.

[...]

PR #813, opened 2022-06-22
Drop `SelectTwoMany::one()` method

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/486

## Breaking Changes

- [x] Drop `SelectTwoMany::one()` method: `SelectTwoMany` is for selecting models of a one-to-many relationship but `SelectTwoMany::one()` returns `Option<(E, Option<F>)>` and the return value is a pair of models instead of `(E, Vec<F>)` which is weird select result for a one-to-many relationship

### DIFF
--- a/src/executor/select.rs
+++ b/src/executor/select.rs
@@ -361,14 +361,6 @@ where
         }
     }
 
-    /// Select one Model
-    pub async fn one<'a, C>(self, db: &C) -> Result<Option<(E::Model, Option<F::Model>)>, DbErr>
-    where
-        C: ConnectionTrait,
-    {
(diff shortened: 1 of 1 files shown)


tyt2y3:
I guess we would like to add some API docs somewhere explaining the absence of such function.
But yeah it's a trap

tyt2y3 (changes requested):
Documentation

tyt2y3:
Actually, it might be "easy" to make two queries, first `one()`, and then `find_related` ?

billy1624:
Docs updated :)

PR #816, opened 2022-06-24
Setup CI for TiDB

billy1624:
## PR Info

Try to setup CI for TiDB


billy1624:
CC @itwaix

itwaix:
The error here is because TiDB does not support nested transactions

PR #822, opened 2022-06-26
Cursor pagination

tyt2y3:
Continue #754 

Things to do:

I released the `clear_order_by` in sea-query 0.25

So the dependency is to upgrade sea-orm to sea-query 0.25.1


tyt2y3:
Is there a PR for upgrading sea-query to 0.25?

billy1624:
> Is there a PR for upgrading sea-query to 0.25?

Hey @tyt2y3, check https://github.com/SeaQL/sea-orm/pull/834

penso:
Is there any plan to have cursor in other directions? Like if we want to have `created_at DESC` to list most recent first?

billy1624:
Hey @penso, thanks for asking!!

This is exactly what `.last(N)` method is for.

```rs
use sea_orm::{entity::*, query::*, tests_cfg::cake};

// Create a cursor that order by `cake`.`id`
let mut cursor = cake::Entity::find().cursor_by(cake::Column::Id);

// Filter paginated result by `cake`.`id` > 1 AND `cake`.`id` < 100
cursor.after(1).before(100);

// Get last 10 rows (order by `cake`.`id` DESC but rows are returned in ascending order)
for cake in cursor.last(10).all(db).await? {
    // Do something on cake: cake::Model
}
```

Docs: https://www.sea-ql.org/SeaORM/docs/basic-crud/select/#cursor-pagination

penso:
Wouldn't it be easier to have:

```
let mut cursor = cake::Entity::find().cursor_by(cake::Column::Id, .desc)
for cake in cursor.first(10).all(db).await? {
}
```

so we don't need to convert/reverse the resulted array if we actually want to deliver the last ID first and in descending order?

billy1624:
Hey @penso, I think we have add two additional method - `first_desc` and `last_desc` - to the cursor:

```rs
// Get first 10 rows (order by `cake`.`id` ASC but rows are returned in descending order)
for cake in cursor.first_desc(10).all(db).await? {
    // Do something on cake: cake::Model
}

// Get last 10 rows (order by `cake`.`id` ASC but rows are returned in descending order)
for cake in cursor.last_desc(10).all(db).await? {
    // Do something on cake: cake::Model
}
```

tyt2y3:
Is there any drawback to just `rev` the `Vec`?

billy1624:
> Is there any drawback to just `rev` the `Vec`?

Good question! @penso?
Note that `first_desc` and `last_desc` are performing `Vec:::rev` under-the-hood loll

penso:
No drawback than I can think of, except the mental thinking. I've done a "manual" cursor this week-end and will try to move to `cursor_by` and see if it's easier or not.

billy1624:
Please! And let us know your experience after using it :)

Issue #808, opened 2022-06-21
Upgrade SQLx dependency to 0.6.x


billy1624:
## Motivation

Upgrading SQLx dependency as well as the downstream dependencies such as `time`, `uuid`, `bigdecimal`, etc.

## Tasks

- [x] Upgrade to SQLx 0.6.x
- [x] Upgrade to time 0.3.x
    - https://github.com/SeaQL/sea-orm/issues/788
- [x] Upgrade to uuid 1.x
    - https://github.com/SeaQL/sea-orm/issues/689

## Additional Information

Potential Breaking Changes
* https://github.com/launchbadge/sqlx/issues/1618  
  SQLite `chrono::DateTime<FixedOffset>` timezone fix
    * `DateTime<FixedOffset>` will be stored in SQLite with the correct timezone instead of always in UTC.
      This was flagged as a "potentially breaking change" since it changes how dates are sent to SQLite.
* https://github.com/launchbadge/sqlx/issues/1901
  Pool fixes and breaking changes
    * Renamed `PoolOptions::connect_timeout` to `acquire_timeout` for clarity.

tyt2y3:
Is there a PR for this yet?

billy1624:
No active PR at the moment :P

Sytten:
Yeah agreed, needs an update to time. I am wondering if chrono support should be just removed, its basically deprecated.
It's a bit annoying that when using `sqlx-postgres`, chrono gets added automatically and also the CLI defaults to chrono without a flag for time.

tyt2y3:
Regarding unwanted dependencies, I think we can take advantage of `?` dependency and solve that

tyt2y3:
sea-query and sea-schema is already upgraded and released to sqlx 0.6
@billy1624 sea-orm side can assign to you?

tyt2y3:
> I am wondering if chrono support should be just removed, its basically deprecated

I also don't oppose to change the default in SeaORM if a lib is deprecated. However is there a source for this information?

Sytten:
Well its not dead but no release in two years and unpatched vulnerability. The current master is very different from the current release. You can keep it with the ? Thats true.

ikrivosheev:
> sea-query and sea-schema is already upgraded to sqlx 0.6 and released
> @billy1624 sea-orm side can assign to you?
> Or you'd need help from Ivan?

Did you mean me?

tyt2y3:
[...]

PR #834, opened 2022-07-03
Bump version of sqlx, sea-query, sea-schema, time and uuid

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/808

## Breaking Changes

- [x] Upgrade `time` to ^0.3
- [x] Upgrade `sqlx` to ^0.6
- [x] Upgrade `uuid` to ^1
- [x] Upgrade `sea-query` to ^0.26
- [x] Upgrade `sea-schema` to ^0.9.2
- [x] https://github.com/launchbadge/sqlx/issues/1618  
  SQLite `chrono::DateTime<FixedOffset>` timezone fix
    * `DateTime<FixedOffset>` will be stored in SQLite with the correct timezone instead of always in UTC.
      This was flagged as a "potentially breaking change" since it changes how dates are sent to SQLite.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ log = { version = "^0" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.8.0", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.26.0", features = ["thread-safe"] }
+sea-query = { version = "^0.26", features = ["thread-safe"] }
 sea-strum = { version = "^0.23", features = ["derive", "sea-orm"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,19 +25,19 @@ path = "src/lib.rs"
 async-stream = { version = "^0.3" }
 async-trait = { version = "^0.1" }
 chrono = { version = "^0", optional = true }
-time = { version = "^0.2", optional = true }
+time = { version = "^0.3", optional = true }
 futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
 log = { version = "^0" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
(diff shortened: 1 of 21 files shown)


Cargo.toml
```diff
@@ -25,19 +25,19 @@ path = "src/lib.rs"
...
-time = { version = "^0.2", optional = true }
+time = { version = "^0.3", optional = true }
 futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
 log = { version = "^0" }
 tracing = { version = "0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.8.0", path = "sea-orm-macros", optional = true }
-sea-query = { version = "^0.24.5", features = ["thread-safe"] }
+sea-query = { version = "^0.26.0", features = ["thread-safe"] }
```
ikrivosheev, line 35: Maybe without pin minor version?

Cargo.toml
```diff
@@ -53,6 +53,7 @@ rust_decimal_macros = { version = "^1" }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 sea-orm = { path = ".", features = ["mock", "debug-print"] }
 pretty_assertions = { version = "^0.7" }
+time = { version = "^0.3", features = ["macros"] }
```
ikrivosheev, line 56: Maybe merge with privies time?

ikrivosheev (commented):
@billy1624 thank you, small comments.

Cargo.toml
billy1624: I don't want `macros` feature to be enabled by default. I just need it for `sea-orm` tests

ikrivosheev: Ah, sorry, these are different files...

Cargo.toml
billy1624: Do you mean the patch version?
- `<major>.<minor>.<patch>`

Yeah, we can relax that

ikrivosheev (approved):
LGTM!

Cargo.toml
billy1624: NVM :D

tyt2y3 (approved):
Awesome! Thank you all

Issue #838, opened 2022-07-04
Too much warnings when  compile code. 


baoyachi:
When compile sea-orm.It's very uncomfortable for a programmer who likes not to warnings.

Suggestion: add clippy check in github action, when check code warnings compile not pass.
```
cargo clippy --all-targets --all-features -- -D warnings
```

billy1624:
Hey @baoyachi, thanks for the initiatives! I agree we should have enforce clippy checks in CI. Please check https://github.com/SeaQL/sea-orm/pull/840

baoyachi:
So good. I have some other suggestions：
* add `cargo fmt`
* add `cargo fix` && `cargo fix` 

use `cargo fix` twice，make ci check code git status is clean, `cargo fix` can check git is dirty ,compile error

detail see: [github action](https://github.com/baoyachi/shadow-rs/blob/master/.github/workflows/check.yml#L31-L36)

```yml
      - name: Check format
        run: cargo fmt --all -- --check
      - name: Check fix
        run: cargo fix && cargo fix
      - name: Check with clippy
        run: cargo clippy --all-targets --all-features -- -D warnings
```

billy1624:
We would run `cargo fmt` before release and we didn't enforce it in CI as we don't want to put too much burden on the contributors. So, I would say adding clippy checks for now is sufficient :)

baoyachi:
Oh, I see.

billy1624:
Anyways, thanks for the help and support!

PR #840, opened 2022-07-05
[CI] run clippy checks & fix clippy warnings

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/838

## Adds

- [x] Clippy CI checks


baoyachi:
LGTM.

tyt2y3:
Thank you for the hard work. I think the problem stems from the fact that we only compile a minimal subset by default.

Issue #841, opened 2022-07-06
Add sea-orm-cli flag to skip deserialization of primary keys for entities


Witcher01:
## Motivation

Expecting a JSON value with a web framework usually makes it return the correct status codes etc. if an entity cannot be deserialized from the request, but currently this is hard to do as the entities by default required the primary key to be present in the JSON, which is not what a `POST` request tends to accept.

## Proposed Solutions

Add a flag to `sea-orm-cli` that enables `#[serde(skip_deserializing)]` for primary keys of entities.

PR #846, opened 2022-07-06
Add CLI option to `sea-orm-cli` to skip primary keys with serde on generation of entities

Witcher01:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/841

## Adds

- [ ] Command line argument `--skip-primary-key-deserialization` to `sea-orm-cli` for the option to skip deserialization of primary keys with serde by adding `#[serde(skip_deserialization)]` to the primary key

As the comment says, this adds the option to skip serializing primary keys with serde in entities generated by `sea-orm-cli`.
The command line flag name is a bit long, so feel free to suggest a different name for it.
The implementation could use some work to feel better, so I'll take suggestions for this as well.

As of now `clap` doesn't treat the new argument as only being allowed to be set when `--with-serde` with a value other than `none` is used.


Witcher01:
Right, I didn't look at the tests. This might take a bit of time

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -594,8 +615,15 @@ impl EntityWriter {
...
-                        #[sea_orm(#ts)]
+                    if skip_pk_deserialization {
+                        quote! {
+                            #[sea_orm(#ts)]
+                            #[serde(skip_deserialization)]
+                        }
+                    } else {
+                        quote! {
+                            #[sea_orm(#ts)]
+                        }
```
billy1624, line 626: Hey @Witcher01, sorry for the delay! Thanks for the great contributions!! Looks good overall :)

I have a minor nitpick. Please check if the refactoring make any sense
- https://github.com/Witcher01/sea-orm/compare/skip-serde-deserialization...billy1624:pr/846

Witcher01: No worries! I'm not sure what you mean, my commit looks good to me, tests run, and running it gives me the right results, too.
Could you elaborate? Thanks.

Witcher01:
I just noticed that skipping the primary key deserialization isn't added in the expanded formats. I'll fix that and push a new commit when I'm done. ETA this evening, hopefully

billy1624:
Thanks! @Witcher01

sea-orm-codegen/src/entity/writer.rs
billy1624: I don't want to keep a mutable `skip_pk_deserialization` and that's why refactored the code :P

PR #847, opened 2022-07-06
change gen migrate init resp data dynamic

baoyachi:
## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## Adds

- [ ] 

## Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]


sea-orm-cli/src/commands.rs
```diff
@@ -219,25 +221,37 @@ pub fn run_migrate_command(
...
+                    handlebars.register_template_string("content", &content).unwrap();
+                    let content = handlebars.render("content", &json!({"datatime": datatime})).unwrap();
+
                     let content = $fn_content(content.to_string());
                     file.write_all(content.as_bytes())?;
                 };
             }
             write_file!("src/lib.rs");
-            write_file!("src/m20220101_000001_create_table.rs");
+            write_file!("src/{{datatime}}_create_table.rs");
```
billy1624, line 254: I think we can rename `m20220101_000001_create_table.rs` to something like `migration_template.rs`. Then, the generated could be named according to the current timestamp e.g. `mYYYYMMDD_000001_create_table.rs`

Issue #828, opened 2022-06-29
Add column name when `error occurred while decoding: Null`


dinhani:
## Motivation
A common error I make is forgetting to map a nullable column using `Option`, so I receive the error message `error occurred while decoding: Null`.

I would like the have the column name in the error message so I can easily identify what column was mapped wrongly.

## Proposed Solutions
It seems the column name is available in all places the Null error is generated, so it would be nice if the column name is passed to it as parameter, so it can be included in error message.

## Additional Information

Example of code where `col` parameter exists and could be used to identify the column that failed with the Null message:
https://github.com/SeaQL/sea-orm/blob/0c96e133250b7c8ab43a6a99931b82c95f59acc1/src/executor/query.rs#L346-L354

tyt2y3:
Thank you for your suggestion. I agree with that. PR is welcomed.

billy1624:
Hey @dinhani, welcome! Sorry for the delay! Yeah, it's definitely doable and making it much easier to debug.

billy1624:
For anyone who is reading this, I'll mark this issue as open for contributions :)

kyoto7250:
@billy1624  

hello, Could you please assign me with this issue?

ikrivosheev:
@kyoto7250 done)

kyoto7250:
@ikrivosheev 

Thank you.

I will try it.

billy1624:
Thanks! @kyoto7250 let me know if you need any help along the way :)
Just open a draft PR

PR #853, opened 2022-07-09
include column name in TryGetError::Null

kyoto7250:
## PR Info
close https://github.com/SeaQL/sea-orm/issues/828 

## Adds
This PR adds column info in TryGetError::Null

Thank you in advance.

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -33,14 +33,16 @@ pub enum TryGetError {
     /// A database error was encountered as defined in [crate::DbErr]
     DbErr(DbErr),
     /// A null value was encountered
-    Null,
+    Null(String),
 }
 
 impl From<TryGetError> for DbErr {
     fn from(e: TryGetError) -> DbErr {
         match e {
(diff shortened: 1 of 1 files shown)


src/executor/query.rs
```diff
@@ -198,7 +200,7 @@ macro_rules! try_getable_mysql {
                     #[allow(unused_variables)]
                     QueryResultRow::Mock(row) => row.try_get(_column.as_str()).map_err(|e| {
                         debug_print!("{:#?}", e.to_string());
-                        TryGetError::Null
+                        TryGetError::Null(_column)
```
tyt2y3, line 203: Since we are using the variable now, we should remove the `_` prefix

kyoto7250: I fixed in https://github.com/SeaQL/sea-orm/pull/853/commits/3ee0662d123abf7af626903b19b1d833f3fb9403

tyt2y3:
Awesome

tyt2y3 (approved):

PR #854, opened 2022-07-09
auto set timestamp column when update & insert

### LINKED ISSUE #827: Automatically update `created_at` & `updated_at` timestamp columns on update
## Motivation

This is a common feature for an ORM. Which "touch" these timestamp columns to record the last update timestamp and create at timestamp of a row.

## Proposed Solutions

User can enable this feature by placing attribute on the corresponding field in the entity model.

```rs
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub name: String,
    #[sea_orm(updated_at)]
    pub updated_at: Option<DateTimeWithTimeZone>,
    #[sea_orm(created_at)]
    pub created_at: Option<DateTimeWithTimeZone>,
}
```

For insert, the value of both `updated_at` and `created_at` will be set to `CURRENT_TIMESTAMP`.

For update, the value of `updated_at` will be set to `CURRENT_TIMESTAMP`.

User can override this behaviour by setting the value of it explicitly, i.e. `Set(xxx)`.

## Additional Information

Related Discussions:
- https://github.com/SeaQL/sea-orm/discussions/819

liberwang1013:
## PR Info

- Closes #827 

## Adds

- [x] add `created_at` & `updated_at` attributes support when define model

## More Details
In this implement, the timestamp column type should be defined as `TIMESTAMP` or `TIMESTAMPTZ`, so the Model must be defined line below
```
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i64,
    #[sea_orm(default_value = "Sam")]
    pub name: String,
    #[sea_orm(created_at, column_type = "TimestampWithTimeZone")]
    #[sea_orm(updated_at, column_type = "Timestamp")]
    pub updated_at: DateTime
}
```


sea-orm-macros/src/derives/entity_model.rs
```diff
@@ -186,6 +188,10 @@ pub fn expand_derive_entity_model(data: Data, attrs: Vec<Attribute>) -> syn::Res
                                                 indexed = true;
                                             } else if name == "unique" {
                                                 unique = true;
+                                            } else if name == "created_at" {
+                                                created_at = true;
+                                            } else if name == "updated_at" {
+                                                updated_at = true;
```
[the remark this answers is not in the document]
billy1624: For me, either way would be fine :D

billy1624 (commented):
Hey @liberwang1013, sorry for the delay. Thanks for the contributions!!

I have added some test cases, refactored a few places as well. Please review and then cherrypick :)
- https://github.com/liberwang1013/sea-orm/compare/auto-update-timestamp-column...billy1624:pr/854

liberwang1013:
> Hey @liberwang1013, sorry for the delay. Thanks for the contributions!!
> 
> I have added some test cases, refactored a few places as well. Please review and then cherrypick :)
> 
>     * [liberwang1013/sea-orm@auto-update-timestamp-column...billy1624:pr/854](https://github.com/liberwang1013/sea-orm/compare/auto-update-timestamp-column...billy1624:pr/854)

Hi @billy1624 , thank you for your commit. I have already cherrypicked your commit, your commit is great. But, now we have to face a new problem that we didn't pass all checks. I have searched some [documents](https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html) try to figure out what's the problem .  In this  [case](https://github.com/SeaQL/sea-orm/runs/7656111873?check_suite_focus=true), it seems that  default value the system given is conflicted with msyql default ```sql_mode``` setttings.
```
CREATE TABLE `check` (
  `id` int NOT NULL AUTO_INCREMENT PRIMARY KEY,
  `pay` varchar(255) NOT NULL,
  `amount` double NOT NULL,
  `updated_at` timestamp NOT NULL,
  `created_at` timestamp NOT NULL
)
```

I think there are three ways to fix it
-  remove the ```NOT NULL``` constrain for ```updated_at``` & ```created_at``` columns
- give the default value ```CURRENT_TIMESTAMP```  for ```updated_at``` & ```created_at``` columns
- disable ```NO_ZERO_DATE``` sql mode in this case

what's your opinion?

tyt2y3:
Yes, actually if you want a robust timestamps, the best way is to rely on DB's native `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`. And I think we need better SeaQuery support for this

billy1624:
> Yes, actually if you want a robust timestamps, the best way is to rely on DB's native `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`. And I think we need better SeaQuery support for this

Note that not all database support such `ON UPDATE CURRENT_TIMESTAMP` syntax. For example, PostgreSQL doesn't have it. Instead, a trigger is needed to perform the operation. See _https://stackoverflow.com/a/1036010/7059723_

billy1624:
Okay, we have two problems here

---

1. Define timestamp column with default value of `CURRENT_TIMESTAMP`

```rs
#[sea_orm(table_name = "check")]
pub struct Model {
    ...
    #[sea_orm(updated_at, default_expr = "Func::current_timestamp()")]
    pub updated_at: DateTimeWithTimeZone,
    #[sea_orm(created_at, default_expr = "Func::current_timestamp()")]
}
```

Or, we provide a shorthand for it

```rs
#[sea_orm(table_name = "check")]
pub struct Model {
    ...
    #[sea_orm(updated_at, default_current_timestamp)]
    pub updated_at: DateTimeWithTimeZone,
    #[sea_orm(created_at, default_current_timestamp)]
}
```

Related issues:
- https://github.com/SeaQL/sea-query/issues/347

---

2. Provide db specific way to set `ON UPDATE CURRENT_TIMESTAMP`

For example,
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html
- PostgreSQL: https://stackoverflow.com/a/1036010/7059723
- SQLite: No easy way to achieve it except https://stackoverflow.com/q/6578439/7059723

Related issues:
- https://github.com/SeaQL/sea-query/issues/403

liberwang1013:
> 
```
#[sea_orm(table_name = "check")]
pub struct Model {
    ...
    #[sea_orm(updated_at, default_expr = "Func::current_timestamp()")]
    pub updated_at: DateTimeWithTimeZone,
    #[sea_orm(created_at, default_expr = "Func::current_timestamp()")]
}
```
In this way, I prefer using `CURRENT_TIMESTAMP` directly. In some cases, the developer can use some  sea-orm still unsupported expresssion such as `Generated Columns` features.

liberwang1013:
> Okay, we have two problems here
> 
>     1. Define timestamp column with default value of `CURRENT_TIMESTAMP`
> 
> 
> ```rust
> #[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
> #[sea_orm(table_name = "check")]
> pub struct Model {
>     ...
>     #[sea_orm(updated_at, default_expr = "Func::current_timestamp()")]
>     pub updated_at: DateTimeWithTimeZone,
>     #[sea_orm(created_at, default_expr = "Func::current_timestamp()")]
>     pub created_at: DateTimeWithTimeZone,
> }
> ```
> 
> Also, we could provide a shorthand for it
> 
> ```rust
> #[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
> #[sea_orm(table_name = "check")]
> pub struct Model {
>     ...
>     #[sea_orm(updated_at, default_current_timestamp)]
>     pub updated_at: DateTimeWithTimeZone,
>     #[sea_orm(created_at, default_current_timestamp)]
>     pub created_at: DateTimeWithTimeZone,
> }
> ```
> 
> Related issues:
> 
>     * [Table Column with Default Expression sea-query#347](https://github.com/SeaQL/sea-query/issues/347)
> 
> 
>     2. Provide db specific way to set `ON UPDATE CURRENT_TIMESTAMP`
> 
> 
> For example,
> 
>     * MySQL: https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html
> 
>     * PostgreSQL: https://stackoverflow.com/a/1036010/7059723
> 
>     * SQLite: No easy way to achieve it except https://stackoverflow.com/q/6578439/7059723
> 
> 
> Related issues:
> 
>     * [Add support for Postgresql's function and trigger sea-query#403](https://github.com/SeaQL/sea-query/issues/403)

I think I have found a better way to do this: support `extra` macro attribute.

billy1624:
> I think I have found a better way to do this: support `extra` macro attribute.

I'd say the `extra` attribute is good to have and it serve as a temporary workaround before default expression land. :)

billy1624 (approved):
Thanks!! @liberwang1013

tyt2y3:
https://docs.rs/sea-query/latest/sea_query/table/struct.ColumnDef.html#method.default in SeaQuery now accpets a `Expr` so we can now write:

```rust
let table = Table::create()
    .table(Char::Table)
    .col(ColumnDef::new(Char::FontId).integer().default(12i32))
    .col(
        ColumnDef::new(Char::CreatedAt)
            .timestamp()
            .default(Expr::current_timestamp())
            .not_null(),
    )
    .to_owned();

assert_eq!(
    table.to_string(MysqlQueryBuilder),
    [
        "CREATE TABLE `character` (",
        "`font_id` int DEFAULT 12,",
        "`created_at` timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL",
        ")",
    ]
    .join(" ")
);
```

src/schema/entity.rs
```diff
@@ -211,6 +211,9 @@ where
     if let Some(value) = orm_column_def.default_value {
         column_def.default(value);
     }
+    if let Some(value) = orm_column_def.extra {
+        column_def.default(value);
+    }
```
billy1624, line 216: Found a typo

```suggestion
    if let Some(value) = orm_column_def.extra {
        column_def.extra(value);
    }
```

billy1624:
Hey @liberwang1013, please check https://github.com/liberwang1013/sea-orm/pull/6

src/schema/entity.rs
liberwang1013: Hi @billy1624 thank you for your comment,
what about replace with `column_def.default(SimpleExpr::CustomWithExpr(value, vec![]));`, I found the generated default expr in schema is going to quoted with `'` in this(`column_def.default(value);`) way.

tyt2y3:
Sorry but this will not be on 0.11

sutt0n:
Any updates on this functionality getting in, or has something else similar made its way into the crate?

carlocorradini:
Any update? 🤯
This is super useful ☺️

MalphasSecondCommand:
Can I know when this will be added? This is not only very useful, but modern ORMs should have this feature. SeaORM is pretty good at what it does, this is the only con I have found so far. Not only `created_at` and `updated_at`, but adding `deleted_at` would also be good for soft deletes. I don't know if that's already there, sorry if so 😅.

tyt2y3:
In addition to the comment above (https://github.com/SeaQL/sea-orm/pull/854#issuecomment-1396548404)
this feature was added in https://github.com/SeaQL/sea-orm/pull/1474

https://github.com/SeaQL/sea-orm/blob/698c98f01030b70e8ef011866b631535b20764e9/CHANGELOG.md?plain=1#L1237-L1244

If you want timestamp to be generated client side, you can override `ActiveModelBehaviour` https://github.com/SeaQL/sea-orm/blob/698c98f01030b70e8ef011866b631535b20764e9/tests/common/bakery_chain/cake.rs#L54-L60

tyt2y3:
the presented solution should address the original need for this PR. though not exactly the same API.
this PR has diverged from master so much that, we need to reimplement it in new ways.
I think the `updated_at` feature is still valuable, because there is no native way to do this cross-DB.
we can implement that

wvffle:
> we can implement that

@tyt2y3 Could you link a PR related to this functionality when it is available? And maybe reopen #827 to track the issue?

Issue #753, opened 2022-05-19
Allow user to retrieve connection from underlining pool


Sytten:
## Motivation

For some specific reason (related to us still using sqlx migrations), I happen to need an sqlx connection from the pool which I don't seem to be able to get. It would be nice to have a feature that allows us to get it.

## Proposed Solutions

1. Add a crate feature (`i-hop-in-breaking-changes` or something like that)
2. Add a method under this feature flag to get the raw sqlx connection from the database pool

billy1624:
Hey @Sytten, sorry for the delay. I think it's rather more of a design choice to include such `unstable` features. As SeaORM is designed in a way that's agnostic to both database and database connector. Later, we might swap the connector / introduce more connector and not solely rely on `sqlx`, so I consider the proposed feature as `unstable`.

We could introduce `unstable/get_sqlx_connect_pool` feature and offer the `get_sqlx_connect_pool` method under that feature.

Thoughts? @tyt2y3

tyt2y3:
I think we can have an `sqlx-internals` feature to expose some implementation details under a 'unstable' API.
It's just a type system thing anyway.

frederikhors:
I think we should allow this (https://github.com/SeaQL/sea-query/discussions/437#discussioncomment-3614048).

tyt2y3:
Could do. Although the original design is: you create a SQLx pool first and then create a SeaORM pool from it.

frederikhors:
Yeah, I know it's strange! 😄

I'm using this code right now:

```rust
use sea_orm_migration::prelude::*;
use sqlx::PgPool;

mod rs;

#[tokio::main]
async fn main() {
    let pool = PgPool::connect("...").await.unwrap();

    sqlx::migrate!("sql").run(&pool).await.unwrap();

    let db_conn = Database::connect("...").await;

    rs::Migrator::up(&db_conn, None).await.unwrap();
}
```

But I don't care for now because I'm starting and because it's a one-time bin for new demo deployments...

But it would be awesome to not repeat the connection for `sqlx`.

Or maybe we can have something like `migrate!` in Sea too!

PR #863, opened 2022-07-11
[sea-orm-linter] statement builder with plugin

billy1624:
## PR Info

Drafting a statement builder plugin API. Specifically designed for SeaORM linter :)

CC @kirawi

### DIFF
--- a/src/database/statement.rs
+++ b/src/database/statement.rs
@@ -16,9 +16,50 @@ pub struct Statement {
 }
 
 /// Constraints for building a [Statement]
-pub trait StatementBuilder {
+pub trait StatementBuilder: IntoAnyStatement {
     /// Method to call in order to build a [Statement]
     fn build(&self, db_backend: &DbBackend) -> Statement;
+
+    fn build_with_plugins<I, T>(&self, db_backend: &DbBackend, plugins: I) -> Statement
+    where
+        I: IntoIterator<Item = T>,
+        T: StatementBuilderPlugin,
+    {
+        let stmt = self.build(db_backend);
+        let any_stmt = self.into_any_statement();
+        for plugin in plugins {
+            plugin.run(&any_stmt);
+        }
+        stmt
+    }
+}
+
+#[derive(Debug)]
+pub enum AnyStatement<'a> {
+    Insert(&'a sea_query::InsertStatement),
+    Select(&'a sea_query::SelectStatement),
+    Update(&'a sea_query::UpdateStatement),
+    Delete(&'a sea_query::DeleteStatement),
+    TableCreate(&'a sea_query::TableCreateStatement),
+    TableDrop(&'a sea_query::TableDropStatement),
+    TableAlter(&'a sea_query::TableAlterStatement),
+    TableRename(&'a sea_query::TableRenameStatement),
+    TableTruncate(&'a sea_query::TableTruncateStatement),
+    IndexCreate(&'a sea_query::IndexCreateStatement),
+    IndexDrop(&'a sea_query::IndexDropStatement),
+    ForeignKeyCreate(&'a sea_query::ForeignKeyCreateStatement),
+    ForeignKeyDrop(&'a sea_query::ForeignKeyDropStatement),
+    TypeAlter(&'a sea_query::extension::postgres::TypeAlterStatement),
+    TypeCreate(&'a sea_query::extension::postgres::TypeCreateStatement),
+    TypeDrop(&'a sea_query::extension::postgres::TypeDropStatement),
+}
+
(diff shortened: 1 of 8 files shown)


src/linter.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn run(&self, stmt: &AnyStatement) {
+        match stmt {
+            AnyStatement::Insert(stmt) => InsertQueryLinter::run(stmt),
+            AnyStatement::Select(stmt) => SelectQueryLinter::run(stmt),
+            AnyStatement::Update(stmt) => UpdateQueryLinter::run(stmt),
+            AnyStatement::Delete(stmt) => DeleteQueryLinter::run(stmt),
+            _ => {}
+        }
+    }
+}
```
billy1624, line 33: Linter should be defined in a separate crate. But for simplicity and demo purpose. I just place it inside SeaORM.

src/database/statement.rs
```diff
@@ -16,9 +16,50 @@ pub struct Statement {
...
+        I: IntoIterator<Item = T>,
+        T: StatementBuilderPlugin,
+    {
+        let stmt = self.build(db_backend);
+        let any_stmt = self.into_any_statement();
+        for plugin in plugins {
+            plugin.run(&any_stmt);
+        }
+        stmt
+    }
```
billy1624, line 34: Build a SeaQuery statement with linters (plugins)

src/database/statement.rs
```diff
@@ -16,9 +16,50 @@ pub struct Statement {
...
+}
+
+#[derive(Debug)]
+pub enum AnyStatement<'a> {
+    Insert(&'a sea_query::InsertStatement),
+    Select(&'a sea_query::SelectStatement),
+    Update(&'a sea_query::UpdateStatement),
+    Delete(&'a sea_query::DeleteStatement),
```
billy1624, line 42: Any statement without `dyn`

kirawi:
Thanks! This seems like a better solution than what I was attempting.

src/executor/select.rs
```diff
@@ -423,7 +423,8 @@ where
         C: ConnectionTrait,
     {
         let builder = db.get_database_backend();
-        let stmt = builder.build(&self.query);
+        let linter = QueryLinter::new(db);
+        let stmt = builder.build_with_plugins(&self.query, [linter]);
```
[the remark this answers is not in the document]
kirawi: I can think of a couple different approaches, but it's unclear to me what the requirements for a minimal integration point for the linter are. This PR seems to favor an explicit use of the linter by the end user over an implicit approach like my PR did. Is that what we're aiming for?

kirawi:
One issue is that there currently isn't a way to access the sqlx pool through `ConnectionTrait`. Maybe a method could be added to return one?

billy1624:
> One issue is that there currently isn't a way to access the sqlx pool through `ConnectionTrait`. Maybe a method could be added to return one?

Hey @kirawi, related issues:
- https://github.com/SeaQL/sea-orm/issues/753

No active PR at the moment.

But why would you need the underlying SQLx driver? Why not just execute the statement via SeaORM?
- https://www.sea-ql.org/SeaORM/docs/basic-crud/raw-sql#use-raw-query--execute-interface

kirawi:
`sea-schema` seems to require an sqlx pool for schema discovery, e.g. https://docs.rs/sea-schema/latest/sea_schema/sqlite/discovery/struct.SchemaDiscovery.html#method.new

tyt2y3:
I think this is a great starting point, especially the `AnyStatement`
Obviously, we should not `Linter::new()` everywhere. Under SeaORM's lint mode, we should
1) run schema discovery and store it inside the connection pool (`Arc` as it should be immutable)
2) initialize a Linter and again `Arc` it (I expect it to be immutable afterwards)
3) run entity schema check first (they may not be in sync!)
4) during query exec, invoke the linter to do some checking

tyt2y3:
This is the *bare minimum* MVP we are looking for

tyt2y3:
And, there should exist a `Linter` trait to act as an adapter, and that we can expose a macro to magically glue things together (like sea_query driver), so in the end, neither `sea-orm` nor `sea-orm-linter` depends on one another.

kirawi:
Is there a reason `sea-schema` requires an sqlx pool to detect the schema, rather than just a connection?

tyt2y3:
> Is there a reason `sea-schema` requires an sqlx pool to detect the schema, rather than just a connection?

No. We can probably change the API of sea-schema to accept a connection

kirawi:
No worries. I read the docs for `Pool` and a clone is cheap.

PR #879, opened 2022-07-16
Check to make sure migration name is valid

AngelOnFira:
I created a migration named `initial-tables`, however, I had to change it since it wasn't valid as a Rust module name.

image

I'm sure there are some other characters that should be filtered out, but I didn't look too deep. Also, I created a custom error for this, but I'm not super sure if that's the cleanest place to put it. I can make any changes necessary :)

### DIFF
--- a/sea-orm-cli/src/commands.rs
+++ b/sea-orm-cli/src/commands.rs
@@ -253,6 +253,14 @@ pub fn run_migrate_command(
             return Ok(());
         }
         Some(MigrateSubcommands::Generate { migration_name }) => {
+            // Make sure the migration name doesn't contain any characters that
+            // are invalid module names in Rust.
+            if migration_name.contains('-') {
+                return Err(Box::new(MigrationCommandError::InvalidName(
+                    "Invalid character used in migration name".to_string(),
+                )));
+            }
+
             println!("Generating new migration...");
 
             // build new migration filename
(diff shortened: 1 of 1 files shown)


tyt2y3:
Hi @AngelOnFira welcome back!

sea-orm-cli/src/commands.rs
```diff
@@ -253,6 +253,14 @@ pub fn run_migrate_command(
             return Ok(());
         }
         Some(MigrateSubcommands::Generate { migration_name }) => {
+            // Make sure the migration name doesn't contain any characters that
+            // are invalid module names in Rust.
+            if migration_name.contains('-') {
+                return Err(Box::new(MigrationCommandError::InvalidName(
+                    "Invalid character used in migration name".to_string(),
+                )));
+            }
```
billy1624, line 262: Hey @AngelOnFira, sorry for the delay. I think the error message can be more specific? E.g. hyphen `-` cannot be used in migration name

billy1624 (commented):
One more nitpick, could we have a test case for this new behaviour? I want this behaviour to be checked and persisted in years to come :)
Thank you so much for the great contributions!

sea-orm-cli/src/commands.rs
billy1624: Such as...

```suggestion
            // Make sure the migration name doesn't contain any characters that
            // are invalid module names in Rust.
            if migration_name.contains('-') {
                return Err(Box::new(MigrationCommandError::InvalidName(
                    "Hyphen `-` cannot be used in migration name".to_string(),
                )));
            }
```

billy1624 (commented):
Hey @AngelOnFira, I'll continue the work on https://github.com/SeaQL/sea-orm/pull/1155 :)

PR #895, opened 2022-07-20
[sea-orm-linter] continuing statement builder plugin

kirawi:
This is a continuation of #863. It adds an `ExecutorType` enum to distinguish what executor the plugin is being called from, likely needed for `Paginator`'s lints. I am trying to make the plugin loading generic, but I'm currently running into type issues which I will try to fix tomorrow. After this PR is finished, I'll update the linter to use this version of sea-orm.

### DIFF
--- a/src/database/connection.rs
+++ b/src/database/connection.rs
@@ -40,8 +40,8 @@ pub trait ConnectionTrait: Sync {
     }
 
     /// Try to get the statement builder plugins of a connection.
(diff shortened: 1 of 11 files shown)


billy1624:
Hey @kirawi, as discussed on Discord. I think we should prioritise implementing the core of `sea-orm-liner`. Then, we circle back to bridge `sea-orm` with `sea-orm-linter`
Cheers :)

PR #902, opened 2022-07-20
distinct support in sea-orm

### LINKED ISSUE #275: Distinct support?
I saw sea-query supports it, but It doesn't seem like it's supported yet in sea-orm.

kyoto7250:
close #275 

## Adds

This PR adds `distinct` methods in `helper.rs`

- [x]  distinct
- [x] distinct on (only `sqlx-postgres`)

### DIFF
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -8,7 +8,6 @@ use sea_query::{
 };
 pub use sea_query::{Condition, ConditionalStatement, DynIden, JoinType, Order, OrderedStatement};
 
-#[cfg(feature = "sqlx-postgres")]
 use sea_query::IntoColumnRef;
 
 // LINT: when the column does not appear in tables selected from
@@ -198,7 +197,6 @@ pub trait QuerySelect: Sized {
     ///     "SELECT DISTINCT ON (\"name\") \"cake\".\"id\", \"cake\".\"name\" FROM \"cake\" WHERE \"cake\".\"name\" LIKE '%cheese%'"
     /// );
     /// ```
-    #[cfg(feature = "sqlx-postgres")]
     fn distinct_on<C>(mut self, col: C) -> Self
     where
         C: IntoColumnRef,
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -8,6 +8,9 @@ use sea_query::{
 };
 pub use sea_query::{Condition, ConditionalStatement, DynIden, JoinType, Order, OrderedStatement};
 
+#[cfg(feature = "sqlx-postgres")]
+use sea_query::IntoColumnRef;
+
 // LINT: when the column does not appear in tables selected from
 // LINT: when there is a group by clause, but some columns don't have aggregate functions
 // LINT: when the join table or column does not exists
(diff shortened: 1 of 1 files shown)


src/query/helper.rs
```diff
@@ -149,6 +152,61 @@ pub trait QuerySelect: Sized {
...
+    ///         .distinct_on(cake::Column::Name)
+    ///         .build(DbBackend::Postgres)
+    ///         .to_string(),
+    ///     "SELECT DISTINCT ON (\"name\") \"cake\".\"id\", \"cake\".\"name\" FROM \"cake\" WHERE \"cake\".\"name\" LIKE '%cheese%'"
+    /// );
+    /// ```
+    #[cfg(feature = "sqlx-postgres")]
+    fn distinct_on<C>(mut self, col: C) -> Self
+    where
+        C: IntoColumnRef,
```
billy1624, line 204: Following the method signature of SeaQuery's method, we could take a iterator here. e.g.
```rs
pub fn distinct_on<T, I>(mut self, cols: I) -> Self
    where
        T: IntoColumnRef,
        I: IntoIterator<Item = T>
{ ... }
```

src/query/helper.rs
```diff
@@ -149,6 +152,61 @@ pub trait QuerySelect: Sized {
...
+    ///         .filter(
+    ///             Condition::all().add_option(input.name.map(|n| cake::Column::Name.contains(&n)))
+    ///         )
+    ///         .distinct_on(cake::Column::Name)
+    ///         .build(DbBackend::Postgres)
+    ///         .to_string(),
+    ///     "SELECT DISTINCT ON (\"name\") \"cake\".\"id\", \"cake\".\"name\" FROM \"cake\" WHERE \"cake\".\"name\" LIKE '%cheese%'"
+    /// );
+    /// ```
+    #[cfg(feature = "sqlx-postgres")]
```
billy1624, line 201: We try to avoid feature gated method

src/query/helper.rs
```diff
@@ -8,6 +8,9 @@ use sea_query::{
 };
 pub use sea_query::{Condition, ConditionalStatement, DynIden, JoinType, Order, OrderedStatement};
 
+#[cfg(feature = "sqlx-postgres")]
```
billy1624, line 11: This line would be unnecessary if we dropped the feature gate below

billy1624 (commented):
Hey @kyoto7250, thanks for contributing!! Look pretty good. I've some comments
Please check and see if it make sense.
Thanks!! :)

src/query/helper.rs
kyoto7250: I fixed in https://github.com/SeaQL/sea-orm/pull/902/commits/db1111c0e5c5935b7e0782234906b741490628ff

src/query/helper.rs
kyoto7250: I fixed in https://github.com/SeaQL/sea-orm/pull/902/commits/c9871cb2d25de92336501799faa7eb29595ef95b

src/query/helper.rs
billy1624: Thanks!! @kyoto7250

billy1624 (approved):
Thanks again for the great contributions! @kyoto7250

PR #933, opened 2022-08-04
Enable migration generation in modules

remlse:
## Adds

Generating a new migration no works even if migrations are not at the root of their own crate. 

I didn't want to have an entire separate crate just for migrations. So I put them into a module in my regular crate. `sea-orm-cli migrate generate` accepts a flag `--migration-dir`, which seemed perfect for my use case. However, that flag still assumes there is a directory `src/` and that the migrator definition is in `lib.rs`.

If there is a directory `src/` and a file `lib.rs`, this PR doesn't change anything, so I don't think it's a breaking change. If `src/` doesn't exist, it will look in the parent directory instead. If there is no `lib.rs`, it will look for a `mod.rs` instead.

For reference, this is what my file structure looks like:
```
my-app/
├─ src/
│  ├─ db/
│  │  ├─ migration/
│  │  │  ├─ mod.rs/
│  │  │  ├─ m20220101_000001_create_table.rs/
│  │  ├─ migration_cli.rs
│  │  ├─ mod.rs
│  ├─ main.rs
├─ Cargo.toml
```
Both `db/mod.rs` and `db/migration_cli.rs` contain `mod migration;`. That way, both binaries (`main.rs`, the main app, and `migration_cli.rs`, a thin wrapper around the migration cli api) can access my migrations.

I'm looking forward to discussions and suggestions!

### DIFF
--- a/sea-orm-cli/src/cli.rs
+++ b/sea-orm-cli/src/cli.rs
@@ -25,7 +25,11 @@ pub enum Commands {
             global = true,
             short = 'd',
             long,
-            help = "Migration script directory",
+            help = "Migration script directory.
+If your migrations are in their own crate,
+you can provide the root of that crate.
+If your migrations are in a submodule of your app,
+you should provide the directory of that submodule.",
             default_value = "./migration"
         )]
         migration_dir: String,
--- a/sea-orm-cli/src/commands/migrate.rs
+++ b/sea-orm-cli/src/commands/migrate.rs
@@ -123,6 +123,15 @@ pub fn run_migrate_generate(
     Ok(())
 }
 
+/// `get_full_migration_dir` looks for a `src` directory
+/// inside of `migration_dir` and appends that to the returned path if found.
+///
+/// Otherwise, `migration_dir` can point directly to a directory containing the
+/// migrations. In that case, nothing is appended.
+///
+/// This way, `src` doesn't need to be appended in the standard case where
+/// migrations are in their own crate. If the migrations are in a submodule
+/// of another crate, `migration_dir` can point directly to that module.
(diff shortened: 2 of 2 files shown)


billy1624 (commented):
Hey @remlse, welcome! And thanks for the great contributions!

I think we should document the updated behaviour on the `get_full_migration_dir` and `get_migrator_filepath` method because the logic is getter more complex it's not easy to understand at a glance.

Which directory/file be checked for existence then what's the fallback?

Thoughts? :)

billy1624 (commented):
Hey @remlse, just notice a problem. Before updating the migrator file `mod.rs`/`lib.rs`. A backup file will be created with file extension of `.bak`.
- https://github.com/SeaQL/sea-orm/blob/1a2090fb5e7474e230c27395bb634f12a87e4e39/sea-orm-cli/src/commands/migrate.rs#L143-L144

However, now that migrator can be either reside in `mod.rs`/`lib.rs`. We should change the line to:

```rs
let migrator_backup_filepath = migrator_filepath.with_extension("rs.bak");
```

remlse:
@billy1624 thanks for having a look! I think I'll get around to it on the weekend, I'll add some documentation and fix the backup filename.

remlse:
@billy1624 I've added some documentation. Some internally and some to be displayed by `sea-orm-cli migrate -h`. Let me know if I can improve something, I haven't done much documentation before.

billy1624 (approved):
Thanks! @remlse The documentation looks great :)

tyt2y3:
Thank you for the contribution. Too bad for the delay, I am surprised this patch still applies

Issue #969, opened 2022-08-12
Support sea-query-binder


ikrivosheev:
## Motivation

Replace `sea-query-driver` to `sea-query-binder`

## Additional Information

SeaQuery issue: https://github.com/SeaQL/sea-query/issues/383

PR #970, opened 2022-08-12
issues-969 Replace sea-query-driver to sea-query-binder

ikrivosheev:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/969

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/416

## Adds

- Replace `sea-query-driver` to `sea-query-binder`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ authors = ["Chris Tsang <tyt2y7@gmail.com>"]
 edition = "2021"
 description = "🐚 An async & dynamic ORM for Rust"
 license = "MIT OR Apache-2.0"
+homepage = "https://www.sea-ql.org/SeaORM"
 documentation = "https://docs.rs/sea-orm"
 repository = "https://github.com/SeaQL/sea-orm"
 categories = ["database"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ path = "src/lib.rs"
 [dependencies]
 async-stream = { version = "^0.3" }
 async-trait = { version = "^0.1" }
-chrono = { version = "^0.4.20", default-features = false, features = ["clock"], optional = true }
+chrono = { version = "^0.4", optional = true }
 time = { version = "^0.3", optional = true }
 futures = { version = "^0.3" }
 futures-util = { version = "^0.3" }
@@ -35,7 +35,7 @@ tracing = { version = "^0.1", features = ["log"] }
 rust_decimal = { version = "^1", optional = true }
 sea-orm-macros = { version = "^0.9.1", path = "sea-orm-macros", optional = true }
 sea-query = { git = "https://github.com/ikrivosheev/sea-query", version = "^0.27", features = ["thread-safe"] }
-sea-query-binder = { git = "https://github.com/ikrivosheev/sea-query", version = "^0.1" }
+sea-query-binder = { git = "https://github.com/ikrivosheev/sea-query", version = "^0.1", optional = true }
 sea-strum = { version = "^0.23", features = ["derive", "sea-orm"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0", optional = true }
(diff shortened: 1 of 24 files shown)


Cargo.toml
```diff
@@ -68,28 +69,45 @@ default = [
...
+with-uuid = ["uuid", "sea-query/with-uuid", "sea-query-binder?/with-uuid", "sqlx?/uuid"]
+with-time = ["time", "sea-query/with-time", "sea-query-binder?/with-time", "sqlx?/time"]
 sqlx-dep = []
 sqlx-all = ["sqlx-mysql", "sqlx-postgres", "sqlx-sqlite"]
-sqlx-mysql = ["sqlx-dep", "sea-query/sqlx-mysql", "sqlx/mysql"]
-sqlx-postgres = ["sqlx-dep", "sea-query/sqlx-postgres", "sqlx/postgres"]
-sqlx-sqlite = ["sqlx-dep", "sea-query/sqlx-sqlite", "sqlx/sqlite"]
+sqlx-mysql = ["sqlx-dep", "sea-query-binder/sqlx-mysql", "sqlx/mysql"]
+sqlx-postgres = ["sqlx-dep", "sea-query-binder/sqlx-postgres", "sqlx/postgres"]
+sqlx-sqlite = ["sqlx-dep", "sea-query-binder/sqlx-postgres", "sqlx/sqlite"]
```
billy1624, line 81: Found a typo

```suggestion
sqlx-sqlite = ["sqlx-dep", "sea-query-binder/sqlx-sqlite", "sqlx/sqlite"]
```

billy1624 (changes requested):

billy1624 (changes requested):
Hey @ikrivosheev, maybe we could get https://github.com/SeaQL/sea-query/pull/416 done then turn our heads back to this?

Btw... I notice a breaking change in the binder of SQLite. The old haviour is to convert rust decimal and big decimal into `f64` for binding. However, the new behaviour will simply panic.

> thread 'main' panicked at 'Sqlite doesn't support decimal arguments', /home/runner/.cargo/git/checkouts/sea-query-4258bc96379dd872/50ee12e/sea-query-binder/src/sqlx_sqlite.rs:98:21

https://github.com/SeaQL/sea-orm/runs/7913388359?check_suite_focus=true#step:5:300

src/driver/sqlx_sqlite.rs
```diff
@@ -222,10 +222,10 @@ impl From<SqliteQueryResult> for ExecResult {
...
-    let mut query = sqlx::query(&stmt.sql);
+pub(crate) fn sqlx_query(stmt: &Statement) -> sqlx::query::Query<'_, Sqlite, SqlxValues> {
     if let Some(values) = &stmt.values {
-        query = bind_query(query, values);
+        sqlx::query_with(&stmt.sql, SqlxValues(values.clone()))
+    } else {
+        sqlx::query_with(&stmt.sql, SqlxValues(Values(Vec::new())))
     }
-    query
 }
```
billy1624, line 231: I'd like to rewrite this.

Nit:
```rs
pub(crate) fn sqlx_query(stmt: &Statement) -> sqlx::query::Query<'_, Sqlite, SqlxValues> {
    let values = stmt.values.as_ref().map_or(Values(Vec::new()), |values| values.clone());
    sqlx::query_with(&stmt.sql, SqlxValues(values))
}
```

src/driver/sqlx_postgres.rs
```diff
@@ -215,10 +215,10 @@ impl From<PgQueryResult> for ExecResult {
...
-    let mut query = sqlx::query(&stmt.sql);
+pub(crate) fn sqlx_query(stmt: &Statement) -> sqlx::query::Query<'_, Postgres, SqlxValues> {
     if let Some(values) = &stmt.values {
-        query = bind_query(query, values);
+        sqlx::query_with(&stmt.sql, SqlxValues(values.clone()))
+    } else {
+        sqlx::query_with(&stmt.sql, SqlxValues(Values(Vec::new())))
     }
-    query
 }
```
billy1624, line 224: Btw... I'm curious if we could avoid the `clone` by taking ownership of the `stmt: Statement`?

ikrivosheev:
@billy1624 I will return to this PR after merge PRs in SeaQuery.

src/driver/sqlx_sqlite.rs
ikrivosheev: Thank you, done!

src/driver/sqlx_postgres.rs
ikrivosheev: I cannot take ownership of the `stmt: Statement`, because after `sqlx_query` we needed to send metrics:

```rust
pub async fn query_all(&self, stmt: Statement) -> Result<Vec<QueryResult>, DbErr> {
    debug_print!("{}", stmt);

    let query = sqlx_query(&stmt);
    if let Ok(conn) = &mut self.pool.acquire().await {
        crate::metric::metric!(self.metric_callback, &stmt, {
            match query.fetch_all(conn).await {
                Ok(rows) => Ok(rows.into_iter().map(|r| r.into()).collect()),
                Err(err) => Err(sqlx_error_to_query_err(err)),
            }
        })
    } else {
        Err(DbErr::Query(
            "Failed to acquire connection from pool.".to_owned(),
        ))
    }
}
```

billy1624: Right!

We can change `pub statement: &'a crate::Statement` into `pub statement: String` but then, worse still we need to `to_string` on `Statement` ahead of `metrics!`.
- https://github.com/SeaQL/sea-orm/blob/d532deb6363d997beae029a8983d64c7e9574af6/src/metric.rs#L8-L17

So, I'll support as-is - cloning `stmt: Statement`. Later in time, when we're optimizing for performance, We can move metric utility to compile time, i.e. provide a wrapper to existing connection. So that we don't have to clone `Statement` even when metric is turned off.

CC @tyt2y3

ikrivosheev: Can't do it anyway, because:

```
pub fn query_with<'q, DB, A>(sql: &'q str, arguments: A) -> Query<'q, DB, A> ...
```

`query_with` doesnot take ownership of the query...

ikrivosheev: I can make PR for add support `QueryOwned` into sqlx and this would help us.

Issue #960, opened 2022-08-11
`fn column()` also handle enum type


tyt2y3:
Looks like we are doing some magic inside `fn column_list(&self) -> Vec<SimpleExpr> {` of `select.rs`.
It should also be possible to do the same in `fn column()` ? @billy1624 
Then it will be transparent to the user?

_Originally posted by @tyt2y3 in https://github.com/SeaQL/sea-orm/discussions/843#discussioncomment-3370904_

billy1624:
Good call!

PR #973, opened 2022-08-15
Rewrite enum and text casting

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/960

## Adds

- [x] `QuerySelect::column` method also handle enum type

## Changes

- [x] Refactor all enum and text casting in SeaORM

### DIFF
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -38,6 +38,29 @@ pub trait QuerySelect: Sized {
     ///     r#"SELECT "cake"."name" FROM "cake""#
     /// );
     /// ```
+    ///
+    /// Enum column will be casted into text (PostgreSQL only)
+    ///
+    /// ```
+    /// use sea_orm::{entity::*, query::*, tests_cfg::lunch_set, DbBackend};
+    ///
+    /// assert_eq!(
+    ///     lunch_set::Entity::find()
+    ///         .select_only()
+    ///         .column(lunch_set::Column::Tea)
+    ///         .build(DbBackend::Postgres)
+    ///         .to_string(),
+    ///     r#"SELECT CAST("lunch_set"."tea" AS text) FROM "lunch_set""#
+    /// );
+    /// assert_eq!(
+    ///     lunch_set::Entity::find()
+    ///         .select_only()
+    ///         .column(lunch_set::Column::Tea)
+    ///         .build(DbBackend::MySql)
+    ///         .to_string(),
+    ///     r#"SELECT `lunch_set`.`tea` FROM `lunch_set`"#
+    /// );
+    /// ```
     fn column<C>(mut self, col: C) -> Self
     where
         C: ColumnTrait,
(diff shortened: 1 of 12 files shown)


src/query/helper.rs
```diff
@@ -532,3 +555,32 @@ pub(crate) fn unpack_table_alias(table_ref: &TableRef) -> Option<DynIden> {
...
+pub(crate) fn cast_text_as_enum<C>(expr: Expr, col: &C) -> SimpleExpr
+where
+    C: ColumnTrait,
+{
+    cast_enum_text_inner(expr, col, |col, enum_name| {
+        col.as_enum(Alias::new(enum_name))
+    })
+}
+
+fn cast_enum_text_inner<C, F>(expr: Expr, col: &C, f: F) -> SimpleExpr
```
tyt2y3, line 575: I am glad that we finally extracted this piece of common code.

src/query/helper.rs
```diff
@@ -532,3 +555,32 @@ pub(crate) fn unpack_table_alias(table_ref: &TableRef) -> Option<DynIden> {
         | TableRef::DatabaseSchemaTableAlias(_, _, _, alias) => Some(SeaRc::clone(alias)),
     }
 }
+
+pub(crate) fn cast_enum_as_text<C>(expr: Expr, col: &C) -> SimpleExpr
+where
+    C: ColumnTrait,
+{
+    cast_enum_text_inner(expr, col, |col, _| col.as_enum(Alias::new("text")))
```
tyt2y3, line 563: Is it possible to use a const, i.e. `#[derive(Iden)] struct Text` ?

src/query/helper.rs
```diff
@@ -532,3 +555,32 @@ pub(crate) fn unpack_table_alias(table_ref: &TableRef) -> Option<DynIden> {
...
+{
+    cast_enum_text_inner(expr, col, |col, enum_name| {
+        col.as_enum(Alias::new(enum_name))
+    })
+}
+
+fn cast_enum_text_inner<C, F>(expr: Expr, col: &C, f: F) -> SimpleExpr
+where
+    C: ColumnTrait,
+    F: Fn(Expr, &String) -> SimpleExpr,
```
tyt2y3, line 578: Same here, I might have lost the context, but why are we passing enum name around as String? Can we pass around DynIden instead?

src/query/helper.rs
billy1624: Sure!

src/query/helper.rs
billy1624: The `String` comes from `ColumnType::Enum`
- https://github.com/SeaQL/sea-orm/blob/87f7891f175d662280ede2a234421bedac9507ca/src/entity/column.rs#L81

I rewrite it into `F: Fn(Expr, DynIden) -> SimpleExpr`, but still, we need to wrap the `String` in `DynIden` like `SeaRc::new(Alias::new(enum_name))`

tyt2y3: Um... I think we should have used an Iden here. Second question is, how often is this `ColumnType` object being cloned? If it is being cloned frequently then we'd better `Arc` it.

Issue #965, opened 2022-08-12
[codegen] Struct / enum derive `PartialEq` should also derive `Eq`


billy1624:
## Motivation

New clippy lint was added in Rust 1.63.0. which checks for types that derive `PartialEq` and could implement `Eq`.

Example:

```rs
#[derive(PartialEq)]
struct Foo {
    i_am_eq: i32,
    i_am_eq_too: Vec<String>,
}
```

Do this instead:

```rs
#[derive(PartialEq, Eq)]
struct Foo {
    i_am_eq: i32,
    i_am_eq_too: Vec<String>,
}
```

## Proposed Solutions

Check `sea-orm-codegen` where `#[derive(PartialEq)]` should be changed to `#[derive(PartialEq, Eq)]`.

## Additional Information

Clippy lints index: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

baoyachi:
I‘m try

billy1624:
Thanks! @baoyachi

baoyachi:
Oh，I see you've done it。

billy1624:
Nope, it's not

billy1624:
https://github.com/SeaQL/sea-orm/pull/967 doesn't touch codegen code writer

baoyachi:
I’m try

baoyachi:
It's seem's touch sea-query create

```diff
...
/// Value variants
///
/// We want Value to be exactly 1 pointer sized, so anything larger should be boxed.
+ #[derive(Clone, Debug, PartialEq)]
pub enum Value {
    Bool(Option<bool>),
    TinyInt(Option<i8>),
    SmallInt(Option<i16>),
    Int(Option<i32>),
...
```

baoyachi:
So，first need change sea-query?

tyt2y3:
I don't think we need an option for this behaviour. It seems to be universally true that we want Eq and PartialEq wherever possible.

billy1624:
There are only three place in the codegen we need to modify:
- https://github.com/SeaQL/sea-orm/blob/d262501a44c048ed36a17bbf661ced3c5193f455/sea-orm-codegen/src/entity/active_enum.rs#L33-L42
- https://github.com/SeaQL/sea-orm/blob/d262501a44c048ed36a17bbf661ced3c5193f455/sea-orm-codegen/src/entity/writer.rs#L387-L390
[...]

PR #988, opened 2022-08-24
Struct / enum derive PartialEq should also derive Eq

w93163red:
## PR Info

- Closes 
https://github.com/SeaQL/sea-orm/issues/965

## Adds

[codegen] Struct / enum derive PartialEq should also derive Eq

## Changes

Add a function to check if f32/f64 exists on the column. If not, add Eq on the derived part.

Testing:
1. update the previous tests to match the new mechanism. 
2. Add two tests to if Eq is not generated when f32/f64 in the column.


billy1624 (changes requested):
Hey @w93163red, thanks for contributing!! Please check https://github.com/w93163red/sea-orm/pull/1 for the refactoring and fixes :D

w93163red:
I merged the change~ Thanks!

billy1624 (approved):
Thanks again for contributing!! @w93163red

tyt2y3:
Awesome

frederikhors:
It would be very useful to have this as soon as possible.

Every time we generate with 0.9.3 we have to undo the changes (I cannot disable linting in CI).

Can we release 0.9.4 with this only?

tyt2y3:
Can

frederikhors:
> Can do

? Are you going to?

Issue #1007, opened 2022-08-30
sea-orm-cli migrate up can only apply 1 migration


ringrid:
sea-orm-cli version: 0.9.2
`sea-orm-cli migrate up` can only apply 1 pending migration, how does it apply all ?

ctfdavis:
I think the `sea-orm-cli migrate up` only applies 1 pending migration. The source code of the `up` command has a default value `1` for the `num` option:

```rust
#[clap(about = "Apply pending migrations")]
Up {
    #[clap(
        value_parser,
        short,
        long,
        default_value = "1",
        help = "Number of pending migrations to apply"
    )]
    num: u32,
},
#[clap(value_parser, about = "Rollback applied migrations")]
 Down {
    #[clap(
        value_parser,
        short,
        long,
        default_value = "1",
        help = "Number of applied migrations to be rolled back"
    )]
    num: u32,
},
```

The [documentation](https://www.sea-ql.org/SeaORM/docs/migration/running-migration/) is wrong on this one. As it stands, I think there is no way to migrate all pending migrations using the cli.

billy1624:
Hey @ringrid @ctfdavis, the intended behaviour is that `sea-orm-cli migrate up` should apply all pending migration. While `sea-orm-cli migrate up -n 1` only apply the first pending migration.

As a temporary workaround: execute `sea-orm-cli migrate` on your migration crate. It will apply all pending migration at once.

Issue #989, opened 2022-08-25
cargo run -- up can not update the table field


diaojunxian:
## Description

In poem_example，update the table field in entity posts, add two fields, and then add two fields in 

`migration/ m20220120_000001_create_post_table`

```
 .col(ColumnDef::new(Posts::Password).string().not_null())
.col(ColumnDef::new(Posts::Username).string().not_null())
```

```
#[derive(Iden)]
enum Posts {
    Table,
    Id,
    Title,
    Text,
    Password,
    Username
}
```

then execute `migration/main.rs`, 

and the execution result is

![image](https://user-images.githubusercontent.com/19700467/186567841-61744056-1a74-4fd2-b053-491f84127c77.png)

and find that the fields in the database posts table are not updated.

## why

1. update the fields and migration show `Applying 1 pending migrations `,but the result is `No pending migrations`

ctfdavis:
That's the expected behaviour according to the source code:

```rust
async fn up(db: &DbConn, mut steps: Option<u32>) -> Result<(), DbErr> {
    Self::install(db).await?;
    let manager = SchemaManager::new(db);

    if let Some(steps) = steps {
        info!("Applying {} pending migrations", steps);
    } else {
        info!("Applying all pending migrations");
    }

    let migrations = Self::get_pending_migrations(db).await?.into_iter();
    if migrations.len() == 0 {
        info!("No pending migrations");
    }
    // skipped
 }
```

The cli first checks the `steps` option and displays a message immediately. After that, it checks the database to know how many migrations are pending and then displays another message.

Therefore, in your attempt there is no pending migration. To add a new migration, you should generate a new migration as documented [here](https://www.sea-ql.org/SeaORM/docs/migration/writing-migration/#creating-migrations). Alternatively, if you just want to edit the already migrated file, you can first roll it back before running it again. Hope this helps.

billy1624:
Hey @diaojunxian, welcome! Sorry for the delay. I'm pretty sure the migration `m20220120_000001_create_post_table` has already been applied. That's why you alter the migration script of it and execute `cargo run -- up` do nothing for you. Because the migration you just edit has been applied and it won't apply it again.

You have two options:
1. Rollback the migration then re-apply it: execute `cargo run -- down` then `carg run -- up`
2. Create a new migration with scripts to add those two columns to the table

billy1624:
Thanks!! @ctfdavis

billy1624:
This might be related
[...]

PR #1010, opened 2022-09-01
[cli] migrate up should apply all pending migrations

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/989
- Closes https://github.com/SeaQL/sea-orm/issues/1007

## Fixes

- [x] CLI migrate up command shouldn't assume `--num 1` to be the default. Instead, the default should be `None`, aka apply all pending migrations


tyt2y3 (approved):
I agree with the new behaviour. Feel free to merge.

Issue #899, opened 2022-07-20
Inserting active models by `insert_many` with `on_conflict` and `do_nothing` panics if no rows are inserted.


Karahik:
## Description

`insert_many` with `on_conflit` and `do_nothing` panics if every given key conflicts with existing rows.

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/sea-orm-0.9.0/src/executor/insert.rs:121:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems that the code the error message indicates is trying to retrieve the last inserted row, but it could get nothing because insertion did not happen.

I'd expect this to do nothing (without panics).

## Steps to Reproduce

1. 
This is what my example project looks like.
```
.
├── Cargo.lock
├── Cargo.toml
├── sql
│   └── 00000000000000_schema.sql
├── src
│   ├── main.rs
│   └── orm
│       ├── fruits.rs
│       ├── mod.rs
│       └── prelude.rs
└── target
```

Cargo.toml.
```toml
[package]
name = "sea-orm-issue"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
anyhow = "1.0.58"
chrono = { version = "0.4.19", features = ["serde"] }
sea-orm = { version = "0.9.0", features = ["sqlx-postgres", "runtime-tokio-rustls", "macros"] }
serde = { version = "1.0.139", features = ["derive"] }
serde_json = "1.0.82"
sqlx = { version = "0.6.0", features = ["runtime-tokio-rustls", "postgres", "chrono"] }
tokio = { version = "1.20.0", features = ["full"] }

```

This is the example code(main.rs).
```rust
mod orm;

use crate::orm::fruits;
use sea_orm::{sea_query::OnConflict, ActiveValue, ConnectOptions, Database, EntityTrait};

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let opt =
        ConnectOptions::new("postgres://postgres:postgres@localhost:5432/postgres".to_owned());

    let db = Database::connect(opt).await?;

    let fruits = vec![
        fruits::ActiveModel {
            name: ActiveValue::set("Orange".to_owned()),
            ..Default::default()
        },
        fruits::ActiveModel {
            name: ActiveValue::set("Apple".to_owned()),
[...]

PR #1002, opened 2022-08-28
Better error system

tyt2y3:
Continue #750

PR #1021, opened 2022-09-08
PostgreSQL insert many with returning

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/899

## Breaking Changes

- [x] PostgreSQL insert many with returning - throw `DbErr::RecordNotInserted` if none of the records are being inserted

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -55,6 +55,11 @@ pub enum DbErr {
     /// A migration error
     #[error("Migration Error: {0}")]
     Migration(String),
+    /// None of the records are being inserted into the database,
+    /// if you insert with upsert expression that means
+    /// all of them conflict with existing records in the database
+    #[error("RecordNotInserted Error: {0}")]
+    RecordNotInserted(String),
 }
 
 /// Runtime error
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -142,7 +142,10 @@ where
             let cols = PrimaryKey::<A>::iter()
                 .map(|col| col.to_string())
                 .collect::<Vec<_>>();
-            let res = db.query_one(statement).await?.unwrap();
+            let rows = db.query_all(statement).await?;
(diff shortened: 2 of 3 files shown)


src/error.rs
```diff
@@ -17,6 +17,10 @@ pub enum DbErr {
     Json(String),
     /// A migration error
     Migration(String),
+    /// None of the records are being inserted into the database,
+    /// if you insert with upcert expression that means
```
tyt2y3, line 21: upsert

tyt2y3:
Thank you. Since #1002 has been merged, I'd like this to nest under the `RuntimeError` as discussed.

billy1624:
> Thank you. Since #1002 has been merged, I'd like this to nest under the `RuntimeError` as discussed.

I think some of the errors that reside at the `DbErr` root should be a variant of `RuntimeError` as well, such as `UnpackInsertId` and `UpdateGetPrimaryKey`.

So, to be consistent I'd like `RecordNotInserted` to be consistent with those.

https://github.com/SeaQL/sea-orm/blob/9d0f3f6ff6126275f2ed53d8b42ed438a87389c7/src/error.rs#L5-L75

src/executor/insert.rs
```diff
@@ -142,7 +142,10 @@ where
             let cols = PrimaryKey::<A>::iter()
                 .map(|col| col.to_string())
                 .collect::<Vec<_>>();
-            let res = db.query_one(statement).await?.unwrap();
+            let rows = db.query_all(statement).await?;
+            let res = rows.last().ok_or_else(|| {
+                DbErr::RecordNotInserted("None of the records are being inserted".to_owned())
+            })?;
```
tyt2y3, line 148: Actually after a second thought I think it all makes sense.
Because there is no SQLx error at all (if there is, we'd like to keep the `source`)

tyt2y3 (approved):

PR #1041, opened 2022-09-15
Implement `IntoActiveValue` for `time` types.

jimmycuadra:
I tried to implement a [custom active model](https://www.sea-ql.org/SeaORM/docs/advanced-query/custom-active-model/), and one of the columns was `Option<TimeDateTimeWithTimeZone>`. I got a compiler error:

```
error[E0277]: the trait bound `std::option::Option<sea_orm::prelude::TimeDateTimeWithTimeZone>: IntoActiveValue<_>` is not satisfied
```

Looking into the source code, it seemed a simple oversight that this trait was implemented for the `chrono` types but not the `time` types, and it was easy enough to fix since there's already a macro to implement it for new types.

I also noticed that the `time` types are not accounted for in `src/query/json.rs` while the `chrono` types are, which I assume is also an oversight. However, I don't have a need for that at this point and the fix for that seemed less trivial, so I'm just bringing it to your attention.

Thanks for SeaORM!

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -682,6 +682,7 @@ impl_into_active_value!(f32);
 impl_into_active_value!(f64);
 impl_into_active_value!(&'static str);
 impl_into_active_value!(String);
+impl_into_active_value!(Vec<u8>);
 
 #[cfg(feature = "with-json")]
 #[cfg_attr(docsrs, doc(cfg(feature = "with-json")))]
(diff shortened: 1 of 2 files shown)


billy1624 (commented):
Hey @jimmycuadra, thanks for the report and the PR!!

Yes, this is an oversight. And I want to prevent this from happening again by adding test cases and leave a note for all other developers. Please check and merge https://github.com/jimmycuadra/sea-orm/pull/1.

Also, I will address the serialization of `time` types as `serde_json::Value` at https://github.com/SeaQL/sea-orm/pull/1042.

tyt2y3:
Thank you

Issue #945, opened 2022-08-08
[SeaORM] Support `time` crate on SQLite


billy1624:
Native support of `time` crate has been added in PR https://github.com/launchbadge/sqlx/pull/1865 and released as part of SQLx [0.6.1](https://github.com/launchbadge/sqlx/blob/main/CHANGELOG.md).

Now we could bind `time` on to SQLite query statement and get `time` from SQLite directly.

anshulxyz:
Hi @billy1624 I would like to work on this.
I have small experience of working with SQLite and DateTime
- https://github.com/SeaQL/sea-orm/pull/910
- https://github.com/SeaQL/sea-query/pull/400

billy1624:
Hey @anshulxyz, thanks for taking the initiative! Please go ahead :)

billy1624:
Hey @anshulxyz, I'd like to pick up this issue if you haven't start working on it yet :)

anshulxyz:
@billy1624 okay, go ahead! :D

PR #1042, opened 2022-09-15
Serialize `time` types as `serde_json::Value`

billy1624:
## PR Info

- In response to https://github.com/SeaQL/sea-orm/pull/1041#issue-1373972046
  > I also noticed that the `time` types are not accounted for in `src/query/json.rs` while the `chrono` types are, which I assume is also an oversight. However, I don't have a need for that at this point and the fix for that seemed less trivial, so I'm just bringing it to your attention.

- Dependencies:
  - https://github.com/SeaQL/sea-orm/pull/1041

## Adds

- [x] `impl FromQueryResult for JsonValue` now serialize `time` types on MySQL and PostgreSQL results. For SQLite, its support for this should be added as part of https://github.com/SeaQL/sea-orm/issues/945

### DIFF
--- a/tests/time_crate_tests.rs
+++ b/tests/time_crate_tests.rs
@@ -1,6 +1,7 @@
 pub mod common;
 pub use common::{features::*, setup::*, TestContext};
 use sea_orm::{entity::prelude::*, DatabaseConnection, IntoActiveModel};
+use serde_json::json;
 use time::macros::{date, time};
 
 #[sea_orm_macros::test]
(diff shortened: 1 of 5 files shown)


tests/time_crate_tests.rs
```diff
@@ -42,5 +43,31 @@ pub async fn create_transaction_log(db: &DatabaseConnection) -> Result<(), DbErr
...
+    let json = TransactionLog::find().into_json().one(db).await?.unwrap();
+
+    #[cfg(feature = "sqlx-postgres")]
+    assert_eq!(
+        json,
+        json!({
+            "id": 1,
+            "date": "2022-03-13",
+            "time": "16:24:00",
+            "date_time": "2022-03-13T16:24:00",
```
tyt2y3, line 55: I suspect we'd have the sub-second component?

tests/type_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+pub fn it_impl_into_active_value<T: IntoActiveValue<V>, V: Into<Value>>() {}
+
+pub fn it_impl_try_getable<T: TryGetable>() {}
+
+pub fn it_impl_try_from_u64<T: TryFromU64>() {}
+
+#[allow(unused_macros)]
+macro_rules! it_impl_traits {
```
tyt2y3, line 24: This is smart

tyt2y3 (approved):

Issue #1047, opened 2022-09-17
Support to Programmatically Generate Entities


EstebanBorai:
## Motivation

It would be great to be able to generate entities similar to how [migrations can be run
programmatically][1]. As of today we have to [install the sea-orm-cli][2] manually in order
to have entities generated:

```bash
sea-orm-cli generate entity -h
```

This would improve DX by using a single command (`cargo run`) when running a solution
using Sea ORM on development.

[1]: https://www.sea-ql.org/SeaORM/docs/migration/running-migration/#migrating-programmatically
[2]: https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/

## Proposed Solutions

Having an API similar to the one using by `Migrator` could be a first approach.

```rust
use sea_orm::entity::Generator;

Generator::generate_entities()
```

I think we could reuse the code from the Sea ORM Cli similar to what we do for
migrations nowadays.

## Additional Information

My goal is to improve DX by allowing the user to provide custom scripts by taking advantage
of the CLI as an API on runtime.

frederikhors:
I can't believe you're doing the same I'm! :smile:

I opened https://github.com/SeaQL/sea-orm/issues/1046 just before you!

This is my code:

```rust
use sea_orm_cli::{run_generate_command, DateTimeCrate};

pub async fn gen() {
    run_generate_command(
        sea_orm_cli::GenerateSubcommands::Entity {
            database_url: "...database_url here...",
            output_dir: "./entities".to_string(),
            compact_format: false,
            expanded_format: false,
            include_hidden_tables: false,
            tables: None,
            ignore_tables: vec!["".to_string()],
            max_connections: 1,
            database_schema: "public".to_string(),
            with_serde: "none".to_string(),
            with_copy_enums: false,
            date_time_crate: DateTimeCrate::Time,
        },
        true,
    )
    .await
    .unwrap();
}
```

EstebanBorai:
> I can't believe you're doing the same I'm! 😄
> 
> I opened #1046 just before you!
> 
> This is my code:
> 
> ```rust
> use sea_orm_cli::{run_generate_command, DateTimeCrate};
> 
> pub async fn gen() {
>     run_generate_command(
>         sea_orm_cli::GenerateSubcommands::Entity {
>             database_url: "...database_url here...",
>             output_dir: "./entities".to_string(),
>             compact_format: false,
>             expanded_format: false,
>             include_hidden_tables: false,
>             tables: None,
[...]

Issue #1046, opened 2022-09-17
[FR] error[E0063]: missing fields `compact_format`, `database_schema`, `date_time_crate` and 7 other fields in initializer of `GenerateSubcommands`


frederikhors:
I'm trying to use `sea-orm-cli` (0.9.2) as library with the below code but I don't know how to initialize it without all the command's flags:

```rust
use sea_orm_cli::run_generate_command;

pub async fn gen() {
    run_generate_command(
        sea_orm_cli::GenerateSubcommands::Entity {
            database_url: "...connection_url...",
            output_dir: "./entities".to_string(),
            // compact_format: (),
            // expanded_format: (),
            // include_hidden_tables: (),
            // tables: (),
            // ignore_tables: (),
            // max_connections: (),
            // database_schema: (),
            // with_serde: (),
            // with_copy_enums: (),
            // date_time_crate: (),
        },
        true,
    )
    .await;
}
```

The error is:

```
error[E0063]: missing fields `compact_format`, `database_schema`, `date_time_crate` and 7 other fields in initializer of `GenerateSubcommands`
 --> src\seaorm.rs:6:9
  |
6 |         sea_orm_cli::GenerateSubcommands::Entity {
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `compact_format`, `database_schema`, `date_time_crate` and 7 other fields
```

How to fix this?

Should we create a `new()` fn?

Can we `derive(Default)`?

billy1624:
Hey @frederikhors, what's up :)

You can find the defaults here, search "default_value = ":
- https://github.com/SeaQL/sea-orm/blob/4e51b8837a7bcea418001bcab854613a5a61c525/sea-orm-cli/src/cli.rs#L84-L191

But I believe you have figured that out already hahaa
- https://github.com/SeaQL/sea-orm/issues/1047#issuecomment-1250139862

---

> Can we `derive(Default)`?

I don't think we should derive default at here. I'm not sure what's the end goal or what you want to achieve using the sea-orm-cli programmatically.

I think it's the best to program your own logic for the code generation part instead of relying on the sea-orm-cli to do it. That means you should depends directly on sea-orm-codegen like this:
- https://github.com/SeaQL/seaography/blob/003770beb6f085dd711019b426aedf501a10ab5a/generator/src/sea_orm_codegen.rs#L9-L19

frederikhors:
[...]

PR #1054, opened 2022-09-21
[cli] generate entity programmatically

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1046
- Closes https://github.com/SeaQL/sea-orm/issues/1047

## Adds

- [x] Refactor `sea-orm-cli` to allow generate entity programmatically. The new API with an entry point of `GenerateEntity::new(database_url)` follow by a bunch of set methods to customize the generation behaviour.

### DIFF
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -45,155 +45,266 @@ pub async fn run_generate_command(
                     .try_init();
             }
 
-            // The database should be a valid URL that can be parsed
-            // protocol://username:password@host/database_name
-            let url = Url::parse(&database_url)?;
-
-            // Make sure we have all the required url components
-            //
-            // Missing scheme will have been caught by the Url::parse() call
-            // above
-            let is_sqlite = url.scheme() == "sqlite";
-
-            let tables = match tables {
-                Some(t) => t,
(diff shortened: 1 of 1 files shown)


EstebanBorai:
Awesome @billy1624! Thanks a lot! Will be checking closely later today.

JonasCir:
Hey thank you so much for tackling this so fast! 

One thing that I came up with is that this code (i.e., `GenerateEntity`)  maybe could be better located in `sea-orm-codegen` and not the CLI. 

Because for me as a developer using it 
```rust
use sea_orm_codegen::GenerateEntity;

GenerateEntity::new(database_url)
                .execute()
                .await?;
```
reads a little bit cleaner/better than `use sea_orm_cli::GenerateEntity`.

But consider this bike-shedding, I'm completely happy with the solution!

JonasCir:
On second thought, it is perfectly fine as we just start the CLI programmatically.

EstebanBorai (commented):
Looking good so far @billy1624!

JonasCir:
Hey just one quick note:
`run_migrate_command` takes the sub-command as input while this uses a build style pattern. I do not have a strong preference for either style, however, it might be a good idea to stick to the same pattern in the re-exported methods.

PR #1071, opened 2022-09-25
feat: support to Rocket-Okapi

### LINKED ISSUE #1070: Support to rocket-okapi
## Motivation

[rocket-okapi](https://github.com/GREsau/okapi) is a library that implements openapi (AKA Swagger). It could be very useful to allow sea-orm to work with this library since having swagger as part of our daily development is a must.

## Proposed Solutions
Implement ```OpenApiFromRequest``` for Connection

## Additional Information
[rocket-okapi](https://github.com/GREsau/okapi)

eum602:
## PR Info
This PR implements ```OpenApiFromRequest``` for Connection. By doing this we can use rocket-okapi, a library that implements OpenApi 3.0, in conjunction with SeaOrm.

- Closes #1070 

## Adds
- Support to rocket-okapi


HigherOrderLogic:
I think this should be a feature cause not everyone uses OpenAPI.

eum602:
> I think this should be a feature cause not everyone uses OpenAPI.

I understand you don't agree @HigherOrderLogic. I agree that not everyone uses OpenApi, but I consider supporting this feature will help increasing adoption of this project (sea-orm) as well as we will enrich  integration with rocket tools. As developers we always want to use libraries that help us as much as possible and that are versatile enough to work with other libraries. Other rust project focused on database drivers have already integrated with that tool so maybe we can just consider doing this addition since this will not affect other features but instead will add one additional point in favor of using sea-orm.

billy1624 (commented):
Hey @eum602, Welcome and thanks for contributing!! First, I'm not familiar with OpenAPI. But yeah, It's a nice addition to SeaORM ecosystem. I think we need demo the usage of it in an example. Perhaps [`rocket-example`](https://github.com/SeaQL/sea-orm/tree/master/examples/rocket_example)?

billy1624:
Btw... I wonder do we have OpenAPI support for other common Rust web API? Such as actix, poem, axum ...etc.

HigherOrderLogic:
> Btw... I wonder do we have OpenAPI support for other common Rust web API? Such as actix, poem, axum ...etc.

There's [Utoipa](https://github.com/juhaku/utoipa)

billy1624:
Thanks! @HigherOrderLogic
I saw `utoipa` can be integrated with Rocket as well. How well is it? Compared to `okapi`?

eum602:
Utoipa looks good also. I like the way how they define the different responses in the controller, seems also there is more people involved in this project. On the other hand it seems that rocket_okapi supports also another format which is [rapidoc](https://rapidocweb.com/). I can add an example of using rocket_okapi maybe in a different example with rocket. Or maybe we can agree first which library to use (or both)

eum602:
@billy1624 @HigherOrderLogic ... just added some examples and screenshots to show how the integration works at the end.

examples/rocket_example/api/Cargo.toml
```diff
@@ -21,7 +21,18 @@ serde_json = { version = "^1" }
...
 migration = { path = "../migration" }
 tokio = "1.20.0"
+serde = "1.0"
+dto = { path = "../dto" }
 
 [dependencies.sea-orm-rocket]
 path = "../../../sea-orm-rocket/lib" # remove this line in your own project and use the git line
 # git = "https://github.com/SeaQL/sea-orm"
+
+[dependencies.rocket_okapi]
```
tyt2y3, line 31: I think @HigherOrderLogic means that this should be an optional cargo feature.

sea-orm-rocket/lib/src/database.rs
```diff
@@ -205,6 +208,16 @@ impl<'a, D: Database> Connection<'a, D> {
     }
 }
 
+impl<'r, D: Database> OpenApiFromRequest<'r> for Connection<'r, D> {
```
tyt2y3, line 211: As such, this block should be feature guarded.

tyt2y3:
Thank you for your contribution all-in-all.

sea-orm-rocket/lib/Cargo.toml
```diff
@@ -24,3 +24,7 @@ version = "0.5.0-rc.1"
 version = "0.5.0-rc.1"
 default-features = false
 features = ["json"]
+
+[dependencies.rocket_okapi]
```
tyt2y3, line 28: I think @HigherOrderLogic means that this should be an optional cargo feature.

examples/rocket_example/api/src/lib.rs
```diff
@@ -145,19 +156,107 @@ async fn run_migrations(rocket: Rocket<Build>) -> fairing::Result {
...
-    rocket::build()
+    let mut building_rocket = rocket::build()
         .attach(Db::init())
         .attach(AdHoc::try_on_ignite("Migrations", run_migrations))
         .mount("/", FileServer::from(relative!("/static")))
         .mount(
             "/",
             routes![new, create, delete, destroy, list, edit, update],
         )
+        .mount(
```
tyt2y3, line 167: Actually, I think it might make sense we make a new example instead of piggy-backing the original example? may be `rocket_okapi_example`?

eum602: Sure @tyt2y3 I will fix all observations. Glad to help! :)

sea-orm-rocket/lib/Cargo.toml
eum602: @tyt2y3 I made this dependency optional.

sea-orm-rocket/lib/src/database.rs
eum602: Added a "cfg" notation.

examples/rocket_okapi_example/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "sea-orm-rocket-okapi-example"
```
eum602, line 2: @tyt2y3 I added a new folder to exemplify the usage of rocket_okapi. Hope this is enough otherwise I could make some changes you guys consider to make it better.

tyt2y3 (approved):
Thank you so much for the contribution.

eum602:
> Thank you so much for the contribution. I liked the screenshots.

Great!! :+1:

Issue #1080, opened 2022-09-28
Use of FromJsonQueryResult will fail unless sea_query module is explicitly imported


nickb937:
## Description

One reference to `sea_query` is not prefixed with `sea_orm::`

## Steps to Reproduce

```
use sea_orm::FromJsonQueryResult
use serde::{Serialize, Deserialize}

#[derive(Serialize, Deserialize, FromJsonQueryResult)]
struct Foo {
  bar: i64
}
```

### Expected Behavior

No compile error

### Actual Behavior

```
error[E0433]: failed to resolve: use of undeclared crate or module `sea_query`
  --> graphql-service/src/graphql/patient/list_appointments.rs:84:34
   |
84 | #[derive(Deserialize, Serialize, FromJsonQueryResult, SimpleObject)]
   |                                  ^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `sea_query`
   |
   = note: this error originates in the derive macro `FromJsonQueryResult` (in Nightly builds, run with -Z macro-backtrace for more info)
```

## Versions

0.9+

PR #1081, opened 2022-09-28
FromJsonQueryResult: add missing module prefix to sea_query

nickb937:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1080

## Fixes

- Code generated from use of `FromJsonQueryResult` references `sea_query` without prefixing it with `sea_orm::`, requiring an explicit module import of `sea_orm::sea_query` in the user module


tyt2y3:
Nice catch, thank you

PR #1100, opened 2022-10-09
Update README.md

Sylk:
Simple verbiage change in the Contribution section. It makes it seems less jank.


tyt2y3:
Interesting, thanks!

PR #1110, opened 2022-10-11
Introducing the error UniqueConstraintViolation as a separate kind. (Recovered)

mohs8421:
Interpreting errors now is usually no longer independent of the database, so the error handling functions in sqlx_common.rs have to be split up. Since the interpretation depends on the backend, I provided a method there, which redirects to the driver implementations. I dropped the mentioned functions in sqlx_common, to avoid anyone relying on these, as that would no longer be valid without knowing the db context.

This PR continues https://github.com/SeaQL/sea-orm/pull/1040
changes from other people might still be missing here.

### DIFF
--- a/src/driver/sqlx_common.rs
+++ b/src/driver/sqlx_common.rs
@@ -1,15 +1,5 @@
 use crate::{DbErr, RuntimeErr};
 
-/// Converts an [sqlx::error] execution error to a [DbErr]
-pub fn sqlx_error_to_exec_err(err: sqlx::Error) -> DbErr {
-    DbErr::Exec(RuntimeErr::SqlxError(err))
-}
-
-/// Converts an [sqlx::error] query error to a [DbErr]
-pub fn sqlx_error_to_query_err(err: sqlx::Error) -> DbErr {
-    DbErr::Query(RuntimeErr::SqlxError(err))
-}
-
 /// Converts an [sqlx::error] connection error to a [DbErr]
 pub fn sqlx_error_to_conn_err(err: sqlx::Error) -> DbErr {
     DbErr::Conn(RuntimeErr::SqlxError(err))
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "mock")]
 use crate::debug_print;
-use crate::{DbErr, SelectGetableValue, SelectorRaw, Statement};
+use crate::{DbBackend, DbErr, SelectGetableValue, SelectorRaw, Statement};
 use std::fmt;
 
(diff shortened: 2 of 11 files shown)


mohs8421:
@billy1624 any update on the further process regarding this?

src/executor/query.rs
```diff
@@ -108,21 +110,21 @@ macro_rules! try_getable_all {
                     QueryResultRow::SqlxMySql(row) => {
                         use sqlx::Row;
                         row.try_get::<Option<$type>, _>(column.as_str())
-                            .map_err(|e| TryGetError::DbErr(crate::sqlx_error_to_query_err(e)))
+                            .map_err(|e| TryGetError::DbErr(DbBackend::MySql.map_query_err(e)))
```
billy1624, line 113: We don't need to map the error into specific DB error at `src/executor/query.rs` because the query has been executed. We are parsing the result here. We can keep it as `crate::sqlx_error_to_query_err(e)`.

billy1624 (commented):
Hey @mohs8421, sorry for the long delay!

src/driver/sqlx_common.rs
```diff
@@ -1,15 +1,5 @@
...
 
-/// Converts an [sqlx::error] execution error to a [DbErr]
-pub fn sqlx_error_to_exec_err(err: sqlx::Error) -> DbErr {
-    DbErr::Exec(RuntimeErr::SqlxError(err))
-}
-
-/// Converts an [sqlx::error] query error to a [DbErr]
-pub fn sqlx_error_to_query_err(err: sqlx::Error) -> DbErr {
-    DbErr::Query(RuntimeErr::SqlxError(err))
-}
```
billy1624, line 11: Personally prefer keeping both method and take a `DbBackend` for converting the error to DB specific error

mohs8421: My point in removing these from the common space was, that the implementations where no longer common, but separate for each backend. Hence it has to be in relation to the backend to see how an error has to be read. Since these methods where mostly wrappers before, I didn't see a loss in removing them.

tyt2y3:
Sorry but I think I have detailed my plan/thought on fixing this issue somewhere and thus this PR either needs a rewrite or to be closed

mohs8421:
> Sorry but I think I have detailed my plan/thought on fixing this issue [here](https://github.com/SeaQL/sea-orm/pull/1040#issuecomment-1272536790) and thus this PR either needs a rewrite or to be closed

I will look into this, but I need time, having other stuff on my workstack for now.

tyt2y3:
Closed in favour of #1707

PR #1119, opened 2022-10-14
Make the PostgreSQL pool accessible

Razican:
## PR Info

This PR makes the SQLx PostgreSQL pool accessible in [`SqlxPostgresPoolConnection`](https://docs.rs/sea-orm/0.9.3/sea_orm/struct.SqlxPostgresPoolConnection.html). This allows using this connection to execute arbitrary SQLx queries if needed, and to optimize the conversion to custom, runtime-only types from query results.

## Adds

- [x] Adds a new function, `SqlxPostgresPoolConnection::pool()`, that returns a reference to the internal [`sqlx::PgPool`](https://docs.rs/sqlx/0.6.2/sqlx/type.PgPool.html).

Let me know if you want me to add this for other backends. I just added it for PostgreSQL because I needed it for a project I'm working on.

## Breaking Changes

This is not a breaking change, since the API receives a new method, and doesn't delete former methods.

### DIFF
--- a/src/driver/sqlx_postgres.rs
+++ b/src/driver/sqlx_postgres.rs
@@ -202,7 +202,7 @@ impl SqlxPostgresPoolConnection {
     }
 }
 
-impl SqlxPostgresConnector {
+impl SqlxPostgresPoolConnection {
     /// Retrieves the internal SQLx PostgreSQL connection pool.
     pub fn pool(&self) -> &PgPool {
         &self.pool
--- a/src/driver/sqlx_postgres.rs
+++ b/src/driver/sqlx_postgres.rs
@@ -202,6 +202,13 @@ impl SqlxPostgresPoolConnection {
     }
 }
 
+impl SqlxPostgresConnector {
+    /// Retrieves the internal SQLx PostgreSQL connection pool.
+    pub fn pool(&self) -> &PgPool {
+        &self.pool
+    }
+}
+
 impl From<PgRow> for QueryResult {
     fn from(row: PgRow) -> QueryResult {
         QueryResult {


tyt2y3:
I think we can expose it via sea-orm-migration, but should not expose it in SeaORM itself.

Razican:
> I think we can expose it via sea-orm-migration, but should not expose it in SeaORM itself.
> Or alternatively it should give up the ownership i.e. `fn (self)`
> Note that you can always clone the pool before constructing the `Database` object.

Well, when you're using Actix, for example, the `Database` object is managed by Actix itself. You can decide and use a `PgPool` as the managed object, but then you cannot use `sea_orm`, since you cannot construct a `Database` from a `PgPool`. Also, that statement about being able to clone the pool before constructing the `Database` object doesn't seem to hold, since it's not possible to use a `PgPool` to construct a `Database` object, you can only use `ConnectOptions`.

Also, I don't understand what does this have to do with migrations. It's useful to use plain SQLx in a database query if the ORM doesn't provide everything you need (for example, dynamic custom types for a dynamic schema at runtime).

I might be missing something, because I've only been working with `sea_orm` for a short time, but what is the issue of giving access to the internal structure here?

billy1624:
Hey @Razican, you can construct `DbConn` from `PgPool`, https://docs.rs/sea-orm/latest/sea_orm/?search=from_sqlx_.

Razican:
> Hey @Razican, you can construct `DbConn` from `PgPool`, [`SqlxPostgresConnector::from_sqlx_postgres_pool`](https://docs.rs/sea-orm/latest/sea_orm/struct.SqlxPostgresConnector.html#method.from_sqlx_postgres_pool).

Indeed! Thank you, that solves my issue, by creating a `DatabaseConnection` only when needed for the ORM.

PR #1124, opened 2022-10-16
sea-orm-cli Implement derives & attributes parameters for entity generation

### LINKED ISSUE #1120: sea-orm-cli [Feature Request] Custom derives and attributes for entity generation
I just finished setting up my first sea-orm project and really like it so far. I am requesting this feature because I want to derive traits on models generated by the cli without needing to do it manually. Additionally, users would need an `attribute` or `attr` (`--attribute ts(export)`) argument to add extra attributes to the struct. For example, if you run

```bash
sea-orm-cli generate entity --derive ts_rs::TS --attribute ts(export)
```

You would get a model like this

```rs
#[derive(Copy, Clone, Debug, PartialEq, DeriveEntityModel, ts_rs::TS)]
#[ts(export)]
pub struct Model {
    pub title: String,
    pub author: String,
}
```

I am ready and willing to contribute this feature if it is deemed worthy.

tsar-boomba:
## PR Info

- Closes #1120 

## Adds

Ability to add bonus derive macros and attributes to generated models.

## Changes

Reusable testing utilities for codegen crate & tests for the new features added.


tsar-boomba:
Should this also add derives to active enums, realizing that they might need the traits derived too?

billy1624 (commented):
Hey @tsar-boomba, thanks for contributing!!

I have some suggested code changes: https://github.com/tsar-boomba/sea-orm/pull/1
Please check and see if it make sense to you :)

billy1624:
> Should this also add derives to active enums, realizing that they might need the traits derived too?
> 
> EDIT: I see this was changed in 0.10.0 and is no longer necessary.

I didn't catch it. Why adding custom derives to active enums is no longer necessary in 0.10.0?

tsar-boomba:
It seems that after 0.10, by default it produces this
```rs
#[sea_orm(column_type = "Custom(\"note_type\".to_owned())", nullable)]
pub r#type: Option<String>,
```
for a column with an enum type (Postgres). Before it would make a separate file called active_enum I believe.

tsar-boomba:
One other thing, the fully qualified name is being mangled when parsed into `TokenStream`.

Before parsing to `TokenStream`:
`"ts_rs::TS"`

After parsing to `TokenStream` (spaces added before and after "::"):
`ts_rs :: TS`

Any idea what is going on with this? The derives work as expected, so maybe this doesn't matter since they're generated files anyways.

billy1624:
> It seems that after 0.10, by default it produces this
> 
> ```rust
> #[sea_orm(column_type = "Custom(\"note_type\".to_owned())", nullable)]
> pub r#type: Option<String>,
> ```
> 
> for a column with an enum type (Postgres). Before it would make a separate file called active_enum I believe.

Oh, please install sea-orm-cli 0.10.1, this has been fixed on latest version.

tsar-boomba:
Alright, I've updated and see the enum is there. So should `--derives` be applied to enums or should there be a `--active-enum-derives` for only active enums, I'd prefer the latter. Also seeing "use of undeclared crate or module sea_query" error on the `DeriveActiveEnum` for the enum.

billy1624 (approved):
Hey @tsar-boomba, sorry for the delay. I'd like to refactor it a bit. I'll merge it into a local branch before it was committed into master. Thanks again for the contributions!!

Issue #962, opened 2022-08-11
More context for ActiveModelBehaviour


mohs8421:
## Motivation
When designing entities it is pretty common to write constraints into the datalayer. Sea-orm provides the `ActiveModelBehaviour` to allow users to fill in these pieces.
However, it is usually necessary to do some further analysis in these cases, like dropping some queries to check for further conditions to allow or deny a certain change. It might also be important to know the state of a record before the change that is going to be made, for example to implement a proper state transition, that might only be valid if some more data has been added to the record.

## Proposed Solutions
It would be best if the `ActiveModelBehaviour`'s signature would change to be like follows: 
```rust
pub trait ActiveModelBehavior: ActiveModelTrait {
    /// Create a new ActiveModel with default values. Also used by `Default::default()`.
    fn new() -> Self {
        <Self as ActiveModelTrait>::default()
    }

    /// Will be called before saving
    // Note, that the former insert parameter is replaced by the option to the old_record, which fullfills the same purpose (and more), but is nevertheless a breaking change.
    fn before_save(self, old_record: Option<Model>, connection: &DatabaseConnection) -> Result<Self, DbErr> {
        Ok(self)
    }

    /// Will be called after saving
    fn after_save(
        model: <Self::Entity as EntityTrait>::Model,
        old_record: Option<Model>,
        connection: &DatabaseConnection,
    ) -> Result<<Self::Entity as EntityTrait>::Model, DbErr> {
        Ok(model)
    }

    /// Will be called before deleting
    fn before_delete(self, old_record: Option<Model>, connection: &DatabaseConnection) -> Result<Self, DbErr> {
        Ok(self)
    }

    /// Will be called after deleting
    fn after_delete(self, old_record: Option<Model>, connection: &DatabaseConnection) -> Result<Self, DbErr> {
        Ok(self)
    }
}
```
It might also be convenient to reduce the amount of parameters and pack them into a parameter-object struct like this:
```rust
pub struct SaveContext {
    old_record: …
    connection: …
}
```

billy1624:
Hey @mohs8421, thanks for the feature request!

Passing `&DatabaseConnection` into `*_save` and `*_delete` is a straightforward and easy change.

However, providing the old record / model isn't a trivial task. Currently, we don't store old value in ActiveModel.

One approach would be adding a `Changed` variant to `ActiveValue`, which keeps a tuple representing the old and new value of a field in `ActiveModel`
[...]

PR #1145, opened 2022-10-24
feat: expose database connection to `ActiveModelBehaviour`'s methods

teenjuna:
## PR Info

This PR partly solves #962, making it possible to mimic database triggers by automatically executing database queries based on other database queries.

I only changed signatures in one existing test and didn't extend it: this change only passes database connection one level deeper, so we can rely on type system to ensure it's working correctly (correct me if I'm wrong).

## Adds

Adds  `&impl ConnectionTrait` to arguments of `ActiveModelBehaviour`'s methods

## Breaking Changes

This is a breaking change since all implementations of `ActiveModelBehaviour` must be updated. The fix is straightforward though.

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -278,11 +278,11 @@ pub trait ActiveModelTrait: Clone + Debug {
         Self: ActiveModelBehavior + 'a,
         C: ConnectionTrait,
     {
-        let am = ActiveModelBehavior::before_save(self, true)?;
+        let am = ActiveModelBehavior::before_save(self, db, true)?;
         let model = <Self::Entity as EntityTrait>::insert(am)
             .exec_with_returning(db)
             .await?;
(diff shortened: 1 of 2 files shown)


teenjuna:
@billy1624 maybe we could add this to v0.11?

tyt2y3:
Sorry if I am too late to the discussion.

1) prefer the `<C: ConnectionTrait>` generic syntax
2) the method signature has to be changed to async, and it has the potential to create an infinite recursion. `async_trait` Boxed it so it might compile

billy1624 (commented):
Hey @teenjuna, sorry for the delay and thanks for the contributions. Yes, we could include this in 0.11 release.

I'm going to merge this into a local branch where I'd make some changes to it.

teenjuna:
Hey @billy1624, sorry I didn't finish the work. Got very busy...

PR #1149, opened 2022-10-24
use GAT to elide StreamTrait lifetime

nappa85:
Hello guys, I'm back!
Thursday Rust 1.65.0 will be stabilized, and with it GATs will land on stable.
I've seen only this morning you've release a release candidate for 0.10.0, but this work is portable to the new version for sure.
Long story short, using GATs we can remove the lifetime from StreamTrait and move it internally.
To be honest, that lifetime has given me more than one headache, I think this is a big usability improvement.
It's still a work in progress, I'm still testing, I've already seen some corner cases and I'll work on it tomorrow, but I wanted to share with you this improvement so you can start thinking about landing an improvement that needs next stable compiler to work
Best regards

edit: I wrote TAIT instead of GAT, I need some rest...

### DIFF
--- a/src/database/connection.rs
+++ b/src/database/connection.rs
@@ -34,15 +34,17 @@ pub trait ConnectionTrait: Sync {
 }
 
 /// Stream query results
-pub trait StreamTrait<'a>: Send + Sync {
+pub trait StreamTrait: Send + Sync {
     /// Create a stream for the [QueryResult]
-    type Stream: Stream<Item = Result<QueryResult, DbErr>> + Send;
+    type Stream<'a>: Stream<Item = Result<QueryResult, DbErr>> + Send
+    where
+        Self: 'a;
 
     /// Execute a [Statement] and return a stream of results
-    fn stream(
+    fn stream<'a>(
         &'a self,
         stmt: Statement,
-    ) -> Pin<Box<dyn Future<Output = Result<Self::Stream, DbErr>> + 'a + Send>>;
+    ) -> Pin<Box<dyn Future<Output = Result<Self::Stream<'a>, DbErr>> + 'a + Send>>;
 }
 
 /// Spawn database transaction
--- a/src/database/db_connection.rs
+++ b/src/database/db_connection.rs
@@ -159,15 +159,15 @@ impl ConnectionTrait for DatabaseConnection {
 }
 
 #[async_trait::async_trait]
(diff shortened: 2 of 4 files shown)


tyt2y3:
Welcome back!

Yes I think GAT can be used to replace the use of `async_trait` and other things.
As I have said, there are three hard things in Rust: lifetime, orphan rule and generic associated types.
And this library is (going to) tackle all three.

Awesome! Let me know the progress or problems.

tyt2y3:
> How do you think to use GAT to replace async_trait

I think I learnt this technique here: https://www.sobyte.net/post/2022-04/rust-gat-async-trait/

> Are you planning to write concrete futures for all trait return types?

No concrete plan for now

nappa85:
> > How do you think to use GAT to replace async_trait
> 
> I think I learnt this technique here: https://www.sobyte.net/post/2022-04/rust-gat-async-trait/
Yes, this uses both GATs and TAITs
GAT allow you to do
```trait Foo {
  type Bar<'a>: AnotherTrait<'a>;
}```
but when you impl the trait you still need to use a concrete type, the only difference is that it can have a lifetime that isn't global to the trait himself
with TAIT you can use `impl AnotherTrait<'a>` for your type Bar<'a>

**Comment by tyt2y3** [comment]
Umm... I think the original concern is that `async_trait` basically "Box the Future", and GAT removes that.
It's still pretty unergonomic, but we can hide the details with yet another macro.

**Comment by nappa85** [comment]
> Umm... I think the original concern is that `async_trait` basically "Box the Future", and GAT removes that. It's still pretty unergonomic, but we can probably hide the details with yet another macro. It might make sense to use this technique on TransactionTrait, but this is probably another task.

Yes, what I'm trying to say is that to replace async_trait you need both GAT and TAIT, like in your article
```
#![feature(generic_associated_types)] //GAT
#![feature(type_alias_impl_trait)] //TAIT
```

or else you need to create a concrete future type for every return, like tokio does internally

tyt2y3:
Ah sad I forgot TAIT is still unstable

nappa85:
Thinking about the breaking change, maybe you'll prefer merging this only on 0.10.0 to not introduce breaking changes on 0.9.x

tyt2y3:
I think we can close this?

tyt2y3:
Since this requires newer Rust version, I don't think we can backport this

Issue #1156, opened 2022-10-26
`Eq` is not implemented for `Vec<f64>` error when using "postgress-array" feature 


tjallingt:
## Description

The "postgres-array" feature is not taken into account when generating `Eq` derives.
This causes the CLI to derive `Eq` for models containing `Vec<f64>` which will result in the following compilation error:

```
error[E0277]: the trait bound `f64: std::cmp::Eq` is not satisfied
   --> app\src\entities\sensor.rs:17:5
    |
6   | #[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
    |                                                      -- in this derive macro expansion
...
17  |     pub parameter_value: Vec<f64>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::cmp::Eq` is not implemented for `f64`
```

## Steps to Reproduce

1. create the following table in a postgres database:

```sql
create table sensor (
    id uuid primary key default gen_random_uuid(),
    parameter_value double precision[] not null
);
```

2. enable the "postgres-array" feature in Cargo.toml

```toml
sea-orm = { version = "0.10", features = [
  "sqlx-postgres",
  "runtime-tokio-native-tls",
  "macros",
  "postgres-array",
] }
```

3. Run `sea generate entity` to generate the entities
4. Check/compile the project

### Expected Behavior

This should compile, no `Eq` derive on the "sensor" Model.

### Actual Behavior

This does not compile, there is a `Eq` derive on the "sensor" Model.

### Reproduces How Often

Always

## Versions

sea-orm 0.10.0
sea-orm-cli 0.10.0

## Additional Information

The `Eq` derive was added in this PR but the logic does not take into account floats inside `Vec`s https://github.com/SeaQL/sea-orm/pull/988

billy1624:
Hey @tjallingt, thanks for catching this!! I just fix it on this PR
- https://github.com/SeaQL/sea-orm/pull/1158

We'll release a new patch shortly :)

tjallingt:
Thank you so much for the quick fix, love the project so far!

billy1624:
Thanks for the support and love :D

PR #1158, opened 2022-10-26
Fix - vector of float & double does not derive Eq

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1156

## Fixes

- [x] [cli] generate model with `Vec<f32>` / `Vec<f64>` should not derive `Eq` on the model struct


sea-orm-codegen/src/entity/writer.rs
```diff
@@ -1182,7 +1213,7 @@ mod tests {
     #[test]
     fn test_gen_expanded_code_blocks() -> io::Result<()> {
         let entities = setup();
-        const ENTITY_FILES: [&str; 9] = [
```
tyt2y3, line 1185: try [; _]

billy1624: I wish that works. Seems it's coming but not on stable Rust just yet
- https://github.com/rust-lang/rust/issues/85077

PR #1161, opened 2022-10-27
use GAT to elide StreamTrait lifetime

nappa85:
Clone of PR #1149, but on master branch


tyt2y3:
```
generic associated types are unstable
```

umm @billy1624 aren't we using the latest stable Rust?

am I missing something here - it some bit of it still in unstable realm?

nappa85:
> ```
> generic associated types are unstable
> ```
> 
> umm @billy1624 aren't we using the latest stable Rust?
> 
> am I missing something here - is some bit of it still in unstable realm?

GATs should stabilize today

nappa85:
Well, I'm wrong again, 1.65.0 exits in a week

Il gio 27 ott 2022, 19:11 Chris Tsang ***@***.***> ha scritto:

tyt2y3:
Yeah I have been using it without the `feature` tag for a while (in nightly). Let's merge this after 1.65

tyt2y3:
Tried again! Cool, I think the clippy error is unrelated to this PR and we should fix it on master @billy1624 
Other than that, I think this can be merged.

Actually, I have no clue about Rust's release schedule.

nappa85:
> Tried again! Cool, I think the clippy error is unrelated to this PR and we should fix it on master @billy1624 Other than that, I think this can be merged.

Yes, it's not from my PR

> Actually, I have no clue about Rust's release schedule.

A new stable every 6 weeks https://www.whatrustisit.com/

tyt2y3:
One question, I think this does not break the API despite ABI change, so justifiable as a minor release. Agree?

nappa85:
Well, as far as I know, you released 0.10.0 as a release candidate, so this
change won't be a problem

Il dom 6 nov 2022, 05:57 Chris Tsang ***@***.***> ha scritto:

tyt2y3:
I released 0.10.2 anywya

tyt2y3:
Mentioning https://github.com/rust-lang/rust/pull/96709

tyt2y3:
@nappa85 Would you mind sharing a snippet that bugs you before but is solved by this change?

I am drafting the release blog post for this feature and I would like to illustrate it with an example.

I imagine you've had a generic 'stream processor' function that can accept different `DatabaseConnection`, in which the `'a` lifetime propagates to other places?

nappa85:
It wasn't a real bug, more a lifetime problem, let me explain...

I always ask for a generic connection C: ConnectionTrait and, when needed, also StreamTrait and/or TransactionTrait. This way my functions can be called on a normal connection or on a transaction, or anything else.

With old code, when you need StreamTrait, you end up needing a lifetime, e.g.
```rust
async fn do_something<'a, C>(conn: &'a C) -> Result<...>
where C: ConnectionTrait + StreamTrait<'a> {...}
```

Most of the times it worked like a charm, but sometimes it won't compile telling you that the connection is still borrowed when it really isn't, but the problem is generated by connection reference and stream sharing the same lifetime.

It was on situations like that:
```rust
async fn complex_situation<C>(conn: &C) -> Result<...>
where C: ConnectionTrait + TransactionTrait
{
  do_something(&txn).await?;
  txn.commit().await?;// here we have the error, because it tells txn is still borrowed
}
```

Unfortunately I don't have a MVP demonstrating the problem.

With GAT, the same function loses the lifetime, e.g.
```rust
async fn do_something<C>(conn: &C) -> Result<...>
where C: ConnectionTrait + StreamTrait {...}
```

This is cleaner, and the problem is solved because now the stream has his own lifetime disconnected from connection reference.

ikrivosheev:
@tyt2y3 I think we need to upgrade minimum rust version after merge GAT...

tyt2y3:
> @tyt2y3 I think we need to upgrade minimum rust version after merge GAT...

I think I did bumped all rust_version , did I miss something?

@nappa85 Thank you for the elaboration!

ikrivosheev:
> > @tyt2y3 I think we need to upgrade minimum rust version after merge GAT...
> 
> I think I did bumped all rust_version , did I miss something?
> 
> @nappa85 Thank you for the elaboration!

Yes, you did! Sorry, I don't find it in PR.

PR #1165, opened 2022-10-27
Fix transaction metrics

nappa85:
Kind of embarrassing, I don't know how it's possible I haven't noticed this before, but there wasn't the `metrics!` macro applied to transactions selects, only to execute...


tyt2y3:
Great that we found out now!

PR #1177, opened 2022-11-01
Unchanged ActiveValue as Set

billy1624:
## New Features

- [x] Converting `Unchanged` active value as `Set`

### DIFF
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -154,6 +154,13 @@ fn derive_active_model(all_fields: IntoIter<Field>) -> syn::Result<TokenStream> 
                     #(#field: sea_orm::ActiveValue::not_set()),*
                 }
             }
+
+            fn val_as_set(&mut self, c: <Self::Entity as EntityTrait>::Column) {
+                match c {
+                    #(<Self::Entity as EntityTrait>::Column::#name => self.#field.val_as_set(),)*
+                    _ => panic!("This ActiveModel does not have this field"),
+                }
+            }
         }
     ))
 }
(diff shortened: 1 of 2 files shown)


tyt2y3:
Umm I think there might be more self-explanatory method names

billy1624:
Agree! Let me brainstorm it tonight in dream.
Any other naming suggestions are welcomed!

billy1624:
How about `unchanged_value_into_set`?

billy1624:
Okay, renamed

tyt2y3:
I think we should instead only add two methods to `ActiveModel`: `iter` and `iter_mut`.
And then the user can do:

```rust
active_model.iter_mut(|(_col, &mut val)| {
    *val = match val {
        ActiveValue::Unchanged(v) => ActiveValue::Set(v),
        _ => (),
    };
});
```

billy1624:
Let user perform the conversion on their own?

billy1624:
We just provide a API to effortlessly iterate all fields in the ActiveModel.

And we should include a code snippet in the cookbook on how to update all `Unchanged` fields into `Set`

tyt2y3:
May we could have a `rechange` method on ActiveValue:

```rust
active_model.iter_mut(|(_col, &mut val)| {
    *val = val.rechange();
});
```

tyt2y3:
May be we don't have to bound the `V`

tyt2y3:
Ah I see what you mean, sad :(

billy1624:
Back the the naming problem. How about `set_dirty`?

billy1624:
Any opinion on the naming? @tyt2y3

tyt2y3:
Back to square 1, can we have `reset` and `reset_all`?

billy1624:
Renamed :D

tyt2y3 (approved):

PR #1181, opened 2022-11-01
added reexports for sqlx errors

e-rhodes:
## PR Info
 With the new `RuntimeErr` in `DbErr` it's much nicer to condition on different database errors. However, since we can't directly interact with the private `SqlxError` or the database-dependent `Mysql/Pg/Sqlite` errors, we don't get the full benefit of this change.

## New Features

- Adds reexports for `sqlx` error types so users can interact with them directly.

## Changes

- Changes existing `use` for `SqlxError` to reexport as `pub use`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,12 @@
 #[cfg(feature = "sqlx-dep")]
-use sqlx::error::Error as SqlxError;
+pub use sqlx::error::Error as SqlxError;
+#[cfg(feature = "sqlx-mysql")]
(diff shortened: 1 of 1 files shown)


tyt2y3:
Let me think about it, I think we used to have a policy to not 'leaking SQLx types'

billy1624:
I think this can be offered under `seaql-internal` feature. Just like what we did at https://github.com/SeaQL/sea-orm/pull/1297

tyt2y3:
Yes, I agree at least a feature guard is necessary.

billy1624:
Okay, I will claim the task

PR #1192, opened 2022-11-05
Make sqlx logging configurable for seaorm-rocket

jayay:
## New Features

- [x] `sea_orm_rocket::Config` was extended by the boolean field `sqlx_logging`, which defaults to `true` if not set and can be used to disable query logging to the console. This comes in handy if the query logging is not desired such as fuzzing  environments or very busy deployments. The value needs to be forwarded into the `sea_orm::ConnectOptions` struct by the user of this library, as shown in the changes to `examples/rocket_example/api/src/pool.rs`.

## Bug Fixes

- [ ] 

## Breaking Changes

- [x] This PR includes a breaking change. `sea_orm_rocket::Config` got a new `pub` field.

## Changes

- [ ]

### DIFF
--- a/examples/rocket_example/api/src/pool.rs
+++ b/examples/rocket_example/api/src/pool.rs
@@ -26,7 +26,8 @@ impl sea_orm_rocket::Pool for SeaOrmPool {
         options
             .max_connections(config.max_connections as u32)
             .min_connections(config.min_connections.unwrap_or_default())
-            .connect_timeout(Duration::from_secs(config.connect_timeout));
+            .connect_timeout(Duration::from_secs(config.connect_timeout))
+            .sqlx_logging(config.sqlx_logging);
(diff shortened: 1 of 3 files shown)


tyt2y3 (approved):

tyt2y3:
Awesome, although sadly this is a breaking change

jayay:
@tyt2y3  Oh no, I didn't pay attention to that! There's no way to implement this as a non-breaking change, or am I missing anything?  Do you think the `Config` struct could have `pub(crate)` fields in the future (and a `::new()` function), so that new fields can be added gradually?

tyt2y3:
It's okay. It's breaking because of the pub field in struct, so it's part of the design. I will release a new version of that crate and update the examples. Thanks for the contribution!

PR #1226, opened 2022-11-14
Update COMMUNITY.md

nappa85:
Adds logic-lock to crates list


billy1624 (approved):
Thanks!! @nappa85
Great to see you again :)

nappa85:
We talked about that functionality in #469, but since it's not supported by SqLite I've made an external crate
The same code should work also on PostgreSQL, but I haven't tested since it's outside my use-case
If you think it's more a core functionality, feel free to integrate it

tyt2y3:
Awesome work! I think it might worth having an `extension` concept in SeaORM (just like SeaQuery), where we can ship all database-specific features in it.

Issue #1023, opened 2022-09-08
Replace all `unwrap()` with `expect()`


billy1624:
## Motivation

We should refactor all `unwrap()` in SeaORM and use `expect()` insead. This will force us to provide a explicit message when panicking. This is far more readable than a muted panic.

baoyachi:
I wish to take this into account by the way, necessary for error segmentation :https://github.com/SeaQL/sea-orm/issues/881

tyt2y3:
Oh Nice I wanted to say this a long time ago.

But it's also good to think about whether we can avoid panicking on a case by case basis.

tyt2y3:
+1

billy1624:
Same here.

Btw, we should document all the method that throw panic, specify under what circumstance it will panic.

Deny this linter and it will tell you which methods need to be documented
https://rust-lang.github.io/rust-clippy/master/#missing_panics_doc

billy1624:
> I wish to take this into account by the way, necessary for error distinguish :#881

Hey @baoyachi, what do you mean?

I think in general, we should avoid panic. And should throw Err instead.
Did you mean that?

billy1624:
> Can we have a new label named "Motivation" with prefixes like beginner, moderate and expert ....... Would help to filter the issues

I'm going to do a cleanup on the open issues and label some beginner friendly issues as "good first issue". And convert Q&A into discussion.

baoyachi:
> > Can we have a new label named "Motivation" with prefixes like beginner, moderate and expert ....... Would help to filter the issues
> 
> I'm going to do a cleanup on the open issues and label some beginner friendly issues as "good first issue". And convert Q&A into discussion.

I agree

PR #1231, opened 2022-11-16
Cleanup panic and unwrap

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1023

## Changes

- [x] Implemented `From<DbErr> for TransactionError<E>`


src/database/db_connection.rs
```diff
@@ -207,7 +217,9 @@ impl TransactionTrait for DatabaseConnection {
             DatabaseConnection::MockDatabaseConnection(conn) => {
                 DatabaseTransaction::new_mock(Arc::clone(conn), None).await
             }
-            DatabaseConnection::Disconnected => panic!("Disconnected"),
+            DatabaseConnection::Disconnected => {
+                Err(DbErr::Conn(RuntimeErr::Internal("Disconnected".to_owned())))
```
ikrivosheev, line 221: Some thoughts. Maybe is better to store in `RuntimeErr::Internal` `Cow<'static, str>` instead `String`? And we don't need call: `to_owned`?

billy1624: Very cool! Great suggestion! I think this is actually an optimization over the use of `String`. `RuntimeErr::Internal(Cow<'static, str>)` requires no heap allocation given that it was constructed via a `&str`, i.e. `RuntimeErr::Internal("Disconnected")`.

Thoughts? @tyt2y3

billy1624: But still... we need `"xxx".into()`. Correct me if I get it wrong @ikrivosheev

tyt2y3: I don't think `Cow` is good to be part of the public API data structure. More like an internal implementation optimization.
And I think it's okay to allocate in Error paths.

billy1624:
Please review!

src/error.rs
```diff
@@ -55,6 +55,9 @@ pub enum DbErr {
     /// A migration error
     #[error("Migration Error: {0}")]
     Migration(String),
+    /// A mock error
+    #[error("Mock Error: {0}")]
+    Mock(String),
```
tyt2y3, line 60: Can we reuse `RuntimeErr::Internal`?

src/database/mock.rs
```diff
@@ -92,76 +92,83 @@ impl MockDatabase {
...
-            transaction.push(statement);
-        } else {
-            self.transaction_log.push(Transaction::one(statement));
+        match self.transaction.as_mut() {
+            Some(transaction) => transaction.push(statement),
+            None => self.transaction_log.push(Transaction::one(statement)),
         }
-        if counter < self.exec_results.len() {
-            Ok(ExecResult {
+        match self.exec_results.len() {
```
tyt2y3, line 99: the original `if else` looks better

src/database/db_connection.rs
ikrivosheev: @tyt2y3 ok

tyt2y3:
Looks good for what this PR changed.

tyt2y3:
Looks good for what this PR changes

src/database/mock.rs
```diff
@@ -92,76 +92,83 @@ impl MockDatabase {
...
-                let transaction = self.transaction.take().unwrap();
-                self.transaction_log.push(transaction.into_transaction());
+    fn rollback(&mut self) -> Result<(), DbErr> {
+        match self.transaction.as_mut() {
+            Some(transaction) => {
+                if transaction.rollback(self.db_backend) {
+                    let transaction = self
+                        .transaction
+                        .take()
+                        .expect("No open transaction to rollback");
```
ikrivosheev, line 164: Maybe is better return `Err` instead `panic`?

src/database/mock.rs
```diff
@@ -92,76 +92,83 @@ impl MockDatabase {
...
-                let transaction = self.transaction.take().unwrap();
-                self.transaction_log.push(transaction.into_transaction());
+    fn commit(&mut self) -> Result<(), DbErr> {
+        match self.transaction.as_mut() {
+            Some(transaction) => {
+                if transaction.commit(self.db_backend) {
+                    let transaction = self
+                        .transaction
+                        .take()
+                        .expect("No open transaction to commit");
```
ikrivosheev, line 145: Maybe is better return Err instead panic?

tyt2y3:
This PR mostly touches the `Connection` and `Mock` API, I think there might be other places to cleanup too?

src/database/mock.rs
billy1624: Nice catch!

ikrivosheev: @billy1624 why did you close conversation?

billy1624: Oh I for got to push

ikrivosheev (approved):

billy1624:
> This PR mostly touches the `Connection` and `Mock` API, I think there might be other places to cleanup too?

Yes, but I think we can leave them as-is:

```
20 results - 10 files

src/entity/active_model.rs:
  168              }
  169:             _ => panic!("The arity cannot be larger than 6"),
  170          }

  812              ActiveValue::Set(value) | ActiveValue::Unchanged(value) => value,
  813:             ActiveValue::NotSet => panic!("Cannot unwrap ActiveValue::NotSet"),
  814          }

  844              ActiveValue::Set(value) | ActiveValue::Unchanged(value) => value,
  845:             ActiveValue::NotSet => panic!("Cannot borrow ActiveValue::NotSet"),
  846          }

src/entity/base_entity.rs:
  270              } else {
  271:                 panic!("primary key arity mismatch");
  272              }

  274          if keys.next().is_some() {
  275:             panic!("primary key arity mismatch");
  276          }

  837              } else {
  838:                 panic!("primary key arity mismatch");
  839              }

  841          if keys.next().is_some() {
  842:             panic!("primary key arity mismatch");
  843          }

src/executor/cursor.rs:
  95                  .add(f(c1, v1)),
  96:             _ => panic!("column arity mismatch"),
  97          }

src/executor/execute.rs:
  48              ExecResultHolder::SqlxPostgres(_) => {
  49:                 panic!("Should not retrieve last_insert_id this way")
  50              }

  54                  if last_insert_rowid < 0 {
  55:                     panic!("negative last_insert_rowid")
  56                  } else {

src/query/combine.rs:
  49                              ColumnRef::Asterisk | ColumnRef::TableAsterisk(_) => {
  50:                                 panic!("cannot apply alias for Column with asterisk")
  51                              }

  58                                  ColumnRef::Asterisk | ColumnRef::TableAsterisk(_) => {
  59:                                     panic!("cannot apply alias for AsEnum with asterisk")
  60                                  }

  62                              _ => {
  63:                                 panic!("cannot apply alias for AsEnum with expr other than Column")
  64                              }
  65                          },
  66:                         _ => panic!("cannot apply alias for expr other than Column or AsEnum"),
  67                      };

src/query/delete.rs:
  115                  }
  116:                 ActiveValue::NotSet => panic!("PrimaryKey is not set"),
  117              }

src/query/helper.rs:
  585          }
  586:         _ => panic!("Owner key and foreign key mismatch"),
  587      }

src/query/insert.rs:
  136              } else if self.columns[idx] != av_has_val {
  137:                 panic!("columns mismatch");
  138              }

src/query/update.rs:
   99                  }
  100:                 ActiveValue::NotSet => panic!("PrimaryKey is not set"),
  101              }

src/schema/entity.rs:
  49      if matches!(backend, DbBackend::MySql | DbBackend::Sqlite) {
  50:         panic!("TypeCreateStatement is not supported in MySQL & SQLite");
  51      }

  59          ColumnType::Enum { name, variants } => (name.clone(), variants.clone()),
  60:         _ => panic!("Should be ColumnType::Enum"),
  61      };

```

tyt2y3:
Yeah all of them seems to be in the "Misuse of API" category

tyt2y3:
Except this one:

```
  54                  if last_insert_rowid < 0 {
  55:                     panic!("negative last_insert_rowid")
  56                  } else {
```

Which should either be unreachable or an Runtime error.

tyt2y3:
Just checked the docs, https://www.sqlite.org/c3ref/last_insert_rowid.html, I think it can only return 0

tyt2y3:
Can remove the doc `/// - SQLite last insert row id might be negative integer, in this case it will panic` too?

billy1624:
> Just checked the docs, https://www.sqlite.org/c3ref/last_insert_rowid.html, I think it can only return 0

But why SQLx method return i64? https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteQueryResult.html#method.last_insert_rowid

tyt2y3:
> But why SQLx method return i64? 

That's another question

billy1624:
It's just a wrapper around the SQLite library in C
https://github.com/launchbadge/sqlx/blob/76ae2865005cc79d56972d68d3be30805581103d/sqlx-core/src/sqlite/connection/handle.rs#L58

billy1624:
This PR should goes after all other feature updates

src/executor/execute.rs
```diff
@@ -22,12 +22,23 @@ pub(crate) enum ExecResultHolder {
     /// Holds the result of executing an operation on the Mock database
     #[cfg(feature = "mock")]
     Mock(crate::MockExecResult),
+    Disconnected,
```
tyt2y3, line 25: While I agree on adding Disconnected state to InnerConnection, I don't think it makes sense to add on ExecResultHolder. What problem does it solve here?

src/executor/query.rs
```diff
@@ -19,6 +19,13 @@ pub(crate) enum QueryResultRow {
     SqlxSqlite(sqlx::sqlite::SqliteRow),
     #[cfg(feature = "mock")]
     Mock(crate::MockRow),
+    Disconnected,
```
tyt2y3, line 22: And QueryResultRow too

examples/actix3_example/api/src/lib.rs
```diff
@@ -109,7 +109,7 @@ async fn edit(data: web::Data<AppState>, id: web::Path<i32>) -> Result<HttpRespo
     let post: post::Model = Query::find_post_by_id(conn, id)
         .await
         .expect("could not find post")
-        .unwrap_or_else(|| panic!("could not find post with id {}", id));
+        .unwrap_or_else(|| panic!("could not find post with id {id}"));
```
tyt2y3, line 112: These must belong to the "cleanup clippy" PR

src/executor/execute.rs
billy1624: Not really. I originally want to have a placeholder to avoid the use of

```rs
#[allow(unreachable_patterns)]
_ => unreachable!(),
```

src/executor/query.rs
billy1624: Same here.

> I originally want to have a placeholder to avoid the use of
> 
> ```rust
> #[allow(unreachable_patterns)]
> _ => unreachable!(),
> ```

examples/actix3_example/api/src/lib.rs
billy1624: Right.

billy1624: https://github.com/SeaQL/sea-orm/pull/1436

src/database/db_connection.rs
```diff
@@ -60,9 +67,10 @@ pub(crate) enum InnerConnection {
     Sqlite(PoolConnection<sqlx::Sqlite>),
     #[cfg(feature = "mock")]
     Mock(std::sync::Arc<crate::MockDatabaseConnection>),
+    Disconnected,
 }
 
-impl Default for DatabaseConnection {
+impl Default for InnerConnection {
```
tyt2y3, line 73: I think this is not needed?

tyt2y3:
Ah sad, after a third thought I think InnerConnection::Disconnected is unnecessary

src/database/stream/transaction.rs
```diff
@@ -88,8 +88,14 @@ impl<'a> TransactionStream<'a> {
...
                 }
-                #[allow(unreachable_patterns)]
-                _ => unreachable!(),
+                InnerConnection::Disconnected => {
+                    let _start = _metric_callback.is_some().then(std::time::SystemTime::now);
+                    let stream = Box::pin(futures::stream::iter(
+                        Some(Err(conn_err("Disconnected"))).into_iter(),
+                    ));
+                    let elapsed = _start.map(|s| s.elapsed().unwrap_or_default());
+                    MetricStream::new(_metric_callback, stmt, elapsed, stream)
```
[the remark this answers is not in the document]
billy1624: Will be `unreachable!()` now since we will revert the `Disconnected` variant

src/driver/mock.rs
```diff
@@ -34,13 +34,13 @@ pub trait MockDatabaseTrait: Send + Debug {
...
-    fn begin(&mut self);
+    fn begin(&mut self) -> Result<(), DbErr>;
 
     /// Commit a successful transaction atomically into the [MockDatabase]
-    fn commit(&mut self);
+    fn commit(&mut self) -> Result<(), DbErr>;
 
     /// Roll back a transaction since errors were encountered
-    fn rollback(&mut self);
+    fn rollback(&mut self) -> Result<(), DbErr>;
```
tyt2y3, line 43: I think we shouldn't changes these

tyt2y3:
While converting unwrap into expect is always good, even better if we can make it an Err if the surrounding function is a Result.
But I don't think this PR should change API interfaces from returning nothing into Result.

tyt2y3 (changes requested):
There are too many breaking changes stapled together, which is risky.
Can we avoid all breaking changes, if possible?

src/driver/mock.rs
billy1624: Reverted

billy1624:
> There are too many breaking changes stapled together, which is risky. Can we avoid all breaking changes, if possible?

Sure! Check again :D

src/database/mock.rs
```diff
@@ -162,45 +156,47 @@ impl MockDatabaseTrait for MockDatabase {
...
-            if self.transaction.as_mut().unwrap().commit(self.db_backend) {
-                let transaction = self.transaction.take().unwrap();
-                self.transaction_log.push(transaction.into_transaction());
+        match self.transaction.as_mut() {
+            Some(transaction) => {
+                if transaction.commit(self.db_backend) {
+                    let transaction = self
+                        .transaction
+                        .take()
+                        .expect("There is no open transaction to commit");
```
tyt2y3, line 179: I am sure this is unreachable, as it is already checked to be Some

src/database/mock.rs
```diff
@@ -361,30 +357,36 @@ impl OpenTransaction {
...
-            true
-        } else {
-            self.push(Statement::from_string(
-                db_backend,
-                format!("RELEASE SAVEPOINT savepoint_{}", self.transaction_depth),
-            ));
-            self.transaction_depth -= 1;
-            false
+        match self.transaction_depth {
+            0 => {
```
tyt2y3, line 361: One more indentation is also not necessary

src/database/transaction.rs
```diff
@@ -139,13 +139,13 @@ impl DatabaseTransaction {
...
+                    .map_err(sqlx_error_to_query_err)
             }
             #[cfg(feature = "mock")]
-            InnerConnection::Mock(ref mut c) => {
-                c.begin();
-            }
-        }
+            InnerConnection::Mock(ref mut c) => c.begin(),
+            #[allow(unreachable_patterns)]
+            _ => Err(conn_err("Disconnected")),
```
tyt2y3, line 147: Yeah this seems interesting

src/database/transaction.rs
```diff
@@ -213,32 +213,32 @@ impl DatabaseTransaction {
...
+        }?;
         self.open = false;
         Ok(())
     }
 
     // the rollback is queued and will be performed on next async operation, like returning the connection to the pool
     #[instrument(level = "trace")]
-    fn start_rollback(&mut self) {
+    fn start_rollback(&mut self) -> Result<(), DbErr> {
```
tyt2y3, line 241: This also should not be changed

src/driver/mock.rs
```diff
@@ -157,19 +170,22 @@ impl MockDatabaseConnection {
 
     /// Create a statement block  of SQL statements that execute together.
     #[instrument(level = "trace")]
-    pub fn begin(&self) {
-        self.mocker.lock().unwrap().begin()
+    pub fn begin(&self) -> Result<(), DbErr> {
+        self.mocker.lock().map_err(exec_err)?.begin();
```
tyt2y3, line 174: Returning an Error here does not mean much as the only possibility that a lock fail is another thread already panicked

src/driver/sqlx_sqlite.rs
```diff
@@ -48,8 +48,8 @@ impl SqlxSqliteConnector {
             .url
             .parse::<SqliteConnectOptions>()
             .map_err(sqlx_error_to_conn_err)?;
-        if options.sqlcipher_key.is_some() {
-            opt = opt.pragma("key", options.sqlcipher_key.clone().unwrap());
+        if let Some(ref sqlcipher_key) = options.sqlcipher_key {
```
tyt2y3, line 51: It really could have been:

```suggestion
        if let Some(sqlcipher_key) = &options.sqlcipher_key {
```

`ref` is really a relic now..

src/entity/active_model.rs
```diff
@@ -547,15 +551,18 @@ pub trait ActiveModelTrait: Clone + Debug {
...
 
         // Transform attribute that exists in JSON object into ActiveValue::Set, otherwise ActiveValue::NotSet
         for (col, json_key_exists) in json_keys {
-            if json_key_exists && !am.is_not_set(col) {
-                am.set(col, am.get(col).unwrap());
-            } else {
-                am.not_set(col);
+            match (json_key_exists, am.get(col)) {
+                (true, ActiveValue::Set(value) | ActiveValue::Unchanged(value)) => {
+                    am.set(col, value);
```
tyt2y3, line 561: This is good

src/executor/query.rs
```diff
@@ -276,23 +282,21 @@ macro_rules! try_getable_unsigned {
...
+                    QueryResultRow::SqlxMySql(row) => row
+                        .try_get::<Option<$type>, _>(idx.as_sqlx_mysql_index())
+                        .map_err(|e| sqlx_error_to_query_err(e).into())
+                        .and_then(|opt| opt.ok_or_else(|| err_null_idx_col(idx))),
                     #[cfg(feature = "sqlx-postgres")]
-                    QueryResultRow::SqlxPostgres(_) => {
-                        panic!("{} unsupported by sqlx-postgres", stringify!($type))
-                    }
+                    QueryResultRow::SqlxPostgres(_) => Err(query_err(format!(
+                        "{} unsupported by sqlx-postgres",
```
tyt2y3, line 291: This is not a query error, better to be a Type Error

src/query/update.rs
```diff
@@ -92,11 +92,11 @@ where
     fn prepare_filters(mut self) -> Self {
         for key in <A::Entity as EntityTrait>::PrimaryKey::iter() {
             let col = key.into_column();
-            let av = self.model.get(col);
-            if av.is_set() || av.is_unchanged() {
```
tyt2y3, line 96: I reckon this is because ActiveValue used to be a struct

tyt2y3:
This is still a LOT of changes and hopefully I did not miss a fine detail

src/database/mock.rs
billy1624: How about this? https://github.com/SeaQL/sea-orm/pull/1231/commits/0954fab7a1932ed064f2591ce7dc69143edebd5e

src/database/mock.rs
billy1624: I reverted it https://github.com/SeaQL/sea-orm/pull/1231/commits/840cd543b002b0f6578093d76a08cb1b47b23433

src/database/mock.rs
tyt2y3: Much better

src/database/db_connection.rs
```diff
@@ -351,14 +344,24 @@ impl DatabaseConnection {
...
         }
     }
 
-    /// Get the transaction log as a collection  Vec<[crate::Transaction]>
+    /// Get the transaction log as a collection Vec<[crate::Transaction]>, returning an error if it fails
+    pub fn try_into_transaction_log(self) -> Result<Vec<crate::Transaction>, DbErr> {
+        let mut mocker = self
+            .as_mock_connection()
+            .get_mocker_mutex()
+            .lock()
```
tyt2y3, line 356: Same theory applies

src/database/transaction.rs
billy1624: This is private method of a struct impl. So it should be fine.

src/driver/mock.rs
billy1624: https://github.com/SeaQL/sea-orm/pull/1231/commits/14903a2e6ce435f33e197435d81203381e1a81e9

src/driver/sqlx_sqlite.rs
billy1624: https://github.com/SeaQL/sea-orm/pull/1231/commits/cb9cd13ee67bbb111b8ad37e63814e11116598bd

src/executor/query.rs
billy1624: https://github.com/SeaQL/sea-orm/pull/1231/commits/b665c429356a2b70bbb727ff267552c2ec926cb4

src/database/db_connection.rs
billy1624: https://github.com/SeaQL/sea-orm/pull/1231/commits/3881f0adf7841f91c4031786a7e1faf2568b930a

src/entity/active_model.rs
billy1624: :)

PR #1236, opened 2022-11-21
Close connection explicitly

billy1624:
## New Features

- [x] Adds `close` method to close connection explicitly. (Note: it will be automatically closed when it goes out of scope without closing it explicitly.)


tyt2y3:
I don't think this method can be put in the `ConnectionTrait`, for the fact that `Pool::close` and `Transaction::close` do completely different things, where `Pool::close` is basically endgame.

billy1624:
Clearly I went to far. We should just add a method to close `DatabaseConnection`, that'd be enough.

Issue #1191, opened 2022-11-05
[PIP] Simple data loader


tyt2y3:
I could not think of a good enough API for the 'complex' dataloader yet, but I figured we could ship something incremental.

The problem's gist is not SQL, it's more about the rigid type system in Rust.

Anyway, lets take a look at `ModelTrait`:

```rust
trait ModelTrait {
    ...
    fn find_related<R>(&self, _: R) -> Select<R>
    where
        R: EntityTrait,
        Self::Entity: Related<R>,
    {
        <Self::Entity as Related<R>>::find_related().belongs_to(self)
    }
    ...
}
```

It allows one to find any related entity of a given model.

What we would offer, is a 'vectorized' version of it, meaning given a vec of models we can load all related models of a given relation in one batch.

```rust
impl LoaderTrait for Vec<M>
where M: ModelTrait {
    async fn load_one<R>(&self, _: R) -> Result<Vec<R::Model>>
    where
        R: EntityTrait,
        M::Entity: Related<R>,
    {
        // we should verify this is a has_one relation
    }

    async fn load_many<R>(&self, _: R) -> Result<Vec<Vec<R::Model>>>
    where
        R: EntityTrait,
        M::Entity: Related<R>,
    {
        // we should verify this is a has_many relation
    }
}
```

What the simple loader do under the hood:

1. load all related entities with a subquery
```sql
SELECT [related entity's columns] FROM [related entity table] WHERE [primary key] IN (
    SELECT [related entity's primary key] FROM [relation table, might be parent table or a junction table] WHERE [parent entity primary key] IN ([the list of entity provided to Loader])
)
```
2. aggregate these Models according to the input models's primary key (a hash map would be needed as an intermediate step)
2.1. we should assume that the related entity might be shared; i.e. A1 & A2 both relates to B1, we will only get one B1, so we have to clone it

With this new API, we can now:

```rust
let cakes: Vec<cake::Model> = Cake::find().all(db).await?;
let fruits: Vec<Vec<fruit::Model>> = cakes.load_many(Fruit).await?;
// more relations

// So we can consume them
for (cake: cake::Model, fruits: Vec<fruit::Model>) in std::iter::zip(cakes, fruits) {
    // fruits in this scope belong to cake
}
```

In the end, this would allow us to cascade-load multiple relations eagerly.

tyt2y3:
@karatakis are you interested in working on this?

karatakis:
[...]

PR #1238, opened 2022-11-21
 [PIP] Simple data loader

karatakis:
## PR Info
Allows to query related entities from vector

### Load one
```rust
[...]
    let bakers = baker::Entity::find()
        .all(&ctx.db)
        .await
        .expect("Should load bakers");

    let bakeries = bakers
        .load_one(bakery::Entity::find(), &ctx.db)
        .await
        .expect("Should load bakeries");
[...]
```

### Load many

```rust
[...]
    let bakeries = bakery::Entity::find()
        .all(&ctx.db)
        .await
        .expect("Should load bakeries");

    let bakers = bakeries
        .load_many(baker::Entity::find().filter(baker::Column::Name.like("Baker%")), &ctx.db)
        .await
        .expect("Should load bakers");
[...]
```

- Closes https://github.com/SeaQL/sea-orm/issues/1191

## New Features

- [X] Add simple Dataloader trait


karatakis:
Its not ready yet. I need to do some testing

tyt2y3:
The implementation looks great!

Made a few improvements:
1. Use ValueTuple to replace Vec<Value>
2. Remove Debug trait bounds
3. Slightly improve the test case

May be we still want more test cases?

I am tempted to change `HashMap<String, _>` to `HashMap<ValueTuple, _>`, but of course, `Value` is not hashable because of f32/f64. And I imagine the attempt to make a hashable version of Value to be quite difficult.

src/query/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+                .in_tuples(keys.iter().cloned()),
+            )
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[tokio::test]
+
```
tyt2y3, line 255: How do I remove this line?
```suggestion

```

src/query/loader.rs
```diff
@@ -0,0 +1,340 @@
...
+            cakes,
+            vec![Some(cake::Model {
+                id: 1,
+                name: "New York Cheese".to_owned(),
+            })]
+        );
+    }
+
+    #[tokio::test]
+
```
tyt2y3, line 297:
```suggestion

```

karatakis:
@tyt2y3 thanks for the feedback and the extra work.

@tyt2y3 do you want me to work on making Value Hash 'able?
If you do assign an issue and assign it to me to work on it :smile:.
This is also needed in some cases for Seaography.

Is anything else needed for this issue ?

tyt2y3:
Yes we can come back to this after we got a `HashableValue` .

Also may be awaiting @billy1624 's review

tyt2y3:
> 2.1. we should assume that the related entity might be shared; i.e. A1 & A2 both relates to B1, we will only get one B1, so we have to clone it

Can we also test this?

tyt2y3:
Created issue for HashableValue
https://github.com/SeaQL/sea-orm/issues/1254

karatakis:
Regarding https://github.com/SeaQL/sea-orm/pull/1238#issuecomment-1326635798

It didn't panic, but it was returning None :'(. I fixed it by copying the values before returning.

billy1624 (commented):
Hey @karatakis, sorry for coming in late. nice stuffs!! Thanks for the hard work!!

I think we should also implement the loader for `M: ModelTrait`. Providing a "complete" API, `Vec<M>` implemented `LoaderTrait`, `M` itself should also implement it.

#### Load one

```rust
let bakers = baker::Entity::find()
    .one(&ctx.db) // `one` here, instead of all
    .await?;

let bakeries: Option<_> = bakers // `Option<_>` here, instead of `Vec<Option<_>>`
    .load_one(bakery::Entity::find(), &ctx.db)
    .await?;
```

#### Load many

```rust
let bakeries = bakery::Entity::find()
    .one(&ctx.db) // `one` here, instead of all
    .await?;

let bakers: Vec<_> = bakeries // `Vec<_>` here, instead of `Vec<Vec<_>>`
    .load_many(baker::Entity::find().filter(baker::Column::Name.like("Baker%")), &ctx.db)
    .await?;
```

That means we need two extra associate types to type def the return type of both methods.

```rust
pub trait LoaderTrait {
    // Existing associated type
    type Model: ModelTrait;

    // New associated type to type def the return types
    type LoadOneResult;
    type LoadManyResult;
```

tyt2y3:
> I think we should also implement the loader for M: ModelTrait. Providing a "complete" API, Vec<M> implemented LoaderTrait, M itself should also implement it.

Good point, although I think we don't have to mess the implementation, I suppose we can simply provide an implementation that just wraps a Vec of 1 item.

I want to merge this for 0.11

src/query/loader.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///  Used to eager load has_many relations
+    async fn load_many<R, C>(&self, stmt: Select<R>, db: &C) -> Result<Vec<Vec<R::Model>>, DbErr>
+    where
+        C: ConnectionTrait,
+        R: EntityTrait,
+        R::Model: Send + Sync,
+        <<Self as LoaderTrait>::Model as ModelTrait>::Entity: Related<R>;
+}
+
+#[async_trait::async_trait]
```
tyt2y3, line 32:
```suggestion
#[async_trait]
```

Issue #1123, opened 2022-10-16
[PIP] Event stream


tyt2y3:
## Motivation

To generate a high level event stream of changes to Entities, and serve as a foundation to change capture and other event-based system paradigm.

## Architecture

When this flag is enabled, `DatabaseConnection` will maintain a [broadcast channel](https://docs.rs/async-broadcast/latest/async_broadcast/) (is there non-tokio equivalent?) internally, and users will be able to subscribe to this channel (via `subscribe(&self) -> Receiver<Event>`.

If the execution succeeded, `Inserter`, `Updater` & `Deleter` will extract some semantic information from the query and generate an Event, and push to the channel.

Information we can extract and represent:

1. New Model: the newly inserted ID and content of the new Model
2. Update Model: ID of the model, which fields are changed and the updated values
3. Delete Model: ID of the deleted model, and perhaps content of the deleted Model

Notes 1: We should not buffer events. When no one is subscribing, we will discard all events.
Notes 2: I imagine a subscriber can persist these event, may be even saving to the same database. To prevent infinite echo, we should have an option to ignore certain tables.

## Breaking changes

It seems that we need to convert DatabaseConnection from a enum to a struct.

## Open questions

1. How to allow custom SQL execution to generate events
2. How to allow custom processors to handle `InsertStatement` and `UpdateStatement`

teenjuna:
Question: since broadcast channel is generic, what will be the type?  It could be `dyn ModelTrait`, but is it possible to determine which model it actually is?

billy1624:
I think [async-channel](https://github.com/smol-rs/async-channel) would be a good alternative.

billy1624:
Hey @teenjuna, I think the generic is on the message channel but not the message itself. You can check https://github.com/SeaQL/sea-orm/pull/1239 and see how I create broadcast channel on any supported message channel. More specifically, the `EventStream` trait

netthier:
Would this be for all updates occurring in the database or per table?
Being able to subscribe to updates only in one table would solve the issue @teenjuna mentioned, and also make handling the messages easier (assuming one is only interested in updates from a single table).

tyt2y3:
I think a downstream component would be responsible for filtering and forwarding messages to interested receivers so here the design should focus on the source of truth.

TheCataliasTNT2k:
Will it work with delete events caused by "cascade delete"? So if I delete an entity and related entities are deleted automatically by sea orm, can I get the deleted related entities?
[...]

PR #1239, opened 2022-11-21
Event stream

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1123

## New Features

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ uuid = { version = "^1", default-features = false, optional = true }
 ouroboros = { version = "0.15", default-features = false }
 url = { version = "^2.2", default-features = false }
 thiserror = { version = "^1", default-features = false }
+tokio = { version = "^1.6", default-features = false, features = ["sync"] }
+async-channel = { version = "^1.7", default-features = false }
 
 [dev-dependencies]
 smol = { version = "^1.2" }
@@ -59,6 +61,7 @@ pretty_assertions = { version = "^0.7" }
 time = { version = "^0.3", features = ["macros"] }
 uuid = { version = "^1", features = ["v4"] }
 once_cell = "1.8"
+async-channel = { version = "^1.7" }
 
 [features]
 debug-print = []
--- a/src/database/db_connection.rs
+++ b/src/database/db_connection.rs
@@ -1,6 +1,6 @@
 use crate::{
-    error::*, ConnectionTrait, DatabaseTransaction, ExecResult, QueryResult, Statement,
-    StatementBuilder, StreamTrait, TransactionError, TransactionTrait,
+    error::*, ConnectionTrait, DatabaseTransaction, EventStream, ExecResult, QueryResult,
+    Statement, StatementBuilder, StreamTrait, TransactionError, TransactionTrait,
 };
 use sea_query::{MysqlQueryBuilder, PostgresQueryBuilder, QueryBuilder, SqliteQueryBuilder};
 use std::{future::Future, pin::Pin};
@@ -282,6 +282,15 @@ impl DatabaseConnection {
             }
             _ => {}
         }
+    }
+
+    pub fn set_event_stream<E>(&mut self, event_stream: E) -> E::Receiver
+    where
+        E: EventStream,
+    {
+        let (sender, receiver) = event_stream.subscribe();
+        // TODO: Save the `sender` in `DatabaseConnection`
+        receiver
     }
 }
 
--- a/src/database/event.rs
+++ b/src/database/event.rs
@@ -1,0 +1,81 @@
+use crate::DbErr;
+
+pub trait EventStream {
+    type Sender: EventSender;
+    type Receiver: EventReceiver;
+
+    fn subscribe(self) -> (Self::Sender, Self::Receiver);
+}
+
+#[async_trait::async_trait]
+pub trait EventSender {
+    async fn send(&self, event: Event) -> Result<(), DbErr>;
+}
+
+#[async_trait::async_trait]
+pub trait EventReceiver {
+    async fn recv(&mut self) -> Result<Event, DbErr>;
+}
+
+#[derive(Debug, Clone)]
+pub enum Event {
+    Insert,
+    Update,
+    Delete,
+}
+
+mod event_stream_tokio {
+    use super::*;
+    use tokio::sync::broadcast::{Receiver, Sender};
+
+    impl EventStream for (Sender<Event>, Receiver<Event>) {
+        type Sender = Sender<Event>;
+        type Receiver = Receiver<Event>;
+
+        fn subscribe(self) -> (Self::Sender, Self::Receiver) {
+            self
+        }
+    }
+
+    #[async_trait::async_trait]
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -43,6 +43,8 @@ uuid = { version = "^1", default-features = false, optional = true }
 ouroboros = { version = "0.15", default-features = false }
 url = { version = "^2.2", default-features = false }
 thiserror = { version = "^1", default-features = false }
+tokio = { version = "^1.6", default-features = false, features = ["sync"] }
```
tyt2y3, line 46: I think we need to avoid depending on `tokio`

Cargo.toml
```diff
@@ -43,6 +43,8 @@ uuid = { version = "^1", default-features = false, optional = true }
...
+async-channel = { version = "^1.7", default-features = false }
```
tyt2y3, line 47: This is not a broadcast channel, as each message will only be received by one consumer.

tyt2y3:
https://github.com/jonhoo/bus might be a good option

billy1624:
Hmmmm I think we need a async broadcast channel. How about https://crates.io/crates/async-broadcast?

billy1624:
Design choice:

1. Representing model values in `HashMap`

```rs
#[derive(Debug, Clone)]
pub struct Event {
    pub entity_type_id: TypeId,
    pub action: EventAction,
    pub values: HashMap<String, Value>,
}
```

2. Receiver side: filter by entity and then unpack the values

```rs
let mut tokio_receiver = db.set_event_stream(async_broadcast::broadcast(10));

while let Ok(event) = tokio_receiver.recv().await {
    // Filter by entity
    if event.of_entity::<cake::Entity>() {
        // Unpack the values
        if let Some(val) = event.values.get(cake::Column::Name.as_str()) {
            todo!()
        }
    }
}
```

Diwakar-Gupta:
> Hmmmm I think we need an async broadcast channel. How about https://crates.io/crates/async-broadcast?

I did checked this one here https://pastebin.com/mvMwx3Ch.

There are 2 problems with this library
1. when one the receiver is closed, it closed the channel, Hence no other receiver will get any event and sender wont be able to send event updates. Check after line 30 in the shared Pastebin.
2. broadcast caches the messages ( number of cached messages can be set to 1 `broadcast(cap = 2);` ). tyt2y3 mentioned https://github.com/SeaQL/sea-orm/issues/1123#issue-1410534966 that **Notes 1: We should not buffer events. When no one is subscribing, we will discard all events.**

tyt2y3:
@Diwakar-Gupta thank you for you investigation. Your example is easy to read!

> Close one receiver, which closes all the sender and receivers.

It seems to be deal-breaker. But according to the docs, this is not the intended usage:

> When all Senders or all Receivers are dropped, the channel becomes closed. When a channel is closed, no more messages can be sent, but remaining messages can still be received.

Note that, dropping the Receiver is a different operation from manually closing it.

tyt2y3:
Are you also interested in investigating the `bus` crate?

Diwakar-Gupta:
Hi @tyt2y3 I did some work with `bus` and I think it will work.
About `async-runtime` I don't think bus uses async rather they depend on os-threads. Check this file https://github.com/jonhoo/bus/blob/main/src/lib.rs

Cargo.toml
`bus = "2.3.0"`

```rust
fn main() {
    use bus::Bus;
    let mut tx = Bus::new(2); // len = 2
    let mut rx1 = tx.add_rx();
    
    tx.broadcast(1);

    let mut rx2 = tx.add_rx(); // should not receive 1
    tx.broadcast(2);
    
    assert_eq!(rx1.recv(), Ok(1));
    assert_eq!(rx1.recv(), Ok(2));

    // rx2 don't receive past events
    assert_eq!(rx2.recv(), Ok(2));

    drop(rx2);

    tx.broadcast(3);
    assert_eq!(rx1.recv(), Ok(3));

    // send more events than len and don't consume value
    let mut rx3 = tx.add_rx();
    tx.broadcast(4);
    tx.broadcast(5);
    assert_eq!(tx.try_broadcast(6), Err(6));

    // 1 receiver consume's value 
    assert_eq!(rx3.recv(), Ok(4));
    assert_eq!(tx.try_broadcast(6), Err(6));

    // both receiver consumes value
    assert_eq!(rx1.recv(), Ok(4));
    assert_eq!(tx.try_broadcast(6), Ok(()));
}
```

My observation
1. broadcast works perfectly fine with multiple receivers
2. past event is not send to receiver.
3. Drop works fine.
4. when `Bus::new(len = 2);` len number of messages are send but not consumed yet broadcast will wait till eternity for all the receivers to consume at least one event value. In the official doc they have mentioned to have larger len. read `fn new` https://docs.rs/bus/2.3.0/bus/struct.Bus.html

One other think I wanted to ask should we consider the way javascript web handles event, it takes a function ( function or object in our case ) `onclick(function)` and calls it when event occurs.

billy1624:
Hey @Diwakar-Gupta, thanks for the experiments!! I didn't have time to do it myself just yet. So, glad you helped :))

> 1. when one the receiver is closed, it closed the channel, Hence no other receiver will get any event and sender wont be able to send event updates. Check after line 30.

One shouldn't invoke `Receiver::close()` method. A `Receiver` will be dropped when it goes out of scope as defined in [`impl Drop for Receiver<T>`](https://docs.rs/async-broadcast/0.5.0/async_broadcast/struct.Receiver.html#impl-Drop-for-Receiver%3CT%3E). We want the sender keep alive when there is no receiver but I'm not sure the behaviour of it.

> 2. broadcast caches the messages ( number of cached messages can be set to 1 `broadcast(cap = 2);` ). tyt2y3 mentioned [[PIP] Event stream #1123 (comment)](https://github.com/SeaQL/sea-orm/issues/1123#issue-1410534966) that **Notes 1: We should not buffer events. When no one is subscribing, we will discard all events.**. So future subscribers can get past event notifications.

I found that we cannot spawn a new receiver simply by calling `.clone()`. Use [`new_receiver()`](https://docs.rs/async-broadcast/0.5.0/async_broadcast/struct.Sender.html#method.new_receiver) instead.

Also, can we try [`.set_await_active()`](https://docs.rs/async-broadcast/0.5.0/async_broadcast/struct.Sender.html#method.set_await_active) to `false`? It's `true` by default. With that set, hopefully it won't "cache" past message (event).

billy1624:
Some references:
- https://docs.rs/async-broadcast/0.5.0/async_broadcast/index.html#difference-with-other-broadcast-crates
- https://tokio.rs/tokio/tutorial/channels#tokios-channel-primitives

Diwakar-Gupta:
@billy1624 @tyt2y3 
Have a look at [crossbeam-channel](https://github.com/crossbeam-rs/crossbeam/tree/master/crossbeam-channel#crossbeam-channel) it provides `mpmc` feature and is also merged in rust's standard library here is [pr](https://github.com/rust-lang/rust/pull/93563/) .

Issue #1259, opened 2022-11-28
Generate `Related` for many-to-many tables with extra columns


limjiayi:
This merged pull request describes the current logic for determining many-to-many relations. PR https://github.com/SeaQL/sea-orm/pull/1075

> I updated to logic to only require the intermediate table of a m-to-n relation to be with
> - Exactly 2 foreign keys
> - A primary key formed by exactly 2 columns
> - Number of primary key equals to number of column

But it sometimes makes sense for these junction tables to contain more than 2 columns. For example, a `CustomerProduct` table that links the `Customer` and `Product` tables based on the products that each customer has bought, may have the following columns to record extra information about the relationship.
- `customer_id` (foreign key)
- `product_id` (foreign key)
- `price`
- `quantity`
- `purchase_date`
- `promo_code`

So, the requirement that the number of primary keys be equal to the number of columns should be removed.

PR #1260, opened 2022-11-28
Generate Related for many-to-many relation with extra columns

limjiayi:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1259

## Bug Fixes

- [x] Existing logic for determining a many-to-many relation requires the number of columns in the junction table to be equal to the number of primary keys. This makes codegen work incorrectly for junction tables that record additional information about the relationship, so I have removed this requirement.


billy1624 (commented):
Hey @limjiayi, thanks for contributing!! I second the relaxation.

Thoughts? @tyt2y3

billy1624 (approved):

Issue #1261, opened 2022-11-28
support exclude column


yuhaya:
## Motivation

The data table has 20 fields, but I only want to query 19 of them. Sea-orm does not seem to support field exclusion at present. It is too troublesome to call .column() 19 times

## Proposed Solutions

add select api 

```rust
fn column_exclude<C>(mut self, col: C) -> Self
    where
        C: ColumnTrait,
    {
        // code...
        self
    }
```

billy1624:
Hey @yuhaya, welcome!! Thanks for the suggestions! I just created a PR to perform similar operation.

https://github.com/SeaQL/sea-orm/blob/1520beaa15b86e7eafb7f0679e2c23d59d34c005/src/query/helper.rs#L115-L131

yuhaya:
> Hey @yuhaya, welcome!! Thanks for the suggestions! I just created a PR to perform similar operation.
> 
> https://github.com/SeaQL/sea-orm/blob/1520beaa15b86e7eafb7f0679e2c23d59d34c005/src/query/helper.rs#L115-L131

good!

PR #1264, opened 2022-11-29
Add `QuerySelect::columns` method

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1261

## New Features

- [x] Add `QuerySelect::columns` method: making it possible to conditionally select all columns expect a specific column

### DIFF
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -97,6 +97,72 @@ pub trait QuerySelect: Sized {
         self
     }
 
+    /// Select columns
+    ///
+    /// ```
+    /// use sea_orm::{entity::*, query::*, tests_cfg::cake, DbBackend};
+    ///
+    /// assert_eq!(
+    ///     cake::Entity::find()
+    ///         .select_only()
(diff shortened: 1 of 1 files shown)


src/query/helper.rs
```diff
@@ -97,6 +97,72 @@ pub trait QuerySelect: Sized {
...
+    ///         .select_only()
+    ///         .columns(cake::Column::iter().filter(|col| match col {
+    ///             cake::Column::Id => false,
+    ///             _ => true,
+    ///         }))
+    ///         .build(DbBackend::Postgres)
+    ///         .to_string(),
+    ///     r#"SELECT "cake"."name" FROM "cake""#
+    /// );
+    /// ```
```
billy1624, line 131: `match` were used because neither `Column` nor `ColumnTrait` implemented `Eq`. And we cannot implement `Eq` in `Column`, it will cause nasty conflicting method name since we already have `ColumnTrait::eq` method.

PR #1267, opened 2022-11-30
Aligning IdenStatic with its counterpart in SeaQuery

billy1624:
## PR Info

- Related PR
  - https://github.com/SeaQL/sea-query/pull/508

## Breaking Changes

- [x] `IdenStatic` no longer with `Debug` bound

### DIFF
--- a/sea-orm-macros/src/derives/entity.rs
+++ b/sea-orm-macros/src/derives/entity.rs
@@ -76,7 +76,7 @@ impl DeriveEntity {
                     #expanded_schema_name
                 }
 
-                fn table_name(&self) -> &str {
+                fn table_name(&self) -> &'static str {
                     #table_name
                 }
             }
@@ -126,7 +126,7 @@ impl DeriveEntity {
         quote!(
             #[automatically_derived]
             impl sea_orm::IdenStatic for #ident {
-                fn as_str(&self) -> &str {
+                fn as_str(&self) -> &'static str {
                     <Self as sea_orm::EntityName>::table_name(self)
                 }
             }
--- a/src/entity/base_entity.rs
+++ b/src/entity/base_entity.rs
@@ -4,11 +4,10 @@ use crate::{
     RelationTrait, RelationType, Select, Update, UpdateMany, UpdateOne,
 };
 use sea_query::{Alias, Iden, IntoIden, IntoTableRef, IntoValueTuple, TableRef};
-use std::fmt::Debug;
 pub use strum::IntoEnumIterator as Iterable;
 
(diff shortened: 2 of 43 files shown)


src/entity/base_entity.rs
```diff
@@ -4,13 +4,12 @@ use crate::{
...
 use sea_query::{Alias, Iden, IntoIden, IntoTableRef, IntoValueTuple, TableRef};
-use std::fmt::Debug;
 pub use strum::IntoEnumIterator as Iterable;
 
 /// Ensure the identifier for an Entity can be converted to a static str
-pub trait IdenStatic: Iden + Copy + Debug + 'static {
+pub trait IdenStatic: Iden + Copy + 'static {
     /// Method to call to get the static string identity
-    fn as_str(&self) -> &str;
+    fn as_str(&self) -> &'static str;
```
billy1624, line 12: If we make this returns static reference, it'd be a massive breaking change https://github.com/SeaQL/sea-orm/pull/1267/commits/3cca57f5ac9ea37808ea807bd144893022e6e7cf

src/entity/base_entity.rs
```diff
@@ -21,7 +20,7 @@ pub trait EntityName: IdenStatic + Default {
     }
 
     /// Get the name of the table
-    fn table_name(&self) -> &str;
+    fn table_name(&self) -> &'static str;
```
tyt2y3, line 23: This change is not necessary

src/entity/base_entity.rs
tyt2y3: I think not. I don't think there is a way to get a non-static reference to a string given that the struct must itself is `Copy`

sea-orm-macros/src/derives/entity.rs
```diff
@@ -126,7 +126,7 @@ impl DeriveEntity {
         quote!(
             #[automatically_derived]
             impl sea_orm::IdenStatic for #ident {
-                fn as_str(&self) -> &str {
+                fn as_str(&self) -> &'static str {
                     <Self as sea_orm::EntityName>::table_name(self)
```
billy1624, line 130: Return type of table_name was being changed because of this

src/entity/base_entity.rs
tyt2y3: Only if someone is using an `ArrayString` inside a custom `Entity` definition.

tyt2y3: > Only if someone is using an ArrayString inside a custom Entity definition.

Which I don't think someone had actually done that. So I am okay to this breaking change.

Issue #1286, opened 2022-12-06
How to downgrade sea-orm-cli version?


frederikhors:
If I install with:

```shell
cargo install sea-orm-cli@0.10.2
```

and generate with:

```shell
sea-orm-cli generate entity //......
```

the generated file have:

```rust
//! `SeaORM` Entity. Generated by sea-orm-codegen 0.10.5
```

instead of of `0.10.2`.

If I run

```shell
sea-orm-cli -v
```

I can see:

```shell
sea-orm-cli 0.10.2
```

What is going on?

I need the 0.10.2 version right now.

billy1624:
Hey @frederikhors, please use the following command to degrade `sea-orm-cli`.

```shell
cargo install --version 0.10.2 sea-orm-cli
```

frederikhors:
I tried this too and with `--force` too. Nothing. The version of generated files is 0.10.5 always. Crazy!

billy1624:
Hmmm... why? Did you try uninstall it first then install it again?

frederikhors:
Yeah! Of course!

frederikhors:
I used:

```shell
cargo uninstall sea-orm-cli

Removing C:\Users\Fred\.cargo\bin\sea-orm-cli.exe
Removing C:\Users\Fred\.cargo\bin\sea.exe
```

billy1624:
And try restarting the command prompt after installation

frederikhors:
Then I installed with:

```shell
cargo install --version 0.10.2 sea-orm-cli

...
  Compiling tracing-subscriber v0.3.16
  Compiling clap v3.2.23
  Compiling chrono v0.4.23
  Compiling sqlx v0.6.2
  Compiling sea-query-binder v0.2.2
  Compiling sea-schema v0.10.3
  Compiling sea-orm-codegen v0.10.5
  Compiling sea-orm-cli v0.10.2
  Finished release [optimized] target(s) in 2m 48s
Installing C:\Users\Fred\.cargo\bin\sea-orm-cli.exe
Installing C:\Users\Fred\.cargo\bin\sea.exe
  Installed package `sea-orm-cli v0.10.2` (executables `sea-orm-cli.exe`, `sea.exe`)
```

Now it is 0.10.2.

If I generate the files have the comment (and the differences) of the 0.10.5! Crazy!

frederikhors:
It's crazy because the .exe in cargo's bin directory is the new one!

billy1624:
[...]

PR #1299, opened 2022-12-12
[CLI] sea-orm-cli depends on codegen of the same version

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1286

## Changes

- [x] sea-orm-cli depends on an exact version of codegen, i.e. `=0.10.3`, instead of `^0.10.3` which is the same as `>=0.10.3`.

> Reminder: need to backport this to `0.10.x` branch as well


tyt2y3:
I agree

PR #1300, opened 2022-12-12
Cleaning up dependency

billy1624:
## PR Info

> Caret requirements are an alternative syntax for the default strategy, `^1.2.3` is exactly equivalent to `1.2.3`.
> 
> https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

To make the dependency specification clear and concise, I'd like to define it all in the form of `x.y.z`.

## Changes

- [x] changed all `version = "^x.y.z"` into `version = "x.y.z"`
- [x] disable default features and enable only the needed ones

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,41 +24,41 @@ name = "sea_orm"
 path = "src/lib.rs"
 
 [dependencies]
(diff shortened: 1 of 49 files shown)


tyt2y3 (commented):
✅

PR #1302, opened 2022-12-13
Add wikipedia-speedrun to COMMUNITY.md

hut8:
As per https://github.com/hut8/wikipedia-speedrun/issues/9


tyt2y3:
Interesting! I like this project

PR #1311, opened 2022-12-15
Select into tuple

billy1624:
## New Features

- [x] Added `Select::into_tuple()`

## Breaking Changes

- [x] Added `TryGetable::try_get_by_index()`
- [x] Added `TryGetableMany::try_get_many_by_index()`


tyt2y3 (commented):
I like the overall API 👍

However I think there is too much code duplication, and I think it is possible to refactor `try_get_by_index` and `try_get` with a common `try_get_by<T>`

tyt2y3:
I think the original sqlx `get` is already generic over `usize` and `str` as index, so we can simply delegate the implementation

billy1624:
I think it won't work :(

```rs
fn try_get_by<I: Index>(res: &QueryResult, idx: I) -> Result<Self, TryGetError> { ... }
```

SQLx `try_get` takes `&str` / `usize` but here the compiler only knows the type implemented `Index` trait.

Also, I just found that we cannot implement `sqlx::ColumnIndex` in foreign crate because it's being sealed. https://docs.rs/sqlx/latest/sqlx/trait.ColumnIndex.html

```rs
// Not possible
impl sqlx::ColumnIndex<MySqlRow> for TryGetColumn { ... }
impl sqlx::ColumnIndex<SqliteRow> for TryGetColumn { ... }
impl sqlx::ColumnIndex<PgRow> for TryGetColumn { ... }
```

Then, we have no choice but to store `String` / `usize` and supply that to SQLx's `try_get`. So, I afraid this is the "only" way?

```rs
pub enum TryGetColumn {
    Name(String),
    Index(usize),
}

pub trait IntoTryGetColumn {
    fn into_try_get_column() -> TryGetColumn;
}
```

tyt2y3:
Well, yes and no and yes

tyt2y3 (approved):
Self approve

PR #1346, opened 2022-12-30
Add JsonBinary attribute

### LINKED ISSUE #1344: Generating entity from Postgres with `JsonBinary` field doesn't add attribute to model
## Description

When generating entities from an existing Postgres database, Json Binary fields don't include the attribute `#[sea_orm(column_type = "JsonBinary")]`.

## Steps to Reproduce

This example will go through setting up a migration, running the migration, and then generating entities. This is all being done on a Postgres database.

1. Create a table in a migration, say:

```rust
manager
    .create_table(
        Table::create()
            .table(Task::Table)
            .if_not_exists()
            ...
            .col(ColumnDef::new(Task::Status).json_binary().not_null())
            .col(ColumnDef::new(Task::Payload).json_binary().not_null())
            .to_owned(),
    )
    .await
```

When built, this generates the string

`CREATE TABLE IF NOT EXISTS \"task\" ( \"id\" serial NOT NULL PRIMARY KEY, \"status\" jsonb NOT NULL, \"payload\" jsonb NOT NULL )`

2. Run the migration `sea-orm-cli migrate`
3. Generate the entities `sea-orm-cli generate entity`

### Expected Behavior

The entity:

```rust
pub struct Model {
    pub id: i32,
    #[sea_orm(column_type = "JsonBinary")]
    pub status: Json,
    #[sea_orm(column_type = "JsonBinary")]
    pub payload: Json,
}
```

### Actual Behavior

The entity:

```rust
pub struct Model {
    pub id: i32,
    pub status: Json,
    pub payload: Json,
}
```

### Reproduces How Often

Always

## Versions

```
│   ├── sea-orm v0.10.6
│   │   ├── sea-orm-macros v0.10.6 (proc-macro)
│   │   ├── sea-query v0.27.2
│   │   │   ├── sea-query-derive v0.2.0 (proc-macro)
│   │   ├── sea-query-binder v0.2.2
│   │   │   ├── sea-query v0.27.2 (*)
│   │   ├── sea-strum v0.23.0
│   │   │   └── sea-strum_macros v0.23.0 (proc-macro)
├── sea-orm v0.10.6 (*)
```

From my Docker image, Postgres is:

`image: postgres:14.1`

## Additional Information

n/a

AngelOnFira:
*This PR is still in progress*

## PR Info

- This will close #1344

## Notes

Currently, this only seems to work on compact entity structures. As of writing this, I'm unsure on how the attribute should be added in the expanded form, so I'll need some insights on that.

## TODO

- [ ] Get this to work on expanded entity structure
- [ ] Write tests


AngelOnFira:
So @billy1624 this seems to work correctly now that 

Before (compact form)

```rust
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub status: Json,
    pub payload: Json,
}
```

After (compact form):

```rust
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    #[sea_orm(column_type = "JsonBinary")]
    pub status: Json,
    #[sea_orm(column_type = "JsonBinary")]
    pub payload: Json,
}
```

After (expanded form):

```rust
#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel, Eq, Serialize, Deserialize)]
pub struct Model {
    pub id: i32,
    pub status: Json,
    pub payload: Json,
}

...

#[derive(Copy, Clone, Debug, EnumIter, DerivePrimaryKey)]
pub enum PrimaryKey {
    Id,
}

impl PrimaryKeyTrait for PrimaryKey {
    type ValueType = i32;
    fn auto_increment() -> bool {
        true
    }
}
```

So as we can see, in the expanded form the attributes aren't added. I don't exactly know how they expand, but in the case above I've included the PrimaryKey as it appears to be the expansion of `#[sea_orm(primary_key)]`

What would the next steps be to properly add this attribute to the expanded structure?

Also, re: tests, I'm not sure where a new one would fit best, and I *think* that only Postgres has Binary Json, so would the path `sea-orm-codegen/tests/postgres/binary_json.rs` work?

billy1624 (commented):
Hey @AngelOnFira, thanks for contributing as always!

For the expanded entity, I think it already handled the `JsonBinary` column type.

https://github.com/SeaQL/sea-orm/blob/bef37c3e2d8c3a00a09ddb3dab95ca2bd6d7c92d/sea-orm-codegen/src/entity/column.rs#L149-L150

billy1624 (commented):
> Also, re: tests, I'm not sure where a new one would fit best, and I *think* that only Postgres has Binary Json, so would the path `sea-orm-codegen/tests/postgres/binary_json.rs` work?

Good call! Yes, please :)

AngelOnFira:
> For the expanded entity, I think it already handled the JsonBinary column type.

Oh, is this the part of the extended format that covers that?

```rust
impl ColumnTrait for Column {
    type EntityName = Entity;
    fn def(&self) -> ColumnDef {
        match self {
            Self::Id => ColumnType::Integer.def(),
            Self::Status => ColumnType::JsonBinary.def(),
            Self::Payload => ColumnType::JsonBinary.def(),
        }
    }
}
```

It does look like it's working now, so I must have had something out of sync when I first thought it was an issue :sweat_smile: either way, glad to get the tests wrapped up on this PR :+1:

billy1624 (commented):
Hey @AngelOnFira, thanks for adding test case for the compact entity format. Would it be possible to include a test case for the expanded format as well?

AngelOnFira:
@billy1624 that should be everything :) let me know if any test could be improved!

tyt2y3 (approved):

billy1624 (approved):
Thanks!! @AngelOnFira Appreciated!

PR #1368, opened 2023-01-05
Implement LoaderTrait for slices

billy1624:
## New Features

- [x] Implement `LoaderTrait` for slices: `[M; N]` and `&[M]`

### DIFF
--- a/src/query/loader.rs
+++ b/src/query/loader.rs
@@ -32,8 +32,35 @@ pub trait LoaderTrait {
 #[async_trait]
 impl<M> LoaderTrait for Vec<M>
 where
-    M: ModelTrait,
-    Vec<M>: Sync,
+    M: ModelTrait + Sync,
+{
+    type Model = M;
+
+    async fn load_one<R, C>(&self, stmt: Select<R>, db: &C) -> Result<Vec<Option<R::Model>>, DbErr>
+    where
+        C: ConnectionTrait,
+        R: EntityTrait,
+        R::Model: Send + Sync,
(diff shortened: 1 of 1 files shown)


src/query/loader.rs
```diff
@@ -32,8 +32,63 @@ pub trait LoaderTrait {
...
+        R::Model: Send + Sync,
+        <<Self as LoaderTrait>::Model as ModelTrait>::Entity: Related<R>,
+    {
+        self.as_slice().load_many(stmt, db).await
+    }
+}
+
+#[async_trait]
+impl<M, const N: usize> LoaderTrait for [M; N]
+where
```
tyt2y3, line 62: I don't think we needed this, as it's just trivial casting `[M; N]` to `&[M]`?

billy1624: Yes, it's trivial to cast it. But wouldn't it be awesome we can write... directly?

```rs
let fruits = [fruit::Model {
    id: 1,
    name: "Apple".to_owned(),
    cake_id: Some(1),
}];

let cakes = fruits
    .load_one(cake::Entity::find(), &db)
    .await?;
```

tyt2y3: In your example it think it will "auto as ref" so `as_slice` is not needed?

billy1624: I tested it. It said "`[M; N]` does not implement ..."
So, it won't "auto as ref".

tyt2y3: Ah sad, but I still don't like it.

billy1624: I'll remove it then :(

Issue #1378, opened 2023-01-09
Atomic migrations


billy1624:
### Discussed in https://github.com/SeaQL/sea-orm/discussions/1059

<div type='discussions-op-text'>

<sup>Originally posted by **gd87429** September 23, 2022</sup>
If you run `Migrator::up()` in parallel (like when you spin up multiple servers), is there a chance the same migration might run more than once?

If yes -- are there plans to build-in support for atomic migrations, or should the user make sure migration cannot be applied twice (like checking if data exists before seeding the table)?</div>

billy1624:
Related Discussions:
- https://github.com/SeaQL/sea-orm/discussions/1324
- https://github.com/SeaQL/sea-orm/discussions/1059

PR #1379, opened 2023-01-09
Atomic migration

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1378

## New Features

- [x] Executes "atomic" migrations on Postgres, that is migrations will be executed with a transaction when the target database is Postgres
- [x] Starts a transaction and execute SQL inside migration up and down
```rs
use sea_orm_migration::prelude::*;
use sea_orm_migration::sea_orm::{entity::*, query::*};

pub struct Migration;

#[async_trait::async_trait]
impl MigrationTrait for Migration {

        let transaction = db.begin().await?;

        cake::ActiveModel {
            name: Set("Cheesecake".to_owned()),
            ..Default::default()
        }
        .insert(&transaction)
        .await?;

        transaction.commit().await?;

        if std::env::var_os("ABOARD_MIGRATION").eq(&Some("YES".into())) {
            return Err(DbErr::Migration(
                "Aboard migration and rollback changes".into(),
            ));
        }

        Ok(())
    }

        let transaction = db.begin().await?;

        cake::Entity::delete_many()
            .filter(cake::Column::Name.eq("Cheesecake"))
            .exec(&transaction)
            .await?;

        transaction.commit().await?;

        Ok(())
    }
}

mod cake {
    use sea_orm_migration::sea_orm::entity::prelude::*;

    pub struct Model {
        #[sea_orm(primary_key)]
        pub id: i32,
        pub name: String,
    }

    pub enum Relation {}

}
```

## Breaking Changes

- [x] Changed the type of `SchemaManager::conn` to `SchemaManagerConnection`
- [x] Changed the input type of `SchemaManager::new()` to `IntoSchemaManagerConnection`
- [x] Changed the return type of `SchemaManager::get_connection()` to `&SchemaManagerConnection<'c>`
- [x] Changed the connection parameter of `MigratorTrait`'s methods
    - [x] `get_migration_models`, `get_migration_with_status`, `get_pending_migrations`, `get_applied_migrations`, `install` and `status` methods take `ConnectionTrait` as the connection
    - [x] `fresh`, `refresh`, `reset`, `up` and `down` methods take `IntoSchemaManagerConnection` as the connection

## Changes

- [x] Added `SchemaManagerConnection` and `IntoSchemaManagerConnection`


tyt2y3:
Just to clear some questions in my mind, so the updated default behaviour will be wrapping the entire `up/down` execution in a transaction, only if it's postgres?

billy1624:
> Just to clear some questions in my mind, so the updated default behaviour will be: wrapping the entire `up/down` execution in a transaction, only if it's postgres?

Hey @tyt2y3, the new behaviour is:

The `fresh`, `refresh`, `reset`, `up` and `down` methods now take a `&DatabaseConnection` or a `&DatabaseTransaction`.

Then, inside each method, it will start a transaction to execute the migrations only if it's a Postgres connection. For non-Postgres connection, it just use connection as-is and execute the migrations as usual.

sea-orm-migration/tests/main.rs
```diff
@@ -107,6 +107,33 @@ async fn run_migration(url: &str, db_name: &str, schema: &str) -> Result<(), DbE
     assert!(!manager.has_table("cake").await?);
     assert!(!manager.has_table("fruit").await?);
 
+    // Tests rolling back changes of "migrate up" when running migration on Postgres
+    if matches!(db.get_database_backend(), DbBackend::Postgres) {
+        println!("\nRoll back changes when encounter errors");
+
+        // Set a flag to throw error inside `m20230109_000001_seed_cake_table.rs`
+        std::env::set_var("ABOARD_MIGRATION", "YES");
```
tyt2y3, line 115: I suspect this should be `abort` ?

sea-orm-migration/src/manager.rs
```diff
@@ -1,20 +1,26 @@
...
     TableTruncateStatement,
 };
-use sea_orm::{ConnectionTrait, DbBackend, DbConn, DbErr, StatementBuilder};
+use sea_orm::{ConnectionTrait, DbBackend, DbErr, StatementBuilder};
 use sea_schema::{mysql::MySql, postgres::Postgres, probe::SchemaProbe, sqlite::Sqlite};
 
 /// Helper struct for writing migration scripts in migration file
 pub struct SchemaManager<'c> {
-    conn: &'c DbConn,
+    conn: SchemaManagerConnection<'c>,
```
tyt2y3, line 13: Actually, I think we can `Box<dyn ConnectionTrait>` it as well?

sea-orm-migration/tests/migrator/m20230109_000001_seed_cake_table.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let db = manager.get_connection();
+
+        let transaction = db.begin().await?;
```
tyt2y3, line 12: Why open a transaction here? Just want to confirm, the behaviour should be no different whether we open a transaction or not? Because it's already inside a transaction (on Postgres)?

sea-orm-migration/tests/migrator/m20230109_000001_seed_cake_table.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        cake::ActiveModel {
+            name: Set("Cheesecake".to_owned()),
+            ..Default::default()
+        }
+        .insert(&transaction)
+        .await?;
+
+        transaction.commit().await?;
```
tyt2y3, line 21: For any backend, aborting the migration BEFORE committing makes more sense?

sea-orm-migration/tests/main.rs
billy1624: Oppps, what a typo

sea-orm-migration/tests/migrator/m20230109_000001_seed_cake_table.rs
billy1624: 1. I want to show it's possible to open a transaction inside migrate `up` & `down`.

2. For Postgres, yes, it's no difference. Because as you said you're already inside a transaction. However, it's not the case, for MySQL and SQLite. Opening a transaction is vital because the migration itself isn't atomic. If you want to seed the database atomically then you need to start a transaction yourself.

sea-orm-migration/tests/migrator/m20230109_000001_seed_cake_table.rs
billy1624: Sure! see if both transactions got rolled back at the same time

sea-orm-migration/tests/migrator/m20230109_000001_seed_cake_table.rs
tyt2y3: Got it, cool

tyt2y3 (approved):
OK. I also have some adjustments to make

Issue #1376, opened 2023-01-09
Syntax Error when saving active model that sets nothing


billy1624:
### Discussed in https://github.com/SeaQL/sea-orm/discussions/1371

<div type='discussions-op-text'>

<sup>Originally posted by **lihe07** January  8, 2023</sup>
I have a ActiveModel like this;
```rust
ActiveModel {
    id; Unchanged(1),
    name: NotSet,
    password: NotSet
}
```
The SQL query for saving this looks like that:
```sql
UPDATE
  `users`
SET
WHERE
  `users`.`id` = ?
```
And that leads to a syntax error (empty set section).

I'm considering whether it's sea-orm's task to determine if the ActiveModel does noting or not.</div>

lihe07:
Thanks! I'd be happy to work this out if I can!
(and I don't need a patch for my project either)

billy1624:
Hey @lihe07, thanks for the interest!! Sorry, I didn't realise you want to take this. Please check my PR at https://github.com/SeaQL/sea-orm/pull/1384

lihe07:
Never mind! I'll keep an eye out too, if there's anything I can do to help.

PR #1384, opened 2023-01-10
Noop when update without providing any values

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1376

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/578

## Bug Fixes

- [x] When updating / saving an existing `ActiveModel` with all its value being `Unchanged` or `NotSet`, the expected behaviour should be no-op. That's to skip firing any update statement.


Cargo.toml
```diff
@@ -34,8 +34,8 @@ tracing = { version = "0.1", default-features = false, features = ["attributes",
 rust_decimal = { version = "1", default-features = false, optional = true }
 bigdecimal = { version = "0.3", default-features = false, optional = true }
 sea-orm-macros = { version = "0.10.3", path = "sea-orm-macros", default-features = false, optional = true }
-sea-query = { version = "0.28", features = ["thread-safe"] }
-sea-query-binder = { version = "0.3", default-features = false, optional = true }
+sea-query = { version = "0.28", git = "https://github.com/SeaQL/sea-query", branch = "get-values-from-update-statement", features = ["thread-safe"] }
+sea-query-binder = { version = "0.3", git = "https://github.com/SeaQL/sea-query", branch = "get-values-from-update-statement", default-features = false, optional = true }
```
tyt2y3, line 38:
```suggestion
sea-query = { version = "0.28.3", features = ["thread-safe"] }
sea-query-binder = { version = "0.3", default-features = false, optional = true }
```

src/executor/update.rs
```diff
@@ -29,8 +28,9 @@ where
         <A::Entity as EntityTrait>::Model: IntoActiveModel<A>,
         C: ConnectionTrait,
     {
-        // so that self is dropped before entering await
```
tyt2y3, line 32: :(

tyt2y3 (approved):

src/entity/active_model.rs
```diff
@@ -1423,8 +1423,8 @@ mod tests {
             vec![
                 Transaction::from_sql_and_values(
                     DbBackend::Postgres,
-                    r#"UPDATE "fruit" SET  WHERE "fruit"."id" = $1 RETURNING "id", "name", "cake_id""#,
```
tyt2y3, line 1426: Didn't know we put such a test case here

billy1624: Yep. That's bad :(

billy1624: It's a red flag that no one notice

PR #1385, opened 2023-01-10
changed DatabaseConnection to DatabaseTransaction for endpoints in poem example

TheCataliasTNT2k:
## PR Info
I swapped the DatabaseConnection to DatabaseTransaction for all endpoints in the poem example, to show, how this can be done.
I do not know, if this is the way to go, but at least it is a way and it works.

### Problems and possible Solutions with my way to go:
- I need to put the database connection into `AppState` even though no endpoint uses it (I found no other way to get it into the async closure within `around`)
- The `.data()` needs to be after the `.around()`, otherwise the db connection can not be found
- I could not find a way to supply the endpoints with a non-cloneable transaction. I need the object to provide it for the endpoint, and I need it later for `commit` or `rollback`.
So if the transaction gets cloned within the endpoint, you might run into issues.
Possible solutions: Providing only a (mutable) reference to the endpoint.
Or: Getting the object back from poem after the endpoint finishes.
But Poem does not support either, as far as I see.

I will leave this open for now, maybe some of you find better ways to do things.
Link to issue in poem repo:

### DIFF
--- a/examples/poem_example/api/src/lib.rs
+++ b/examples/poem_example/api/src/lib.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::sync::Arc;
 
 use entity::post;
 use migration::{Migrator, MigratorTrait};
@@ -7,9 +8,9 @@ use poem::error::InternalServerError;
 use poem::http::StatusCode;
 use poem::listener::TcpListener;
 use poem::web::{Data, Form, Html, Path, Query};
-use poem::{get, handler, post, EndpointExt, Error, IntoResponse, Result, Route, Server};
+use poem::{get, handler, post, EndpointExt, Error, IntoResponse, Result, Route, Server, Endpoint};
 use poem_example_core::{
-    sea_orm::{Database, DatabaseConnection},
+    sea_orm::{Database, DatabaseConnection, DatabaseTransaction, TransactionTrait},
     Mutation as MutationCore, Query as QueryCore,
 };
 use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


examples/poem_example/api/src/lib.rs
```diff
@@ -30,24 +31,22 @@ struct Params {
 }
 
 #[handler]
-async fn create(state: Data<&AppState>, form: Form<post::Model>) -> Result<impl IntoResponse> {
+async fn create(txn: Data<&Arc<DatabaseTransaction>>, form: Form<post::Model>) -> Result<impl IntoResponse> {
     let form = form.0;
-    let conn = &state.conn;
 
-    MutationCore::create_post(conn, form)
+    MutationCore::create_post(txn.0, form)
```
billy1624, line 37: Hey @TheCataliasTNT2k, I think you can start a transaction with 
```rs
let conn = &state.conn;

MutationCore::create_post(&txn, form)
```

examples/poem_example/core/src/mutation.rs
```diff
@@ -5,7 +5,7 @@ pub struct Mutation;
 
 impl Mutation {
     pub async fn create_post(
-        db: &DbConn,
+        db: &DatabaseTransaction,
         form_data: post::Model,
     ) -> Result<post::ActiveModel, DbErr> {
```
billy1624, line 10: Then, change the `db` to takes any types implemented `ConnectionTrait`:
```rs
pub async fn create_post<C: ConnectionTrait>(
    db: &C,
    form_data: post::Model,
)
```

examples/poem_example/api/src/lib.rs
TheCataliasTNT2k: Yes, this is possible, but the idea was, that you do not need to worry about committing or anything like that.
Use the transaction as you want, 
return OK for commit
return Error for rollback

examples/poem_example/core/src/mutation.rs
TheCataliasTNT2k: Would be an idea, yes. But I wanted to show, that you can use transactions just like the normal connections.

PR #1392, opened 2023-01-13
model_extra_derives to be also added to coresponding enums

gocmarek:
_As first thanks a lot for the job you are doing to deliver us this great piece of software!_

As i see you have added extra option for sea-orm-cli "model_extra_derives" which is realy helping a lot and saving time - we don't need to add extra implementation of our code to have all needed derives. 

...but there is a little problem because this option is not adding those derives to coresponding enums which is normally needed to be able to do some deeper mapping etc.. 

**Please marge this PR** to add this little gap, I think other people also need it.

### DIFF
--- a/sea-orm-codegen/src/entity/active_enum.rs
+++ b/sea-orm-codegen/src/entity/active_enum.rs
@@ -12,7 +12,7 @@ pub struct ActiveEnum {
 }
 
 impl ActiveEnum {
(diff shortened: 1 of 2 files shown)


billy1624 (commented):
Hey @gocmarek, thanks for reporting this problem and file a PR for it. I think we need extra documentation for this breaking changes. We should edit the help message of `sea-orm-cli generate entity` command and mention the `--model-extra-derives` string will also be applied to the generated active enums.

robertazzopardi:
Any update on this @gocmarek, would like to see this landed too!

PR #1394, opened 2023-01-13
Don't call auto_increment for columns where it's not set

### LINKED ISSUE #1357: Hitting 'negative last_insert_rowid' panic with Sqlite
## Description

I have a table with a primary ID that is a hash, stored as a signed integer (since unsigned 64 bit is not supported). Auto-increment is turned off for that table.
When storing a new element, if the hash as signed is negative, then sea-orm panics on that line: https://github.com/SeaQL/sea-orm/blob/master/src/executor/execute.rs#L43

## Steps to Reproduce

- Create an SQlite table with an i64 primary key: https://github.com/nitnelave/lldap/blob/c64d32e2c0cc698c61ca40e8364ab0cd7934cd46/server/src/domain/model/jwt_refresh_storage.rs#L12
- Insert a negative value.

### Expected Behavior

Since auto-increment is turned off for that table, `last_insert_id` should probably not be called to avoid a panic.

### Actual Behavior

Panic.

### Reproduces How Often

Always.

## Versions

Using sea-orm 0.10.3 (no other version override). Running on alpine, using the bundled Sqlite.

## Additional Information

The query, for reference: https://github.com/nitnelave/lldap/blob/c64d32e2c0cc698c61ca40e8364ab0cd7934cd46/server/src/infra/sql_backend_handler.rs#L49

nitnelave:
Fixes #1357

### DIFF
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -137,28 +137,30 @@ where
 {
     type PrimaryKey<A> = <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey;
     type ValueTypeOf<A> = <PrimaryKey<A> as PrimaryKeyTrait>::ValueType;
-    let last_insert_id_opt = match db.support_returning() {
-        true => {
-            let cols = PrimaryKey::<A>::iter()
-                .map(|col| col.to_string())
-                .collect::<Vec<_>>();
-            let rows = db.query_all(statement).await?;
-            let res = rows.last().ok_or_else(|| {
-                DbErr::RecordNotInserted("None of the records are being inserted".to_owned())
-            })?;
-            res.try_get_many("", cols.as_ref()).ok()
-        }
-        false => {
-            let last_insert_id = db.execute(statement).await?.last_insert_id();
-            ValueTypeOf::<A>::try_from_u64(last_insert_id).ok()
-        }
-    };
     let last_insert_id = match primary_key {
         Some(value_tuple) => FromValueTuple::from_value_tuple(value_tuple),
-        None => match last_insert_id_opt {
-            Some(last_insert_id) => last_insert_id,
-            None => return Err(DbErr::UnpackInsertId),
-        },
+        None => {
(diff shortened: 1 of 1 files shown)


nitnelave:
I didn't find a simple way to add a test, since the bakery schema doesn't have any model with an integer, non-auto-increment primary key and it doesn't seem easy to add a custom schema for a test. If you have ideas or hints on how to add the tests, I can do that.

tyt2y3:
Umm no clear reason why the tests failed

tyt2y3:
I see why the test fail here.

src/executor/insert.rs
```diff
@@ -137,28 +137,30 @@ where
...
-                .map(|col| col.to_string())
-                .collect::<Vec<_>>();
-            let rows = db.query_all(statement).await?;
-            let res = rows.last().ok_or_else(|| {
-                DbErr::RecordNotInserted("None of the records are being inserted".to_owned())
-            })?;
-            res.try_get_many("", cols.as_ref()).ok()
-        }
-        false => {
-            let last_insert_id = db.execute(statement).await?.last_insert_id();
```
tyt2y3, line 152: We should still execute the statement here, but just not trying to call `last_insert_id` from it unless it's really needed.

tyt2y3:
@billy1624 can you revise the implementation?

src/executor/insert.rs
```diff
@@ -137,28 +137,30 @@ where
...
-            ValueTypeOf::<A>::try_from_u64(last_insert_id).ok()
```
tyt2y3, line 153: We should also defer this `try_from_u64`

src/executor/insert.rs
```diff
@@ -137,28 +137,30 @@ where
...
-        }
-    };
     let last_insert_id = match primary_key {
         Some(value_tuple) => FromValueTuple::from_value_tuple(value_tuple),
```
tyt2y3, line 141: To make it consistent with `support_returning = true` path

tyt2y3:
Close in favour of #1403

PR #1403, opened 2023-01-18
Don't call last_insert_id if not needed

### LINKED ISSUE #1357: Hitting 'negative last_insert_rowid' panic with Sqlite

tyt2y3:
Fixes #1357

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -58,8 +58,8 @@ pub enum DbErr {
     /// None of the records are being inserted into the database,
     /// if you insert with upsert expression that means
     /// all of them conflict with existing records in the database
-    #[error("RecordNotInserted Error: {0}")]
-    RecordNotInserted(String),
+    #[error("None of the records are being inserted")]
+    RecordNotInserted,
 }
 
 /// Runtime error
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -138,36 +138,33 @@ where
     type PrimaryKey<A> = <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey;
     type ValueTypeOf<A> = <PrimaryKey<A> as PrimaryKeyTrait>::ValueType;
 
-    let last_insert_id = if db.support_returning() {
-        let mut rows = db.query_all(statement).await?;
-        let row = match rows.pop() {
-            Some(row) => row,
-            None => {
-                return Err(DbErr::RecordNotInserted(
-                    "None of the records are being inserted".to_owned(),
-                ))
-            }
-        };
-        match primary_key {
-            Some(value_tuple) => Ok(FromValueTuple::from_value_tuple(value_tuple)),
-            None => {
-                let cols = PrimaryKey::<A>::iter()
-                    .map(|col| col.to_string())
-                    .collect::<Vec<_>>();
-                row.try_get_many("", cols.as_ref())
-            }
+    let last_insert_id = match (primary_key, db.support_returning()) {
+        (Some(value_tuple), _) => {
+            db.execute(statement).await?;
+            FromValueTuple::from_value_tuple(value_tuple)
         }
-    } else {
-        let res = db.execute(statement).await?;
-        match primary_key {
-            Some(value_tuple) => Ok(FromValueTuple::from_value_tuple(value_tuple)),
-            None => {
-                let last_insert_id = res.last_insert_id();
-                ValueTypeOf::<A>::try_from_u64(last_insert_id)
-            }
+        (None, true) => {
+            let mut rows = db.query_all(statement).await?;
+            let row = match rows.pop() {
+                Some(row) => row,
+                None => {
+                    return Err(DbErr::RecordNotInserted(
+                        "None of the records are being inserted".to_owned(),
+                    ))
+                }
+            };
+            let cols = PrimaryKey::<A>::iter()
+                .map(|col| col.to_string())
+                .collect::<Vec<_>>();
+            row.try_get_many("", cols.as_ref())
+                .map_err(|_| DbErr::UnpackInsertId)?
         }
-    }
-    .map_err(|_| DbErr::UnpackInsertId)?;
+        (None, false) => {
+            let res = db.execute(statement).await?;
+            let last_insert_id = res.last_insert_id();
+            ValueTypeOf::<A>::try_from_u64(last_insert_id).map_err(|_| DbErr::UnpackInsertId)?
+        }
+    };
 
     Ok(InsertResult { last_insert_id })
 }
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -140,18 +140,17 @@ where
 
     let last_insert_id = match (primary_key, db.support_returning()) {
         (Some(value_tuple), _) => {
-            db.execute(statement).await?;
+            let res = db.execute(statement).await?;
+            if res.rows_affected() == 0 {
+                return Err(DbErr::RecordNotInserted);
+            }
             FromValueTuple::from_value_tuple(value_tuple)
         }
         (None, true) => {
             let mut rows = db.query_all(statement).await?;
             let row = match rows.pop() {
                 Some(row) => row,
-                None => {
-                    return Err(DbErr::RecordNotInserted(
-                        "None of the records are being inserted".to_owned(),
-                    ))
-                }
+                None => return Err(DbErr::RecordNotInserted),
             };
             let cols = PrimaryKey::<A>::iter()
                 .map(|col| col.to_string())
@@ -161,6 +160,9 @@ where
         }
         (None, false) => {
             let res = db.execute(statement).await?;
+            if res.rows_affected() == 0 {
+                return Err(DbErr::RecordNotInserted);
+            }
             let last_insert_id = res.last_insert_id();
             ValueTypeOf::<A>::try_from_u64(last_insert_id).map_err(|_| DbErr::UnpackInsertId)?
         }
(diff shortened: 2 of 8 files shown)


src/executor/insert.rs
```diff
@@ -137,29 +138,49 @@ where
...
+    }
+
+    let insert_result = if db.support_returning() {
+        let mut rows = db.query_all(statement).await?;
+        if rows.is_empty() {
+            return Err(DbErr::RecordNotInserted(
+                "None of the records are being inserted".to_owned(),
+            ));
         }
+        QueryOrExecResult::Query(rows.remove(rows.len() - 1))
```
nitnelave, line 154: `rows.pop().unwrap()`?

src/executor/insert.rs
```diff
@@ -137,29 +138,49 @@ where
...
-            let res = rows.last().ok_or_else(|| {
-                DbErr::RecordNotInserted("None of the records are being inserted".to_owned())
-            })?;
-            res.try_get_many("", cols.as_ref()).ok()
-        }
-        false => {
-            let last_insert_id = db.execute(statement).await?.last_insert_id();
-            ValueTypeOf::<A>::try_from_u64(last_insert_id).ok()
+
+    enum QueryOrExecResult {
```
nitnelave, line 142: It's quite awkward to have an enum where you already know the type it has in each branch. Can we refactor otherwise? To get rid of the `unreachable`.

src/executor/insert.rs
nitnelave: Or even better, match `rows.pop()` with `None` -> `RecordNotInserted`

src/executor/insert.rs
```diff
@@ -137,29 +138,49 @@ where
...
     };
+
     let last_insert_id = match primary_key {
-        Some(value_tuple) => FromValueTuple::from_value_tuple(value_tuple),
-        None => match last_insert_id_opt {
-            Some(last_insert_id) => last_insert_id,
-            None => return Err(DbErr::UnpackInsertId),
-        },
-    };
+        Some(value_tuple) => Ok(FromValueTuple::from_value_tuple(value_tuple)),
```
nitnelave, line 160: It seems to be we just have to add a `db.execute(statement).await?` here rather than introduce an entirely new type.

src/executor/insert.rs
```diff
@@ -137,29 +138,49 @@ where
...
+        None => {
+            if db.support_returning() {
+                match insert_result {
+                    QueryOrExecResult::Query(row) => {
+                        let cols = PrimaryKey::<A>::iter()
+                            .map(|col| col.to_string())
```
nitnelave, line 166: That might be a bigger change, but I feel that `try_get_many` should accept a slice of `AsRef<str>` and `IdenStatic` should implement `AsRef<str>` so we don't have to do this whole conversion at all.

nitnelave (changes requested):

tyt2y3:
I don't see issues/1357 being run, but I assure it works on my machine(TM)

src/executor/insert.rs
tyt2y3: I agree, but that might be another PR

src/executor/insert.rs
tyt2y3: No. It has to execute whether Some or None.

src/executor/insert.rs
tyt2y3: I think the alternative is that we split the implementation.

src/executor/insert.rs
nitnelave: So, if we add the execute statement in the `Some` branch, we can inline the top of the function in the `None` branch, further reducing the amount of work done in case we provide the primary key. Each branch would have its own singular await point.

What's the rationale for reducing the number of await points?

src/executor/insert.rs
tyt2y3: Right

tyt2y3:
Refactored. How about now?

src/executor/insert.rs
```diff
@@ -137,29 +137,38 @@ where
...
-            let cols = PrimaryKey::<A>::iter()
-                .map(|col| col.to_string())
-                .collect::<Vec<_>>();
-            let rows = db.query_all(statement).await?;
-            let res = rows.last().ok_or_else(|| {
-                DbErr::RecordNotInserted("None of the records are being inserted".to_owned())
-            })?;
-            res.try_get_many("", cols.as_ref()).ok()
+
+    let last_insert_id = if db.support_returning() {
```
nitnelave, line 141: That's much better! Do we absolutely care about `RecordNotInserted` if the primary key is provided? How can we detect it if the db doesn't support returning?

If we don't especially care about it, we could rewrite it as : 
```rust
match (primary_key, db.support_returning()) {
  (Some(value_tuple), _) => {db.execute(); Ok(FromValueTuple)},
  (None, true) =>{db.query_all() ... },
  (None, false) => {res = db.execute(); res.last_insert_id()},
}
```

WDYT?

tyt2y3: Yes, I think we care. Because that's the case when we inserted a duplicate record?

> How can we detect it if the db doesn't support returning

Now it's a compile-time hardcoded value, with SQLite being false. Ideally, SQLite support should be determined by its version detected runtime.

nitnelave: No, I mean how can we detect that the record was not inserted if the db doesn't support returning?

tyt2y3: Umm, I see what you mean

nitnelave: Could we use https://stackoverflow.com/questions/6551903/getting-the-number-of-affected-rows-for-a-sqlite-statement-using-the-c-api ?

nitnelave: For sqlite, by calling `sqlite3_changes()` (or by selecting `changes()` in the query), we can get the number of rows affected. With that, we can see if it's 0 (didn't insert anything) or >0 (inserted at least one thing). That wouldn't catch the case where we insert >1 value but only some failed, but for single value inserts, it would already be helpful.

tyt2y3: I think the code as of "Refactor" can be released in 0.10 and further changes should go into 0.11

nitnelave: Sounds good. Just FYI, I tested `select changes();` on SQlite3, it seems to do what we want: after a single insert, it'll return 1 if the insert was successful, and 0 otherwise. And it's scoped to the DB connection as well, so it should be usable even in async contexts.

But anyway, inserting a single duplicate row returns an error, so probably the `await?` should trigger a return, no?

nitnelave (approved):

src/executor/insert.rs
tyt2y3: Yes. It should be a UNIQUE CONSTRAINT ERROR unless if we specified ON CONFLICT DO NOTHING.

src/executor/insert.rs
```diff
@@ -137,29 +137,37 @@ where
 {
     type PrimaryKey<A> = <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey;
     type ValueTypeOf<A> = <PrimaryKey<A> as PrimaryKeyTrait>::ValueType;
-    let last_insert_id_opt = match db.support_returning() {
-        true => {
+
+    let last_insert_id = match (primary_key, db.support_returning()) {
+        (Some(value_tuple), _) => {
+            let res = db.execute(statement).await?;
+            if res.rows_affected() == 0 {
```
nitnelave, line 144: oh, if we have that, that's great!

nitnelave (approved):
Looks great!

src/executor/insert.rs
```diff
@@ -137,29 +137,37 @@ where
 {
     type PrimaryKey<A> = <<A as ActiveModelTrait>::Entity as EntityTrait>::PrimaryKey;
     type ValueTypeOf<A> = <PrimaryKey<A> as PrimaryKeyTrait>::ValueType;
-    let last_insert_id_opt = match db.support_returning() {
-        true => {
+
+    let last_insert_id = match (primary_key, db.support_returning()) {
```
billy1624, line 141: Nice rewrite!!

src/error.rs
```diff
@@ -58,8 +58,8 @@ pub enum DbErr {
     /// None of the records are being inserted into the database,
     /// if you insert with upsert expression that means
     /// all of them conflict with existing records in the database
-    #[error("RecordNotInserted Error: {0}")]
-    RecordNotInserted(String),
+    #[error("None of the records are being inserted")]
+    RecordNotInserted,
```
billy1624, line 62: I think this is a breaking change that should goes into 0.11.x

billy1624 (approved):

src/error.rs
tyt2y3: It's also not in 0.10 actually

Issue #1377, opened 2023-01-09
How to "TRUNCATE table_name" in sea-orm.


billy1624:
### Discussed in https://github.com/SeaQL/sea-orm/discussions/1375

<div type='discussions-op-text'>

<sup>Originally posted by **baoyachi** January  5, 2023</sup>
with reference link: https://www.postgresql.org/docs/current/sql-truncate.html</div>

baoyachi:
LGTM

baoyachi:
”So it's trivial to bring it into SeaORM.“

 I don't think it's trivial. Because I need use `TRUNCATE  table_name` before the unit test started

billy1624:
Closing this now, as discussed at https://github.com/SeaQL/sea-orm/pull/1409#issuecomment-1403266432

PR #1409, opened 2023-01-20
Support truncate

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1377

## New Features

- [x] Truncate a table in the database; A `TRUNCATE` statement will be executed on Postgres and MySQL; A `DELETE` statement will be executed on SQLite
```rs
assert_eq!(Entity::find().all(db).await?.len(), 3);

assert_eq!(Entity::truncate().exec(db).await?.rows_affected, 3);

assert_eq!(Entity::find().all(db).await?.len(), 0);
```

### DIFF
--- a/src/entity/base_entity.rs
+++ b/src/entity/base_entity.rs
@@ -1,7 +1,7 @@
 use crate::{
     ActiveModelTrait, ColumnTrait, Delete, DeleteMany, DeleteOne, FromQueryResult, Insert,
     ModelTrait, PrimaryKeyToColumn, PrimaryKeyTrait, QueryFilter, Related, RelationBuilder,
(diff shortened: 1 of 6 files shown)


billy1624:
> 1. TRUNCATE is a DDL, so it'd naturally fit into sea-schema / sea-orm-migration, and the use case seems to be inside migrations as well

Got it!

> 2. There could be an `Entity::clear()` or `delete_all()` method that simply executes a `DELETE WITHOUT FILTER`

Then I don't think it'd be necessary. `delete_many` does that already.

```rs
fruit::Entity::delete_many()
    .exec(db)
    .await?;
```

Issue #1372, opened 2023-01-08
Allow limit to take -1


112RG:
## Motivation
When writing web API there is often a time when you would like to limit the items returned by a request via query params. But sometimes these can be not set so we want to return all in that case
Currently in sea-orm when you provide a .limit() there is no way of saying just return all. Having something like -1 be a valid option taken and would then just ignore the limit all together would help clean up code where you have 

```rust
pub async fn get_latest_albums(
    db: &DatabaseConnection,
    size: Option<u64>,
) -> anyhow::Result<Vec<entity::album::Model>> {
    Ok(entity::album::Entity::find()
        .order_by_desc(entity::album::Column::CreatedAt)
        .limit(size.unwrap_or(50))
        .all(db)
        .await?)
}
```

Currently, I would have to setup crazy high default for the unwrap. Or have another function that doesn't have a .limit()

## Proposed Solutions
Allow .limit() to take None or -1 and return all items

JonasCir:
If this get's implemented, I think we should use `Option::None` to represent this, using magic values has its downsides.

112RG:
Yes thinking about it ``Option:None`` seems better. My original plan was 
```rust  
.limit(size.unwrap_or(u64::MAX))
```

But it just throws a error ``called `Result::unwrap()` on an `Err` value: TryFromIntError(())'``

dragonnn:
I don't think any of that is a good idea, why not just do it like that:
```rust
let mut query = entity::album::Entity::find()
        .order_by_desc(entity::album::Column::CreatedAt);
if let Some(size) = size {
     Ok(query.limit(size).all(db).await?)
} else {
    Ok(query.all(db).await?) //I think all needs to be replaced with find_all? Need to check
}
```
That should work.

112RG:
> I don't think any of that is a good idea, why not just do it like that:
> 
> ```rust
> let mut query = entity::album::Entity::find()
>         .order_by_desc(entity::album::Column::CreatedAt);
> if let Some(size) = size {
>     Ok(query.limit(size).all(db).await?)
> } else {
>     Ok(query.all(db).await?) //I think all needs to be replaced with find_all? Need to check
> }
> ```
> 
> That should work.

I know that will that is what I currently have but then I need if else blocks everywhere.
Adding that into code like this 

```rust
pub async fn get_albums(
    db: &DatabaseConnection,
    size: Option<u64>,
    sort: Option<String>,
[...]

PR #1410, opened 2023-01-20
Allows offset & limit to take `Option<u64>`

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1372

## New Features

- [x] `QuerySelect::offset()` and `QuerySelect::limit()` take either `Option<64>` or `u64`

## Breaking Changes

- [x] Changed the parameter of `QuerySelect::offset()` and `QuerySelect::limit()` to `OptionalU64`

### DIFF
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -190,6 +190,15 @@ pub trait QuerySelect: Sized {
     ///         .to_string(),
     ///     "SELECT `cake`.`id`, `cake`.`name` FROM `cake`"
     /// );
+    ///
+    /// assert_eq!(
+    ///     cake::Entity::find()
+    ///         .offset(10)
+    ///         .offset(None) // This will reset the offset
+    ///         .build(DbBackend::MySql)
+    ///         .to_string(),
+    ///     "SELECT `cake`.`id`, `cake`.`name` FROM `cake`"
+    /// );
     /// ```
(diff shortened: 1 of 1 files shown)


tyt2y3:
I don't like this API, something like https://docs.rs/sea-query/latest/sea_query/query/struct.SelectStatement.html#method.conditions would serve a more general usecase.

In any case I think `fn foo<T>(param: T) where T: Into<Option<u64>>` will do the trick better.

billy1624:
Hey @tyt2y3, check check https://github.com/SeaQL/sea-orm/pull/1415 for the `maybe` API

billy1624:
I still think making `limit` and `offset` to accept `Into<Option<u64>>` is worth doing. @tyt2y3  please advice

tyt2y3:
I think the API behaviour here should be if None is passed in, it should clear the limit/offset instead of no-op.

billy1624:
> I think the API behaviour here should be if None is passed in, it should clear the limit/offset instead of no-op.

Yes, you're right! How could I miss that. Let me add that in.

tyt2y3:
That's why I said the original intention seems to be better handled by `apply_if` API instead of this one

billy1624:
Still hesitate on publishing this API?

tyt2y3 (approved):

PR #1412, opened 2023-01-21
Proc macro parsing with darling and test

### LINKED ISSUE #394: Produce compile errors when unexpected proc_macros attributes are provided
Related to #391

Diwakar-Gupta:
# Still Working need feedback
## PR Info

Fixes 
Problem discussed at https://github.com/SeaQL/sea-orm/issues/394#issuecomment-1396947927. 

## Related
https://github.com/SeaQL/sea-orm/pull/1560

## Tasks
- [x] Replace manually parsing with [`darling`](https://github.com/TedDriggs/darling)
- [x] Write test for parser

## File Changed
1. `sea-orm-macros/Cargo.toml` added `darling` dependency
2. `sea-orm-macros/src/derives/model.rs` modified parser and added test


Diwakar-Gupta:
One thing that i did not understand is
https://github.com/SeaQL/sea-orm/blob/9f6b5664eba59c5eafeb6ce91e8946af47983fdd/sea-orm-macros/src/derives/model.rs#L34-L39

this code is trying to extract attribute from one of this
https://github.com/SeaQL/sea-orm/blob/9f6b5664eba59c5eafeb6ce91e8946af47983fdd/sea-orm-macros/src/attributes.rs#L5-L15

for the case of `derive/model.rs` it resulted in `Entity` while the list of attributes  is `[Clone, Debug, PartialEq, DeriveEntityModel, table_name]` is `DeriveEntityModel` mapped to `Entity`?

billy1624 (commented):
Hey @Diwakar-Gupta, thanks for the work!!

> 1. the test needs to be done in a different way

I think the test is on the right track! :)

> 2. we need to change structure struct `DeriveModel`

Yes, would be better if we replace `DeriveModel` with `DeriveModelDarling`. As `DeriveModelDarling` already have all the info of `DeriveModel`.

Btw... a small nitpick. Should we name the derive data struct without the `Daring` prefix? For example, naming `DeriveModelDarling` as `DeriveModel`.

> 3. Code generation need's to be changed

I'm not sure what you mean here. Could you please clarify?

> 4. if this is going in right direction or not.

Good POC! Was in the right direction!
Appreciated the help :)

Diwakar-Gupta:
Thanks for the feedback.

I will look after replacing `DeriveModel` with `DeriveModelDarling`.

tyt2y3:
Wow, interesting. I have never used darling. Seems like it can make the code much cleaner!

billy1624:
Yep, wanna get rid of the parsing mess ASAP loll

Diwakar-Gupta:
1. Previously `DeriveModel` had processed data, removing it has introduced some extra processing in `impl_from_query_result` and  `impl_model_trait` function. But it is manageable atleast in my openion.
2. **struct is only supported** this test is done by `darling` now
```
#[darling(supports(struct_named))]
struct DeriveModel {
```

3. Now Entity ident is calculated differently

```
struct DeriveModel {
    #[darling(skip)]
    entity_ident: Option<Ident>,
}
...
        let sea_attr = derive_attr::SeaOrm::try_from_attributes(&input.attrs)
            .map_err(Error::Syn)?
            .unwrap_or_default();

        receiver.entity_ident = Some(sea_attr.entity.unwrap_or_else(|| format_ident!("Entity")));
```
4. Column Ident is converted to a function
```
impl DeriveField {
    pub fn get_column_ident(&self) -> Ident {
        match &self.enum_name {
            Some(name) => format_ident!("{}", &name),
            _ => {
                let ident = self.ident.as_ref().unwrap().to_string();
                let ident = trim_starting_raw_identifier(ident).to_camel_case();
                let ident = escape_rust_keyword(ident);
                let ident = format_ident!("{}", &ident);
                ident
            }
```
We can get rid of this if we can specify the default expression in `struct DeriveField`, but I am not sure how to do this
```
struct DeriveField {
    #[default_expr= get_column_ident]
    column_ident: Option<String>,
```
5. Test is yet to be written

billy1624:
Hey @Diwakar-Gupta, I think you leave a comment here yesterday night but it's now gone. Anything I could help?

sea-orm-macros/src/derives/model.rs
```diff
@@ -1,95 +1,74 @@
...
+    #[darling(default)]
+    ignore: bool,
+    #[darling(default)]
+    enum_name: Option<String>,
+    #[darling(skip)]
+    column_ident: Option<Ident>,
+}
+
+#[derive(Debug, FromDeriveInput)]
+#[darling(supports(struct_named))]
```
Diwakar-Gupta, line 29: this will make sure that we only support `struct` and not `enum`.

so I replaced  `Error::InputNotStruct` with `Parsing(darling::Error)`

and changed handling  in `pub fn expand_derive_model` respectively.

sea-orm-macros/src/derives/model.rs
```diff
@@ -1,95 +1,74 @@
...
     fn new(input: syn::DeriveInput) -> Result<Self, Error> {
-        let fields = match input.data {
-            syn::Data::Struct(syn::DataStruct {
-                fields: syn::Fields::Named(syn::FieldsNamed { named, .. }),
-                ..
-            }) => named,
-            _ => return Err(Error::InputNotStruct),
+        let mut receiver = match DeriveModel::from_derive_input(&input) {
+            Ok(model) => model,
+            Err(error) => return Err(Error::Parsing(error)),
```
Diwakar-Gupta, line 41: Currently I cant match if this error is because of  deriving in `enum` because some error related enum is private in `darling` repo I have raised issue for that. https://github.com/TedDriggs/darling/issues/222#issue-1559510121

sea-orm-macros/src/derives/model.rs
```diff
@@ -1,95 +1,74 @@
...
+#[derive(Debug, FromField)]
+#[darling(attributes(sea_orm), allow_unknown_fields = true)]
+struct DeriveField {
+    ident: Option<Ident>,
+    #[darling(default)]
+    ignore: bool,
+    #[darling(default)]
+    enum_name: Option<String>,
+    #[darling(skip)]
+    column_ident: Option<Ident>,
```
Diwakar-Gupta, line 25: `darling` cant parse column_ident because it has out own logic so made it Option and then calculated if in `fn new`.

sea-orm-macros/src/derives/model.rs
```diff
@@ -185,13 +165,127 @@ impl DeriveModel {
...
 /// Method to derive an ActiveModel
 pub fn expand_derive_model(input: syn::DeriveInput) -> syn::Result<TokenStream> {
-    let ident_span = input.ident.span();
-
     match DeriveModel::new(input) {
         Ok(model) => model.expand(),
-        Err(Error::InputNotStruct) => Ok(quote_spanned! {
-            ident_span => compile_error!("you can only derive DeriveModel on structs");
-        }),
+        Err(Error::Parsing(err)) => Ok(darling::Error::write_errors(err)),
```
Diwakar-Gupta, line 170: This highlights the ident causing error like it was done previously.

Diwakar-Gupta:
The comment was a report of changes done; Then I did some more changes and it became irrelevant.
Here is the latest summary.

All this are also commented in code.

1. Previously `DeriveModel` had processed data, removing it has introduced some extra processing in `impl_from_query_result` and  `impl_model_trait` functions. But it is manageable at least in my opinion.

2. **struct is only supported** this is ensured by `darling` now, and default darling error message is shown to user for now. cant give custom error message unless `darling` makes some necessary module public.

sea-orm-macros/src/derives/model.rs
```diff
@@ -1,95 +1,74 @@
...
-use std::iter::FromIterator;
-use syn::{punctuated::Punctuated, token::Comma, Ident, Lit, Meta};
+use quote::{format_ident, quote};
+use syn::Ident;
 
 enum Error {
-    InputNotStruct,
+    Parsing(darling::Error),
     Syn(syn::Error),
 }
```
Diwakar-Gupta, line 14: this will make sure that we only support `struct` and not `enum`.

so I replaced  `Error::InputNotStruct` with `Parsing(darling::Error)`

and changed handling  in `pub fn expand_derive_model` respectively.

sea-orm-macros/src/derives/model.rs
Diwakar-Gupta, line 29: this will make sure that we only support `struct` and not `enum`.

so I replaced  `Error::InputNotStruct` with `Parsing(darling::Error)`

and changed handling  in `pub fn expand_derive_model` respectively.

sea-orm-macros/src/derives/model.rs
Diwakar-Gupta: Resolved with `daling` version `0.14.3` new error message is ```Unsupported shape `enum`. Expected struct with named fields.```

billy1624 (commented):
Hey @Diwakar-Gupta, I saw your implementation then I decided to try it before leaving comments. My experiment result: https://github.com/SeaQL/sea-orm/pull/1560

Feel free to copy the useful bits to your PR. Hopefully we can combine the "good" into single PR.

Diwakar-Gupta:
Thanks @billy1624 I will study the reference you have given.

Diwakar-Gupta:
hey @billy1624 i have used your pr https://github.com/SeaQL/sea-orm/pull/1560 and fixed the merge conflict.

Issue #1364, opened 2023-01-04
Composite primary key with enum causes generated entity to fail to compile


Pure-Peace:
## Description

Composite primary key with enum causes generated entity to fail to compile.

### Entities generated from the database:

```rust
/* `sea_orm_active_enums.rs */

//! `SeaORM` Entity. Generated by sea-orm-codegen 0.10.5

#[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
#[sea_orm(
    rs_type = "String",
    db_type = "Enum",
    enum_name = "version"
)]
pub enum Version {
    #[sea_orm(string_value = "v1")]
    V1,
    #[sea_orm(string_value = "v2")]
    V2,
}
```

```rust
/* `report.rs */

//! `SeaORM` Entity. Generated by sea-orm-codegen 0.10.5

use super::sea_orm_active_enums::Version;
use sea_orm::entity::prelude::*;

#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
#[sea_orm(table_name = "report")]
pub struct Model {
    #[sea_orm(primary_key, auto_increment = false)]
    pub id: i64,
    #[sea_orm(primary_key, auto_increment = false)]
    pub version: Version
}

```

### The generated entity fails to compile with the following error:

```bash
error[E0277]: the trait bound `sea_orm_active_enums::Version: TryFromU64` is not satisfied
  --> demo.rs
   |
   | #[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
   |                                   ^^^^^^^^^^^^^^^^^ the trait `TryFromU64` is not implemented for `sea_orm_active_enums::Version`
   |
   = help: the following other types implement trait `TryFromU64`:
             (A, B)
             (A, B, C)
             (A, B, C, D)
             (A, B, C, D, E)
             (A, B, C, D, E, F)
             JsonValue
             Vec<u8>
             bool
           and 23 others
   = note: required for `(i64, sea_orm_active_enums::Version)` to implement `TryFromU64`
note: required by a bound in `sea_orm::PrimaryKeyTrait::ValueType`
  --> sea-orm-0.10.5\src\entity\primary_key.rs:49:11
   |
49 |         + TryFromU64;
   |           ^^^^^^^^^^ required by this bound in `sea_orm::PrimaryKeyTrait::ValueType`
[...]

PR #1414, opened 2023-01-21
Use `ActiveEnum` field as composite primary key

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1364

## Breaking Changes

- [x] Provided a blanket implementation of `TryFromU64` for all `ActiveEnum`
```rs
impl<T> TryFromU64 for T
where
    T: ActiveEnum,
{
    fn try_from_u64(_: u64) -> Result<Self, DbErr> {
        Err(DbErr::ConvertFromU64(
            "Fail to construct ActiveEnum from a u64, if your primary key consist of a ActiveEnum field, its auto increment should be set to false."
        ))
    }
}
```


tests/enum_primary_key_tests.rs
```diff
@@ -0,0 +1,104 @@
...
+            .unwrap()
+    );
+
+    let _ = ActiveModel {
+        category: Set(Some(Category::Big)),
+        color: Set(Some(Color::Black)),
+        ..model.into_active_model()
+    }
+    .save(db)
+    .await?;
```
tyt2y3, line 65: I don't understand the intention of this bit, and I think it SHOULD fail to insert instead of passing

billy1624: It does an update instead of insert.

```rs
// The `..model.into_active_model()` expands into

let _ = ActiveModel {
        category: Set(Some(Category::Big)),
        color: Set(Some(Color::Black)),
        id: Unchanged( ... )
    }
    .save(db)
    .await?;
```

tyt2y3: Right

tyt2y3: Can we also add an insert test case

billy1624: We can insert a row with duplicate ID. But I cannot assert the exact error, since it's database dependent. https://github.com/SeaQL/sea-orm/pull/1414/commits/66cffd4a2a3b5c1a45b26d46b6971eda0a3da8a6

tyt2y3: Yes, `is_err` would do

billy1624: Alright, good to go

tyt2y3 (approved):
Cool

PR #1415, opened 2023-01-21
`QueryTrait::maybe`: a general purpose query customizer

billy1624:
Idea from @tyt2y3 on https://github.com/SeaQL/sea-orm/pull/1410#issuecomment-1399181799

> I don't quite like this API, something like https://docs.rs/sea-query/latest/sea_query/query/struct.SelectStatement.html#method.conditions would serve a more general usecase, may be something like:
> 
> ```rust
> fn maybe<T, F>(&mut self, val: Option<T>, if_some: F)
> where F: FnOnce(&mut Self, v: T)
> ```
> 
> In any case I think `fn foo<T>(param: T) where T: Into<Option<u64>>` will do the trick better.

## New Features

- [x] Added `QueryTrait::maybe()` method to act on query with an `Option<T>` parameter

### DIFF
--- a/src/query/traits.rs
+++ b/src/query/traits.rs
@@ -24,7 +24,7 @@ pub trait QueryTrait {
         )
     }
 
-    /// Perform some operations on the [QueryTrait::QueryStatement] with the given `Option<T>` value
+    /// Apply a operation on the [QueryTrait::QueryStatement] if the given `Option<T>` is `Some(_)`
     ///
     /// # Example
     ///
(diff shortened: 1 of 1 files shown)


billy1624 (commented):
Nice one!! @tyt2y3

Approved~

src/query/traits.rs
```diff
@@ -23,4 +23,35 @@ pub trait QueryTrait {
             self.as_query().build_any(query_builder.as_ref()),
         )
     }
+
+    /// Apply a operation on the [QueryTrait::QueryStatement] if the given `Option<T>` is `Some(_)`
```
tyt2y3, line 27:
```suggestion
    /// Apply an operation on the [QueryTrait::QueryStatement] if the given `Option<T>` is `Some(_)`
```

tyt2y3 (approved):

PR #1420, opened 2023-01-24
Enable --universal-time by default

tyt2y3:
I think we can enable this by default on 0.11

Dependent: https://github.com/SeaQL/seaql.github.io/pull/84

### DIFF
--- a/sea-orm-cli/src/cli.rs
+++ b/sea-orm-cli/src/cli.rs
@@ -79,9 +79,20 @@ pub enum MigrateSubcommands {
             action,
             long,
             default_value = "true",
-            help = "Generate migration file based on Utc time instead of Local time"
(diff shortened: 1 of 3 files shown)


billy1624 (commented):
A good step forward! I think it's better to have a fallback option where the entity file is named with local time. Something like `--local-time`?

I'm more than happy to add that for you. What's your opinion?

tyt2y3:
I think `--universal-time=false` is always available

billy1624:
Flag in `clap` does not take value unless it's an option. I.e. `--universal-time=true` https://github.com/clap-rs/clap/issues/1649#issuecomment-661274943

I'll go and add a `--local-time` flag and make it mutual exclusive to each other.

billy1624 (approved):

PR #1425, opened 2023-01-26
RecordNotFound -> RecordNotUpdated

tyt2y3:
This makes the behaviour consistent with `RecordNotInserted`


billy1624 (approved):
Thanks!! @tyt2y3 It make sense to have a dedicated "not found" error only for update.

PR #1427, opened 2023-01-27
Cast primary key column inside returning expression of insert statement

billy1624:
## Bug Fixes

- [x] `DeriveEntityModel` with multiple `select_as` and `save_as` attributes will produce an error because excessive commas are being generated 

## Changes

- [x] Cast primary key column inside returning expression of insert statement

```sql
INSERT INTO "posts" ("id", "title", "text")
  VALUES (CAST(1 AS TEXT), 'News wrap up 2022', 'brbrbrrrbrbrbrr...')
  RETURNING CAST("id" AS INTEGER)
```


tyt2y3 (approved):
No idea who discovered this bug, but more tests are good

PR #1428, opened 2023-01-27
examples(axum): use tokio runtime

cemremengu:
Axum example should use `tokio` runtime instead of `async-std`


tyt2y3 (approved):
Ah! Can't believe we overlooked it.

Issue #1333, opened 2022-12-23
0.10.6: still issues with `conflicting implementations of trait`


frederikhors:
Since https://github.com/SeaQL/sea-orm/issues/1287 is closed I just tried 0.10.6: unfortunately the issue is still there:

```rust
impl Related<super::tenant::Entity> for Entity {
    fn to() -> RelationDef {
        super::house::Relation::Tenant.def()
    }
    fn via() -> Option<RelationDef> {
        Some(super::house::Relation::Player.def().rev())
    }
}

impl Related<super::tenant::Entity> for Entity {
    fn to() -> RelationDef {
        super::offer::Relation::Tenant.def()
    }
    fn via() -> Option<RelationDef> {
        Some(super::offer::Relation::Player.def().rev())
    }
}
```

with error:

```log
error[E0119]: conflicting implementations of trait `sea_orm::Related<generated::tenant::Entity>` for type `generated::player::Entity`
  --> src\entities\player.rs:65:1
   |
56 | impl Related<super::tenant::Entity> for Entity {
   | ---------------------------------------------- first implementation here
...
65 | impl Related<super::tenant::Entity> for Entity {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `generated::player::Entity`

error[E0119]: conflicting implementations of trait `sea_orm::Related<generated::player::Entity>` for type `generated::tenant::Entity`
   --> src\entities\tenant.rs:135:1
    |
117 | impl Related<super::player::Entity> for Entity {
    | ------------------------------------------------ first implementation here
...
135 | impl Related<super::player::Entity> for Entity {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `generated::tenant::Entity`
```

The entities in this case are:

```
Player {
    pub tenant_id: String,
}

House {
    pub tenant_id: String,
    pub player_id: String,
}

Offer {
    pub tenant_id: String,
    pub player_id: String,
}
```

I'll downgrade to 0.10.2 for now.

On other entites there is the same issue too (because of tenant_id).

billy1624:
Hey @frederikhors, sorry for the delay. I was not able to reproduce the error. Could you please provide a sample SQL like below?

```sql
create table if not exists player
(
    id integer
);

create unique index if not exists player_id_uindex
    on player (id);

[...]

PR #1435, opened 2023-01-30
Generate entity for duplicated many-to-many paths

billy1624:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1333

## Bug Fixes

- [x] The implementation of `Related<R>` with `via` and `to` methods will not be generated if there exists multiple paths via an intermediate table. Like in the schema defined below - Path 1. `users <-> users_votes <-> bills`, Path 2. `users <-> users_saved_bills <-> bills`

```sql
CREATE TABLE users
(
  id uuid  PRIMARY KEY  DEFAULT uuid_generate_v1mc(),
  email TEXT UNIQUE NOT NULL,
  ...
);
```
```sql
CREATE TABLE bills
(
  id uuid  PRIMARY KEY  DEFAULT uuid_generate_v1mc(),
  ...
);
```
```sql
CREATE TABLE users_votes
(
  user_id uuid REFERENCES users (id) ON UPDATE CASCADE ON DELETE CASCADE,
  bill_id uuid REFERENCES bills (id) ON UPDATE CASCADE ON DELETE CASCADE,
  vote boolean NOT NULL,
  CONSTRAINT users_bills_pkey PRIMARY KEY (user_id, bill_id)
);
```
```sql
CREATE TABLE users_saved_bills
(
  user_id uuid REFERENCES users (id) ON UPDATE CASCADE ON DELETE CASCADE,
  bill_id uuid REFERENCES bills (id) ON UPDATE CASCADE ON DELETE CASCADE,
  CONSTRAINT users_saved_bills_pkey PRIMARY KEY (user_id, bill_id)
);
```

> Adapted from https://github.com/SeaQL/sea-orm/issues/1333#issuecomment-1407791955

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -250,11 +250,15 @@ jobs:
         with:
           toolchain: stable
           components: rustfmt
-      - run: cargo fmt --manifest-path ${{ matrix.path }} --all -- --check
       - run: cargo build --manifest-path ${{ matrix.path }}
       - run: cargo test --manifest-path ${{ matrix.path }}
       - if: ${{ contains(matrix.path, 'core/Cargo.toml') }}
         run: cargo test --manifest-path ${{ matrix.path }} --features mock
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+          components: rustfmt
+      - run: cargo fmt --manifest-path ${{ matrix.path }} --all -- --check
 
   issues-matrix:
     name: Issues Matrix
--- a/sea-orm-codegen/Cargo.toml
+++ b/sea-orm-codegen/Cargo.toml
@@ -26,3 +26,4 @@ tracing = { version = "0.1", default-features = false, features = ["log"] }
 
 [dev-dependencies]
 pretty_assertions = { version = "0.7" }
+sea-orm = { path = "../", default-features = false, features = ["macros"] }
--- a/sea-orm-codegen/rustfmt.toml
+++ b/sea-orm-codegen/rustfmt.toml
@@ -1,4 +1,5 @@
 ignore = [
     "tests/compact/*.rs",
     "tests/expanded/*.rs",
+    "tests_cfg",
 ]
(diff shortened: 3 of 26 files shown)


.github/workflows/rust.yml
```diff
@@ -252,11 +252,15 @@ jobs:
...
-      - run: cargo fmt --manifest-path ${{ matrix.path }} --all -- --check
       - run: cargo build --manifest-path ${{ matrix.path }}
       - run: cargo test --manifest-path ${{ matrix.path }}
       - if: ${{ contains(matrix.path, 'core/Cargo.toml') }}
         run: cargo test --manifest-path ${{ matrix.path }} --features mock
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+          components: rustfmt
+      - run: cargo fmt --manifest-path ${{ matrix.path }} --all -- --check
```
tyt2y3, line 263: what are these bits about though

billy1624: Install nightly channel with rustfmt component. Then, run cargofmt with it.

billy1624: Only night's fmt respect `rustfmt.toml` config file that's why.

billy1624:
This PR includes a written-entity to generated-entity test cases. Which works like this:

1. `sea-orm` is being added as dev-dependency on the `codegen` crate. All entity files under `codegen/tests_cfg` module are being compiled and checked.
2. The tests were performed at the end of `transformer.rs`. Using `create_table_from_entity` to generate SeaQuery create table statement from the SeaORM entity.
3. Then, the statement was passed to `EntityTransformer` and followed by `EntityWriter::gen_compact_code_blocks`.
4. Finally, assert the output from the writer againsts the file defined in `codegen/tests_cfg`

tyt2y3:
This is a very substantial change and seems to only fix a special case. Can we postpone this?

billy1624:
Okay, we can ship this with 0.11.1 and backport this to 0.10.x

Shall we?

tyt2y3:
I mean, to postpone it after 0.11

kev0960:
Just curious when this can be merged? It's the only blocker that we cannot upgrade to a newer version

tyt2y3:
I think the problem will exsit no matter you upgrade or not. I can rebase the commit on master, so you can install it and try it out locally.

tyt2y3 (approved):

PR #1442, opened 2023-02-02
Refactor `select_as` & `save_as` implementation generated by derive macors

billy1624:
## Changes

- [x] More readable implementation for `select_as` & `save_as` methods

### DIFF
--- a/src/entity/column.rs
+++ b/src/entity/column.rs
@@ -931,10 +931,7 @@ mod tests {
 
                 fn select_as(&self, expr: Expr) -> SimpleExpr {
                     match self {
-                        Self::Two => SimpleExpr::cast_as(
-                            Into::<SimpleExpr>::into(expr),
-                            Alias::new("integer"),
-                        ),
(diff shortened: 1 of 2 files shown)


src/entity/column.rs
```diff
@@ -1198,20 +1193,15 @@ mod tests {
...
-                            Into::<SimpleExpr>::into(expr),
-                            Alias::new("integer"),
-                        ),
+                        Self::Two => expr.cast_as(Alias::new("integer")),
                         _ => self.select_enum_as(expr),
                     }
                 }
 
-                fn save_as(&self, expr: Expr) -> SimpleExpr {
+                fn save_as(&self, val: Expr) -> SimpleExpr {
```
tyt2y3, line 1201: Why change it to val?

billy1624: The rename won't affect the implementation on downstream

tyt2y3:
Please also update the documentation

src/entity/column.rs
tyt2y3: 🫡

PR #1454, opened 2023-02-07
Adding community link for tutorial

BrooksPatton:
## PR Info

- Related to #403 

## Changes

- [x] Added reference to online tutorial series


tyt2y3 (commented):
I like this! Not sure why I missed this PR

tyt2y3:
I have been watching the videos and found them interesting. Just one note when going through the Updates chapters:

You don't actually have to apply the `.filter(Id.eq(_))` in `Tasks::update()`. If you look at the docs, https://docs.rs/sea-orm/latest/sea_orm/entity/trait.EntityTrait.html#method.update , the `where id = ?` clause is automatically added.

On the other hand, `Entity::update_many` does not add a `where` clause.

BrooksPatton:
Oh nice! Thanks for the info. When I do the second version I'll be sure to include that info.

Issue #826, opened 2022-06-27
default_expr method not found in `sea_orm::ColumnDef`


TKFRvisionOfficial:
## Description

default_expr does not seem to be implemented

## Steps to Reproduce

```rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "stuff")]
pub struct Model {
	  #[sea_orm(default_expr = "gen_random_uuid()", unique)]
	  pub cool_uuid: Uuid,
}
```

### Expected Behavior

A uuid should be generated if none specified.

### Actual Behavior

```
3 | #[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
  |                                   ^^^^^^^^^^^^^^^^^ method not found in `sea_orm::ColumnDef`
```

### Reproduces How Often

Always

## Versions

```
????????? sea-orm v0.7.1
???   ????????? sea-orm-macros v0.7.0 (proc-macro)
???   ????????? sea-query v0.23.0
???   ???   ????????? sea-query-derive v0.2.0 (proc-macro)
???   ????????? sea-strum v0.23.0
???   ???   ????????? sea-strum_macros v0.23.0 (proc-macro)

Windows 10
```
## Additional Information
(Cargo Expand)
```rust
            Self::CoolUuid => sea_orm::prelude::ColumnType::Uuid
                .def()
                .unique()
                .default_expr("gen_random_uuid()")
```

ikrivosheev:
@TKFRvisionOfficial hello! Thank you, this is strange, because sea-query has an issue: https://github.com/SeaQL/sea-query/issues/347 I am working on it, but I need more time.

billy1624:
Hey @TKFRvisionOfficial, sorry for the delay. We will support `default_expr` in `sea-query` first. Then, we will make the API available on `sea-orm` as well.

tyt2y3:
I think sea-query now supports it, so it should be pretty easy to add macro support in sea-orm

pratushrai0309:
@billy1624 What I exactly need to do in this?

billy1624:
Update [`ColumnDef`](https://docs.rs/sea-orm/latest/sea_orm/entity/struct.ColumnDef.html) to hold `Option<sea_query::SimpleExpr>` instead of `Option<sea_query::Value>`.

Then, update the macros:

[...]

PR #1474, opened 2023-02-12
Add default_expr support for ColumnDef

### LINKED ISSUE #1398: Be able to set the default value on a timestamp column to the CURRENT_TIMESTAMP
## Motivation

When creating a new table with timestamp columns - a created_at, updated_at combo, for instance - it would be very handy to be able to set the default value of the column to be the `CURRENT_TIMESTAMP` function as such:

```sql
CREATE TABLE IF NOT EXISTS sample_table (
  created_at timestamp without time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
  updated_at timestamp without time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
);
```

## Proposed Solutions

Potentially make this a supported default impl fn on the `ColumnDef` struct. Maybe like:

```rust
ColumnDef::new(ApiKey::CreatedAtTs).date_time().not_null().default_current_timestamp()
```

Diwakar-Gupta:
## PR Info

Fixes 
#1398  and #826

## File Changed
1. `sea-orm-macros/src/derives/entity_model.rs` parse attribute and generate function call.
2. `src/entity/column.rs` updated ColumnDef to store `SimpleExpr` instead of `Value` and updated  `default_value` function.

## Tests
Have removed `PartialEq` from `ColumnDef` because `SimpleExpr` does not implement  `PartialEq`, thats why test using `assert_eq` on `ColumnDef` are failing.

To make a temporarily fix we can implementing `eq` method for `ColumnDef` for test only and for supported features only.


Diwakar-Gupta:
I had certain questions:
1. Should I add a test for `default_expr` and may be `default_value`?
2. Should I comment test for now.

billy1624 (commented):
Hey @Diwakar-Gupta, thanks for contributing!!

> 1. Should I add a test for `default_expr` and may be `default_value`?

Yes, please!

> 2. Should I comment test for now.

I think it's possible to implement `PartialEq` for `sea_query::SimpleExpr` but it's not trivial. I can try though.

Diwakar-Gupta:
Thanks I will do that, additionally if you are going to impl `PartialEq` I had done some research for that it will be helpfull.

**
I went through all the fields in `SimpleExpr` and the problem is caused by `pub type DynIden = SeaRc<dyn Iden>;`   in `sea-query/src/types.rs`. This `Iden` is created by `macro_rules! iden_trait` https://github.com/SeaQL/sea-query/blob/94a1ab7290075e8a8bbb7a10e5207350def3455c/src/types.rs#L18.
If we can make it implement `PartialEq` it will solve the problem. 
But `SeaRc<dyn Iden>` is actually `Rc` from standard library which does not implement `PartialEq` and we cannot implement that in other crate than `rust` itself. I am stuck here.
**

tyt2y3:
> I think it's possible to implement `PartialEq` for `sea_query::SimpleExpr` but it's not trivial. I can try though.

We discussed a while before. The gist is `DynIden` is a foreign type we can't simply impl PartialEq for.

But I imagine it could be implemented as follows:

```rust
(*self).get_type_id() == (*other).get_type_id() && self.to_string() == other.to_string()
```

Checking type id is necessary, because two types with the same name is not deemed to be equal.

billy1624:
Hey @Diwakar-Gupta @tyt2y3, I just did a POC, https://github.com/SeaQL/sea-query/pull/620.

ikrivosheev:
Hello! Why do we need `PartialEq` for `ColumnDef`? Does something using this?

billy1624:
Hey @Diwakar-Gupta, I just upgraded SeaQuery dependency to https://github.com/SeaQL/sea-query/pull/620

lazulit3:
> Hey @Diwakar-Gupta, don't worry about the error at https://github.com/SeaQL/sea-orm/actions/runs/4500149841/jobs/7918900727?pr=1474. The root cause is that SeaORM and SeaSchema depend on two different version of SeaQuery. We can fix it after [SeaQL/sea-query#620](https://github.com/SeaQL/sea-query/pull/620) has been merged.

I see that this PR was merged, is there still integration / release / deps work remaining? Am just curious what to keep an eye on regarding blockers / next-steps. :D

billy1624 (approved):
Thanks again for the PR!! @Diwakar-Gupta

tyt2y3 (approved):
Thank you all

Issue #1490, opened 2023-02-18
Script to manage `COMMUNITY.md`


TheAwiteb:
This script should parse the file and sort the sections with its projects

TheAwiteb:
@billy1624 Can I work on this one?

PR #1492, opened 2023-02-18
Community projects sorter/checker

TheAwiteb:
This will helps us sort community projects and check sorting when a new project is added. Closes https://github.com/SeaQL/sea-orm/issues/1490

## Changes
- Add sup-section for Startups section
- Fix `Programming Libraries`, `Frameworks` and `Scaffolding` sub-section from `Open Source Projects` section and make it 4 hashes
- Fix `RSS aggregator` and `Dev Board` format
- Remove github stars badge [1]
- Sort the projects 

[1]: Do we really need it for each community project? I think it will be a little messy

## Todo
- [x] Create a CI that verifies the community file when it is modified

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,7 +46,10 @@
 #   - Jobs that will always run
 #   - `compile-postgres`
 #   - `postgres`
-
+#
+# Following jobs will be run if the `COMMUNITY.md` file was changed
+#
+#   - `community-file-sort-checker`
 
 name: tests
 
@@ -54,10 +57,12 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'
+      - '!COMMUNITY.md'
       - '.github/ISSUE_TEMPLATE/**'
   push:
     paths-ignore:
       - '**.md'
+      - '!COMMUNITY.md'
       - '.github/ISSUE_TEMPLATE/**'
     branches:
       - master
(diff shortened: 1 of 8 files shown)


TheAwiteb:
I have tested it.
When adding a project, it sorts incorrectly
Screenshot from 2023-02-19 04-35-22

After correcting the location of the project
Screenshot from 2023-02-19 04-41-02

TheAwiteb:
@billy1624 I made this https://github.com/SeaQL/sea-orm/pull/1492/commits/c45eec77bf4680809b4d6a76f1b5cb2143f0fd9f because of the error returned by CI [here](https://github.com/SeaQL/sea-orm/actions/runs/4223569626/jobs/7370267471)

billy1624 (commented):
Hey @TheAwiteb, sorry for coming in late. What a sophisticated parser!!

I think the whole idea of this PR is to:
1. Check all the markdown lists in the COMMUNITY.md are sorted
2. Warn if the list items aren't sorted during CI check

From my observation, correct me if I'm wrong, the parser is tailored to the existing structure of COMMUNITY.md. For example, each section titles is defined in h3 tag and subsections are in h4 tag. Also, each list item follows specific specification: name, description, repository link...etc. I afraid all these are subject to change in the future.

So, I suggest another way of doing it. Is it possible to use https://github.com/lqez/slmd to check the sorting? After running the command `slmd COMMUNITY.md -oi`, its exit code tells us if it's sorted or not.

> We can install it with the command `pip install slmd` on CI. Given that the runner already has `pip` with it, https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md.

One step further, we can `diff` the changes. `slmd COMMUNITY.md COMMUNITY_SORTED.md -i` will output the file to `COMMUNITY_SORTED.md`, then we can somehow `diff` them.

Sorry for proposing a new idea when everything is ready :(

TheAwiteb:
Hi @billy1624. I looked at https://github.com/lqez/slmd and it turns out that it is very inflexible, as it does not sort or check the order of section and sub-section. Also, in the future, if we want, for example, to verify that the links are actually links, and they are https, this is not possible with this tool. And if we want to update the structure of the file in the future, we will only urge the parser, and the rest will remain as it is.

billy1624:
Hey @TheAwiteb, keeping the relative order of each section and sub-section is important to us and it's subject to change in the future. For me, sorting the items within each list would be enough.

PR #1509, opened 2023-02-25
Delete with returning

### LINKED ISSUE #615: [ADD]: Addition of delete many
## Summary

Delete the row(s) on db at the same time returning the data of deleted row(s). 

## Motivation

Various other delete methods, this method is a good method to implement as it makes the ORM more complete.

## Additional Information

`Rough idea is that we need two new methods in DeleteMany, https://github.com/SeaQL/sea-orm/blob/af235168db3a675e889974a34799890350be00e4/src/executor/delete.rs#L34-L46, namely one & all which return single model & array of models respectively. For database support RETURNING, https://docs.rs/sea-orm/0.6.0/sea_orm/query/trait.ConnectionTrait.html#method.support_returning, it will use RETURNING syntax on delete statement to get the data of deleted models.`

AmineZouitine:
## PR Info

- Closes #615

- Dependencies:
  - N/A

- Dependents:
  - N/A

## New Features

- [x] Addition of method to `DeleteOne` structure that returns the deleted element. I based this implementation on the update method.

## Bug Fixes

- N/A

## Breaking Changes

- N/A

## Additional information
The Pull Request doesn't fully match the issue. After reviewing the entire code, I've taken the initiative to stay within the scope of the existing code, particularly the executor/update, which also defines a way to retrieve an element with the "RETURNING" feature. 

## Tests
[Here](https://github.com/AmineZouitine/sea-orm/blob/delete_with_returning/tests/delete_with_returning_tests.rs)

I have written two tests to verify the proper functioning of the new feature:

- `delete_one_with_returning`: This test verifies that the model is correctly retrieved after a delete operation.
- `delete_one_with_returning_not_exist`: This test ensures that the program triggers the correct error when the entry does not exist.

﻿

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -47,6 +47,10 @@ pub enum DbErr {
     /// if the record has been correctly updated, otherwise this error will occur
     #[error("Failed to get primary key from model")]
     UpdateGetPrimaryKey,
+    /// When deleting, a model should know its primary key to check
+    /// if the record has been correctly deleted, otherwise this error will occur
+    #[error("Failed to get primary key from model")]
+    DeleteGetPrimaryKey,
     /// The record was not found in the database
     #[error("RecordNotFound Error: {0}")]
     RecordNotFound(String),
--- a/src/error.rs
+++ b/src/error.rs
@@ -77,6 +77,10 @@ pub enum DbErr {
     /// May be the table is empty or the record does not exist
     #[error("None of the records are updated")]
     RecordNotUpdated,
+    /// None of the records are Deleted, that means a WHERE condition has no matches.
+    /// May be the table is empty or the record does not exist
+    #[error("None of the records are updated")]
+    RecordNotDeleted,
 }
 
 /// Runtime error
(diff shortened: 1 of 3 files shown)


billy1624 (commented):
Hey @AmineZouitine, thanks for contributing!!

I think it's better to enable `Deleter` to execute a delete statement while returning a single model or vector of model.

I imagine there could be two new methods in the `Deleter`:
1. `async fn one(mut self, db: &C) -> Result<Model, DbErr>`
2. `async fn all(mut self, db: &C) -> Result<Vec<Model>, DbErr>`

Both methods did the same thing except that it yields single / multiple Model. For DB that support returning it should add returning expression to the `DeleteStatement`.

Otherwise, if the DB does not support returning. We need to extract the `WHERE` condition from the delete statement, construct a select statement based on that and then query the rows. Finally, return the selected rows after the delete statement has been executed.

tyt2y3:
closed in favor of #2432

PR #1513, opened 2023-02-28
set unselected `Column` of type `Option<T>` to None

Reverier-Xu:
## PR Info

- Closes #1507 

- Dependencies:
  - none 

- Dependents:
  - none 

## New Features

- Support for filling None on `Column` of type `Option<T>` which is not selected instead of throwing an error.  

## Changes

- Added a branch judgment in `impl TryGetable for Option<T>`

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -140,7 +140,10 @@ impl<T: TryGetable> TryGetable for Option<T> {
         match T::try_get_by(res, index) {
             Ok(v) => Ok(Some(v)),
             Err(TryGetError::Null(_)) => Ok(None),
-            Err(TryGetError::DbErr(DbErr::Query(RuntimeErr::SqlxError(sqlx::Error::ColumnNotFound(_))))) => Ok(None),
+            #[cfg(feature = "sqlx-dep")]
+            Err(TryGetError::DbErr(DbErr::Query(RuntimeErr::SqlxError(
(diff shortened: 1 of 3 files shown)


src/executor/query.rs
```diff
@@ -140,6 +140,7 @@ impl<T: TryGetable> TryGetable for Option<T> {
         match T::try_get_by(res, index) {
             Ok(v) => Ok(Some(v)),
             Err(TryGetError::Null(_)) => Ok(None),
+            Err(TryGetError::DbErr(DbErr::Query(RuntimeErr::SqlxError(sqlx::Error::ColumnNotFound(_))))) => Ok(None),
```
billy1624, line 143: Hey @Reverier-Xu, thanks for contributing!! This is interesting!!

I think this will work if it's feature guarded as below.

```suggestion
#[cfg(feature = "sqlx-dep")]
Err(TryGetError::DbErr(DbErr::Query(RuntimeErr::SqlxError(sqlx::Error::ColumnNotFound(_))))) => Ok(None),
```

Also, can we add a few test cases to proof that it's actually working as expected? We might need to make some similar changes to the mock interface as well.

tyt2y3 (approved):
Thanks for the contribution.
This is a small change, but with a bigger implication.
So we would need to document this as a breaking behavioural change.

PR #1520, opened 2023-03-02
Update heck dependency

sergeiivankov:
## New Features

- [ ] No

## Bug Fixes

- [ ] No

## Breaking Changes

- [ ] No

## Changes

- [ ] Update heck dependency to latest minor version to prevent including multiple versions in used `sea-orm` other crate dependent on `heck` latest version projects

Related with https://github.com/davidpdrsn/bae/pull/12, https://github.com/SeaQL/strum/pull/1 and https://github.com/SeaQL/sea-schema/pull/103

PR #1535, opened 2023-03-11
Drop the use of `sea-strum` and depends on the original `strum` with a tailored `EnumIter` provided

billy1624:
## PR Info

- Dependencies:
  - https://github.com/SeaQL/sea-orm/pull/1520

- [ ] This PR should be merged into master. But I want to substitute the changes of https://github.com/SeaQL/sea-orm/pull/1520 for readability.

## New Features

- [x] Added `sea_orm_macros::EnumIter` to implement `strum::IntoEnumIterator` trait for the derived enum (source code adapted from https://github.com/Peternator7/strum)

## Breaking Changes

- [x] Added `derive` and `strum` features to `sea-orm-macros`
- [x] Replaced `sea-strum` dependency with `strum` in `sea-orm`
- [x] Re-exported `sea_orm_macros::EnumIter` instead of `strum::EnumIter` on the root of `sea-orm`


sea-orm-macros/Cargo.toml
```diff
@@ -29,4 +29,7 @@ sea-orm = { path = "../", features = ["macros"] }
 serde = { version = "1.0", features = ["derive"] }
 
 [features]
+default = ["derive"]
```
tyt2y3, line 32: Why do we need a feature here? I think we always want to enable this right?

sea-orm-macros/Cargo.toml
```diff
@@ -29,4 +29,7 @@ sea-orm = { path = "../", features = ["macros"] }
...
 postgres-array = []
+derive = ["bae"]
+strum = []
```
tyt2y3, line 35: Same, I think we just ship it by default

sea-orm-macros/src/strum/mod.rs
tyt2y3, line 1: Did we actually copied / adapted code from `strum`? If so, may be we should include their license here.

sea-orm-macros/Cargo.toml
billy1624: Backward compatibility. Not a must, I just think it's better to ship with `derive` by default like it used to be.

sea-orm-macros/src/strum/mod.rs
billy1624: We did. Should I put its licence to `sea-orm-macros/src/strum` directory?

sea-orm-macros/Cargo.toml
billy1624: We have two features in `sea-orm-macros`
1. `derive` - provide all derive macros of SeaORM, e.g. `DeriveEntityModel`
2. `strum` - provide the `EnumIter` derive macros

I split the `sea-orm-macros` in half because one can opt-out the `derive` feature in SeaORM by specifying `sea-orm = { version = "*", default-features = false }`. Where the `sea-orm-macros` is a optional dependency. But now, the `strum` reside in it, making it a "compulsory" dependency. Instead, we turn on / off the `derive` feature as needed.

In contrast, we ALWAYS want the `strum` feature to be enabled.

Hence,

```diff
- sea-orm-macros = { version = "0.11.0", path = "sea-orm-macros", default-features = false, optional = true }
+ sea-orm-macros = { version = "0.11.0", path = "sea-orm-macros", default-features = false, features = ["strum"] }

[features]
- macros = ["sea-orm-macros", "sea-query/derive"]
+ macros = ["sea-orm-macros/derive", "sea-query/derive"]
```

sea-orm-macros/Cargo.toml
billy1624: For this, dependency are conditionally enabled when a feature is being enabled.

This should be fine. If we decided to keep `derive` and `strum` as it's now.

sea-orm-macros/src/strum/mod.rs
tyt2y3: That would do

sea-orm-macros/Cargo.toml
tyt2y3: Umm. This sounds complicated. Well but let it be for now.

tyt2y3 (approved):
Awesome

PR #1577, opened 2023-04-04
feat: export `ConnectionTrait` in `sea_orm_migration::prelude`

billy1624:
## PR Info

- Closes https://github.com/SeaQL/seaql.github.io/pull/97

## New Features

- [x] Re-export `ConnectionTrait` in `sea_orm_migration::prelude`

### DIFF
--- a/sea-orm-migration/src/prelude.rs
+++ b/sea-orm-migration/src/prelude.rs
@@ -9,5 +9,5 @@ pub use async_trait;
 pub use sea_orm::{
     self,
     sea_query::{self, *},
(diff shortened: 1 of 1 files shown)


julien-leclercq:
Hello! As nice as an addition that is, I wonder if it's totally discarding the documentation update. 
I am not familiar with your release model but, would not this change be private until 0.12.0 if it stays on master like that?

tyt2y3:
For changes like this we are going to stuff it into the next patch release

PR #1614, opened 2023-04-25
partialEq for value inside activevalue

### LINKED ISSUE #1533: Add contains method for ActiveValue
## Motivation

Sometimes it is necessary to compare a value in a model that is already the active model. Now this is done somewhat inconveniently, you need to use `!is_not_set()` and `as_ref()`. Which makes the code a bit more complicated.

## Proposed Solutions

```rs
let id_set = ActiveValue::Set(1);
let id_unchanged = ActiveValue::Unchanged(1);
let id_not_set = ActiveValue::<i32>::Unchanged(1);

id_set.contains(&1); // true
id_unchanged.contains(&1); // true
id_not_set.contains(&1); // false
```

## Current Workarounds

```rs
let id_set = ActiveValue::Set(1);

!id_set.is_not_set() && id_set.as_ref() == &1
```

Diwakar-Gupta:
## PR Info

- Closes #1533 

- Dependencies:
  No Dependencies

- Dependents:
   No Dependents

## New Features

- [ ] this pr enables to compare ActiveValue dosent matter set or unset 
```rust
assert!(ActiveValue::Set(2) == 2);
assert!(ActiveValue::UnSet(2) == 2);
assert!(ActiveValue::<u8>::NotSet != 2);
```

## Breaking Changes

- [x] This is backward compatable

## Changes

- [x] added `impl<U, V> PartialEq<U> for ActiveValue<V>`


billy1624:
I think this is the best we can get:

```rs
assert!(ActiveValue::Set(2) == 2);
```

We wouldn't be able to make this works:

```rs
assert!(2 == ActiveValue::Set(2));
```

Because we simply can't implement `PartialEq<ActiveValue<V>> for U`

Diwakar-Gupta:
`PartialEq<sea_query::Value>`

This will make sure that we are only comparing with `sea_query::Value` rather than something that can be converted to `sea_query::Value`, am i right?

billy1624:
I mean comparing `sea_orm::ActiveValue<?>` with `sea_query::Value`. e.g.
```rs
assert!(ActiveValue::Set(2) == Value::from(2));
assert!(ActiveValue::Unchanged(String::from("hi")) == Value::from("hi"));
```
However, after second thought, it'd be more intuitive to do... instead
```rs
assert!(ActiveValue::Set(2) == 2);
assert!(ActiveValue::Unchanged(String::from("hi")) == "hi");
```

billy1624 (approved):
Thanks!! @Diwakar-Gupta

src/entity/active_model.rs
```diff
@@ -880,6 +880,19 @@ where
...
+    U: Into<Value>,
+    V: Into<Value> + std::cmp::PartialEq<U>,
+{
+    fn eq(&self, other: &U) -> bool {
+        match self {
+            Self::Set(value) | Self::Unchanged(value) => value.eq(other),
+            Self::NotSet => false,
+        }
+    }
+}
```
billy1624, line 894:
```rs
impl<U, V> PartialEq<Option<U>> for ActiveValue<V>
where
    U: Into<Value>,
    V: Into<Value> + std::cmp::PartialEq<U>,
{
    fn eq(&self, other: &Option<U>) -> bool {
        match (self, other) {
            (Self::Set(value) | Self::Unchanged(value), Some(other)) => value.eq(other),
            (Self::NotSet, Some(_)) | (_, None) => false,
            (Self::NotSet, None) => true,
        }
    }
}
```

src/entity/active_model.rs
```diff
@@ -1306,6 +1319,16 @@ mod tests {
...
 
+    #[test]
+    fn test_active_value_partialeq() {
+        assert!(ActiveValue::Set(2) == 2);
+        assert!(ActiveValue::Set(2) != 3);
+        assert!(ActiveValue::<u8>::NotSet != 3);
+
+        assert!(ActiveValue::Unchanged(String::from("hi")) == "hi");
+        assert!(ActiveValue::Unchanged(String::from("hi")) != "bye");
+    }
```
billy1624, line 1330:
```rs
#[test]
fn test_active_value_partialeq() {
    assert!(ActiveValue::Set(2) == Some(2));
    assert!(ActiveValue::Set(2) != Some(3));
    assert!(ActiveValue::<u8>::NotSet != Some(3));
    assert!(ActiveValue::<u8>::NotSet == None::<u8>);

    assert!(ActiveValue::Unchanged(String::from("hi")) == Some("hi"));
    assert!(ActiveValue::Unchanged(String::from("hi")) != Some("bye"));
}
```

billy1624 (changes requested):
Hey @Diwakar-Gupta, after second thought... I think comparing `ActiveValue<V>` with `Option<V>` make more sense. This way we don't leave out `ActiveValue::NotSet`.

tyt2y3:
> Hey @Diwakar-Gupta, after second thought... I think comparing `ActiveValue<V>` with `Option<V>` make more sense. This way we don't leave out `ActiveValue::NotSet`.

Actually, the solution can be much simpler:

we already have
```rust
impl<V> std::convert::AsRef<V> for ActiveValue<V>
```

we only have to add
```rust
impl<V> std::convert::AsRef<Option<V>> for ActiveValue<V>
```

?

Diwakar-Gupta:
based on the requirement from issue this will work.

Never thought it can be solved with this simplicity.

Diwakar-Gupta:
Can we give thought to possible solution once again

below only utilizes Set and Unchanged only but client api is very simple
```rust
assert!(ActiveValue::Set(2) == 2);
```

this utilizes Set and Unchanged and NotSet 
```rust
assert!(ActiveValue::Set(2) == Some(2));
```

and the third one which get's converetd to Option<>
```rust
assert!(ActiveValue::Set(2).as_ref() == Some(2).as_ref());
```

Expurple:
Note that nowadays we have non-panicking [`ActiveValue::try_as_ref`](https://docs.rs/sea-orm/1.1.14/sea_orm/entity/enum.ActiveValue.html#method.try_as_ref). That may be good enough:

```rust
let active_value = ActiveValue::Set(1);

active_value.try_as_ref() == Some(&1)
```

PR #1622, opened 2023-04-28
Showcase how to define entity with dynamic table name

billy1624:
## PR Info

- Discussed https://github.com/SeaQL/sea-orm/discussions/1616


tests/dyn_table_name_tests.rs
```diff
@@ -0,0 +1,79 @@
...
+        };
+        // Prepare insert statement
+        let mut insert = Entity::insert(model.clone().into_active_model());
+        // Reset the table name of insert statement
+        insert.query().into_table(entity.table_ref());
+        // Execute the insert statement
+        assert_eq!(insert.exec(db).await?.last_insert_id, 1);
+
+        // Prepare select statement
+        let mut select = Entity::find();
```
brendano257, line 46: Is there any way to use the Entity directly in full ORM style, like `Entity::find_by_id(0).one(&db).await?;` here? Or is that lost when we have to initialize the Entity with data?

billy1624: > Or is that lost when we have to initialize the Entity with data?

This is correct. Unfortunately the table name data is lost and we need to set it manually.

PR #1631, opened 2023-05-06
Add Serialize/Deserialize for ActiveValue

videobitva:
Add Serialize/Deserialize for ActiveValue

## PR Info

As of right now there exists "with-json" feature flag, which allows applying changes to an ActiveModel through the JSON Value type. This could be useful when updating some data in the database by sending a JSON to the server through an HTTP API. An alternative way of doing the same thing would be to send the ActiveModel, with all its changes relative tp the Model, directly to the server. This requires the serialization of the ActiveModel. At the moment (sea-orm version 0.11) it is not possible due to Serialize/Deserialize traits not being implemented for ActiveValue. This PR adds the derive for these traits conditionally when the "with-json" feature is enabled. This way the Serialize/Deserialize traits implementation for the exact ActiveModel could be done by the library user, while the implementation for the ActiveValue is behind the feature flag.

## New Features

- [x] Serialize/Deserialize for ActiveValue

## Breaking Changes

- [x] No breaking changes

## Changes

- [x] Add "serde/derive" dependency when building with "with-json" feature flag
- [x] Add Serialize/Deserialize for ActiveValue behind the "with-json" feature flag

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -2,8 +2,8 @@ use crate::{
     error::*, ConnectionTrait, DeleteResult, EntityTrait, Iterable, PrimaryKeyToColumn, Value,
 };
 use async_trait::async_trait;
-use serde::{Serialize, Deserialize};
 use sea_query::{Nullable, ValueTuple};
+use serde::{Deserialize, Serialize};
 
 use std::fmt::Debug;
 
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -2,7 +2,9 @@ use crate::{
     error::*, ConnectionTrait, DeleteResult, EntityTrait, Iterable, PrimaryKeyToColumn, Value,
 };
 use async_trait::async_trait;
+use serde::{Serialize, Deserialize};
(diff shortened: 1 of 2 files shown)


sphw:
What's the current status of this PR? It looks like exactly what I need. I'm happy to push it over the finish line as well

src/entity/active_model.rs
```diff
@@ -36,6 +38,7 @@ pub use ActiveValue::NotSet;
 /// );
 /// ```
 #[derive(Clone, Debug)]
+#[cfg_attr(feature = "with-json", derive(Serialize, Deserialize))]
```
tyt2y3, line 41: We can use sea_query's `value_to_json` to handle the `Serialize` part, but in general ActiveValue does not seem to lend itself for `Deserialize`

tyt2y3:
Thank you for the contribution!

Do you really needed `Deserialize`? I think it's better to leave it out, because it is a complex problem (safety, customization, schema validation etc).

May be also add some test cases under `/tests` to demonstrate the end to end behaviour?

Issue #1255, opened 2022-11-25
DeriveActiveEnum's implicit implementation of Display trait needs to be overridable


nickb937:
In reference to the implementation of the Display trait for https://github.com/SeaQL/sea-orm/issues/955, the solution provided does not always suit all applications because an application may choose to present the enum in a different way, depending on use case.

For example, for clarity in code, it makes sense to write out the genders in their full form, but presentation elsewhere can often be suitably understood by their single letter abbreviations:

```
#[derive(Debug, Clone, Hash, Eq, PartialEq, EnumIter, DeriveActiveEnum, Serialize, Deserialize_repr)]
#[sea_orm(rs_type = "i16", db_type = "SmallInteger")]
#[repr(u16)]
pub enum Gender {
    #[sea_orm(num_value = 1)]
    Male = 1,
    #[sea_orm(num_value = 2)]
    Female = 2,
    #[sea_orm(num_value = 3)]
    Unspecified = 3,
    #[sea_orm(num_value = 4)]
    Other = 4,
}

impl Display for Gender {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}",
               match self {
                   Gender::Male => "M",
                   Gender::Female => "F",
                   Gender::Unspecified => "U",
                   Gender::Other => "O",
               }
        )
    }
}
```

The suggestion from Discord is to implement a `display_value` attribute to overside the default display value of, say, "Male" with "M":

```
    #[sea_orm(num_value = 1, display_value="M")]
    Male = 1,
    #[sea_orm(num_value = 1, display_value="F")]
    Female = 2,
```

simonask:
Ran into this as well.

My problem is that I have a string enum, and the `Display` implementation of `sea_orm::Value` outputs SQL, so the string becomes quoted.

I guess I don't understand why `Display` is the correct trait to implement here. It's meant to produce a user-facing string.

I worked around it by implementing using `#[strum(serialize = "...")]` with the same string value, and then using `maud::Render` to override the `Display` implementation. But this is problematic and error-prone for other reasons (HTML isn't the only user-facing output in my situation).

Please advise. :-)

tyt2y3:
I do think the current implementation of `Display` is weird. To be precise, we should print `Male` instead of `1`.
You can write a custom `Serialize` impl, or you can use a wrapper type and override `Display`.

tyt2y3:
It might be better we rework that implementation

[...]

PR #1636, opened 2023-05-13
Added support for customizing the way `Display` works on `DeriveActiveEnum`

asdfcube:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1255

## New Features

- [x] added `display`, `display_value`, `display_label` attributes to the `#[sea_orm()]` container and variant attribute.
- [x] `DeriveActiveEnum` only implements `std::fmt::Display` trait automatically if one of the attributes are present.
- [x] made users able to customise the display name by using the variant attribute `display = "string literal"`.

## BREAKING CHANGES

- [x] `#[derive(DeriveActiveEnum)]` no longer derives `std::fmt::Display` automatically if `display` or `display_value` or `display_label` are not present in the container attributes.
- [x] `display` defaults to displaying label instead of value now


src/entity/active_enum.rs
```diff
@@ -275,6 +275,64 @@ mod tests {
...
+        #[sea_orm(rs_type = "i32", db_type = "Integer")]
+        enum DefaultDisplay {
+            One = 1,
+            Two = 2,
+            Three = 3,
+        }
+
+        assert_eq!(format!("{}", DefaultDisplay::One), "1");
+        assert_eq!(format!("{}", DefaultDisplay::Two), "2");
+        assert_eq!(format!("{}", DefaultDisplay::Three), "3");
```
tyt2y3, line 291: As per our discussion, I think we should remove the default Display impl, resulting in a breaking change.

src/entity/active_enum.rs
```diff
@@ -275,6 +275,64 @@ mod tests {
...
+
+        // All `label` -> All print label
+        #[derive(PartialEq, Eq, EnumIter, DeriveActiveEnum)]
+        #[sea_orm(rs_type = "i32", db_type = "Integer", display = "label")]
```
tyt2y3, line 295: "label" can be the default. May be we can make `display` an alias of `display = "label"`.

src/entity/active_enum.rs
```diff
@@ -275,6 +275,64 @@ mod tests {
...
+
+        assert_eq!(format!("{}", LabelDisplay::One), "'One'");
+        assert_eq!(format!("{}", LabelDisplay::Two), "'Two'");
+        assert_eq!(format!("{}", LabelDisplay::Three), "'Three'");
+
+        // Default with override
+        #[derive(PartialEq, Eq, EnumIter, DeriveActiveEnum)]
+        #[sea_orm(rs_type = "i32", db_type = "Integer")]
+        enum DefaultDisplayOverride {
+            #[sea_orm(display = "label")]
```
tyt2y3, line 310: Actually, I think this override should be the thing the user wants?

So `display = "hello"` should result in 

```rust
assert_eq!(format!("{}", DefaultDisplayOverride::One), "'hello'");
```

src/entity/active_enum.rs
```diff
@@ -275,6 +275,64 @@ mod tests {
...
+        // All `label` -> All print label
+        #[derive(PartialEq, Eq, EnumIter, DeriveActiveEnum)]
+        #[sea_orm(rs_type = "i32", db_type = "Integer", display = "label")]
+        enum LabelDisplay {
+            One = 1,
+            Two = 2,
+            Three = 3,
+        }
+
+        assert_eq!(format!("{}", LabelDisplay::One), "'One'");
```
tyt2y3, line 302: Umm... I don't have a clear sense, but seems like the `'` is unnecessary in this context? Like the expected should be:

```rust
assert_eq!(format!("{}", LabelDisplay::One), "One");
```

src/entity/active_enum.rs
asdfcube: done in 656c777e8c7d33d2eb4f316be3b7627f07d5c72f

src/entity/active_enum.rs
asdfcube: as discussed, we want `display = label | value | "string literal"` but i cannot get it working, so instead went to `display = "string literal" | display_value | display_label`.

src/entity/active_enum.rs
asdfcube: that relied on the implementation of `Value::to_string`, so...

src/entity/active_enum.rs
asdfcube: that relied on the implementation of `Value::to_string`, so...

src/entity/active_enum.rs
asdfcube: done in 656c777e8c7d33d2eb4f316be3b7627f07d5c72f

src/entity/active_enum.rs
asdfcube: done in 93cdf20277d1cc4294bf5a8301b5daff6a2639f4

src/entity/active_enum.rs
asdfcube: as discussed, we will keep the behaviour

tyt2y3:
@billy1624 @darkmmon please put this on the task list

billy1624:
Sure. I want him to touch on tangible API before entering the magical proc_macro world

Issue #873, opened 2022-07-14
insert_many failing if the models iterator is empty


horacimacias:
## Description

I'm executing an `insert_many` operations where the iterator is the result of mapping/filtering.
Sometimes, as a result of the mapping/filtering, the resulting iterator may be empty.
In this case, the insert_many operation fails; I'd expect this to succeed and the database not being hit at all as there is nothing to do.
I increased the logs and I saw the following:

```
INSERT INTO
  "rules"
VALUES
  (DEFAULT) RETURNING "id"
..... Db { source: Query("error returned from database: null value in column \"id\" violates not-null constraint") } 
```

the entity/struct is the following:
```rust

#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "rules")]
pub struct Model {
    #[sea_orm(primary_key, auto_increment = false)]
    pub id: Uuid,
    pub rule_set: String,
    pub created: NaiveDateTime,
    pub version: i16,
    #[sea_orm(column_type = "JsonBinary")]
    pub rule: Value,
}
```

I'm using yugabyte for this which should be using postgres.

## Steps to Reproduce

1. Execute an insert_many operations where the iterator is empty
2. Confirm success/failure

### Expected Behavior

I'd expect this to succeed and do nothing on the database.

### Actual Behavior

Operation fails and complains about null  ìd` which is the primary key in my model.

### Reproduces How Often

If the iterator is empty, always.

## Versions

```bash
│   │   ├── sea-orm v0.9.0
│   │   │   ├── sea-orm-macros v0.9.0 (proc-macro)
│   │   │   ├── sea-query v0.26.0
│   │   │   │   ├── sea-query-derive v0.2.0 (proc-macro)
│   │   │   │   ├── sea-query-driver v0.2.0 (proc-macro)
│   │   │   ├── sea-strum v0.23.0
│   │   │   │   └── sea-strum_macros v0.23.0 (proc-macro)
│   │   ├── sea-orm-migration v0.9.0
│   │   │   ├── sea-orm v0.9.0 (*)
│   │   │   ├── sea-orm-cli v0.9.1
│   │   │   │   ├── sea-schema v0.9.2
│   │   │   │   │   ├── sea-query v0.26.0 (*)
│   │   │   │   │   └── sea-schema-derive v0.1.0 (proc-macro)
│   │   │   ├── sea-schema v0.9.2 (*)
│   │   ├── sea-schema v0.9.2 (*)
```

[...]

PR #1682, opened 2023-06-01
Added `InsertWithFallback` so that empty iterators passed to `Insert` will be handled with a custom function of the users' choice

asdfcube:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/873

## New Features

- [x] `InsertWithFallback`, a simple wrapper around `Insert` that checks if the iterator passed to the `Insert` is empty and run a function if it is true.
- [x] `FallbackResult`, a simple enum so that the user knows if the fallback was triggered or not.

### DIFF
--- a/src/query/insert.rs
+++ b/src/query/insert.rs
@@ -229,6 +229,132 @@ where
     }
 }
 
+/// The Insert struct but with fallback for empty iterator. Can be constructed through Insert::fallback
+#[derive(Debug)]
+pub struct InsertWithFallback<A, F, R>
+where
+    A: ActiveModelTrait,
+    F: FnOnce() -> R,
+{
+    pub(crate) insert: Insert<A>,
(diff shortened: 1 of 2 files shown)


src/query/insert.rs
```diff
@@ -229,6 +229,132 @@ where
...
+    where
+        M: IntoActiveModel<A>,
+        I: IntoIterator<Item = M>,
+    {
+        self.insert = self.insert.add_many(models);
+        self
+    }
+
+    /// On conflict
+    ///
```
tyt2y3, line 310: These examples seems to have not related to the new feature?

tyt2y3 (changes requested):
Can you add some test cases to demo the API?

tyt2y3:
We'll probably let @darkmmon take over

Issue #1194, opened 2022-11-06
simplify specific db errors matching


Samiisd:
## Motivation

Currently, matching DB specific errors is highly verbose, and I believe some common errors could benefit from being simplified. Current solution is the following: https://github.com/SeaQL/sea-orm/discussions/1152

E.g.
```rust
match user::Entity::insert_many(models).exec(&state.db).await {
    Err(DbErr::Query(RuntimeErr::SqlxError(sqlx::Error::Database(e))))
        if e.code().unwrap().eq("23505") =>
    {
        // 23505 is postgresql error code for violation on column unique constraint
        Ok()
    }
    Err(e) => Err(e),
    Ok(_) => Ok()
}
```

## Proposed Solutions

I don't know every use cases and common errors, but I believe this ORM would benefit from simplifying error handling. 
Another generic enum error could be created + custom DB error mapping logic to this newly created enum.

E.g.
```
enum DbErr {
     ForeignKeyViolation { column: String, key: int },
     NotNullViolation { column: String },
     UniqueViolation { column: String },
     ...
}
```

tyt2y3:
See: https://github.com/SeaQL/sea-orm/pull/1040#issuecomment-1272536790

billy1624:
Closed by https://github.com/SeaQL/sea-orm/pull/1707

PR #1707, opened 2023-06-13
Added enum for SQLx Error and basic functions for it

darkmmon:
## PR Info

- Closes  https://github.com/SeaQL/sea-orm/issues/1194

## New Features

- [ ]  added a simpler SQL error enum in addition to the general DbErr

## Changes

- [ ]

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -146,16 +146,15 @@ where
 pub enum SqlErr {
     /// error for inserting a record with a key that already exists in the table
     #[error("Cannot have record with same key")]
-    UniqueConstraintViolation(),
+    UniqueConstraintViolation,
     /// error for Foreign key is not primary key
     #[error("Cannot add non-primary key from other table")]
-    ForeignKeyConstraintViolation(),
+    ForeignKeyConstraintViolation,
 }
 
 #[allow(dead_code)]
 impl DbErr {
-    /// convert generic DbErr by sqlx to SqlErr
-    /// return none if input is not any of SqlErr
+    /// Convert generic DbErr by sqlx to SqlErr, return none if the error is not any type of SqlErr
     pub fn sql_err(&self) -> Option<SqlErr> {
         #[cfg(any(
             feature = "sqlx-mysql",
@@ -168,15 +167,32 @@ impl DbErr {
             | DbErr::Query(RuntimeErr::SqlxError(sqlx::Error::Database(e))) = self
             {
                 let error_code = e.code().unwrap_or_default();
-                let error_code_expanded = error_code.deref();
+                let _error_code_expanded = error_code.deref();
                 #[cfg(feature = "sqlx-mysql")]
                 if e.try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()
                     .is_some()
                 {
-                    match error_code_expanded {
-                        "23000" => return Some(SqlErr::UniqueConstraintViolation()),
-                        "1586" => return Some(SqlErr::UniqueConstraintViolation()),
-                        "1452" => return Some(SqlErr::ForeignKeyConstraintViolation()),
+                    let error_number = e
+                        .try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()?
+                        .number();
+                    match error_number {
+                        // 1022 Can't write; duplicate key in table '%s'
+                        // 1062 Duplicate entry '%s' for key %d
+                        // 1169 Can't write, because of unique constraint, to table '%s'
+                        // 1586 Duplicate entry '%s' for key '%s'
+                        1022 | 1062 | 1169 | 1586 => {
+                            return Some(SqlErr::UniqueConstraintViolation)
+                        }
+                        // 1216 Cannot add or update a child row: a foreign key constraint fails
+                        // 1217 Cannot delete or update a parent row: a foreign key constraint fails
+                        // 1451 Cannot delete or update a parent row: a foreign key constraint fails (%s)
+                        // 1452 Cannot add or update a child row: a foreign key constraint fails (%s)
+                        // 1557 Upholding foreign key constraints for table '%s', entry '%s', key %d would lead to a duplicate entry
+                        // 1761 Foreign key constraint for table '%s', record '%s' would lead to a duplicate entry in table '%s', key '%s'
+                        // 1762 Foreign key constraint for table '%s', record '%s' would lead to a duplicate entry in a child table
+                        1216 | 1217 | 1451 | 1452 | 1557 | 1761 | 1762 => {
+                            return Some(SqlErr::ForeignKeyConstraintViolation)
+                        }
                         _ => return None,
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -139,3 +139,60 @@ where
...
+    }
+}
+
+impl PartialEq for SqlErr {
+    fn eq(&self, other: &Self) -> bool {
+        self.to_string() == other.to_string()
+    }
+}
+
+impl Eq for SqlErr {}
```
billy1624, line 198: Consider implementing these two trait via derive.
https://doc.rust-lang.org/rust-by-example/trait/derive.html

src/error.rs
```diff
@@ -139,3 +139,65 @@ where
...
+            {
+                if e.try_downcast_ref::<SqlxMySqlError>().is_some() {
+                    if e.code().unwrap().eq("1062") | e.code().unwrap().eq("1586") {
+                        return Some(SqlErr::UniqueConstraintViolation());
+                    };
+                    if e.code().unwrap().eq("1452") {
+                        return Some(SqlErr::ForeignKeyConstraintViolation());
+                    };
+                }
+            }
```
billy1624, line 177: I believe this can be simply as

```suggestion
                #[cfg(feature = "sqlx-mysql")]
                if e.try_downcast_ref::<SqlxMySqlError>().is_some() {
                    if e.code().unwrap().eq("1062") | e.code().unwrap().eq("1586") {
                        return Some(SqlErr::UniqueConstraintViolation());
                    };
                    if e.code().unwrap().eq("1452") {
                        return Some(SqlErr::ForeignKeyConstraintViolation());
                    };
                }
```

src/error.rs
```diff
@@ -139,3 +139,67 @@ where
...
+            {
+                let error_code = e.code().unwrap_or_default();
+                let error_code_expanded = error_code.deref();
+                #[cfg(feature = "sqlx-mysql")]
+                if e.try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()
+                    .is_some()
+                {
+                    match error_code_expanded {
+                        "23000" => return Some(SqlErr::UniqueConstraintViolation()),
+                        "1586" => return Some(SqlErr::UniqueConstraintViolation()),
```
billy1624, line 178: I wonder why we have two MySQL error codes representing the same unique constraint violation error?

Can we have a comment documenting the corresponding web page of each database's error code?

tests/sql_err_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+    // but the whole test is useless in that case anyway.
+    #[allow(unused_variables)]
+    let error: DbErr = cake
+        .into_active_model()
+        .insert(db)
+        .await
+        .expect_err("inserting should fail due to duplicate primary key");
+
+    let sqlerr = error.sql_err();
+    assert_eq!(sqlerr.unwrap(), SqlErr::UniqueConstraintViolation())
```
billy1624, line 56: Maybe?

```suggestion
    assert_eq!(error.sql_err(), Some(SqlErr::UniqueConstraintViolation()))
```

src/error.rs
```diff
@@ -139,3 +139,67 @@ where
...
+/// An error from unsuccessful SQL query
+#[derive(Error, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum SqlErr {
+    /// error for inserting a record with a key that already exists in the table
+    #[error("Cannot have record with same key")]
+    UniqueConstraintViolation(),
+    /// error for Foreign key is not primary key
+    #[error("Cannot add non-primary key from other table")]
+    ForeignKeyConstraintViolation(),
```
billy1624, line 152: Do we have extra data for each error type? If no, then we can simply remove the `()`.

```suggestion
    /// error for inserting a record with a key that already exists in the table
    #[error("Cannot have record with same key")]
    UniqueConstraintViolation,
    /// error for Foreign key is not primary key
    #[error("Cannot add non-primary key from other table")]
    ForeignKeyConstraintViolation,
```

billy1624 (commented):
Hey Ivan, nice! The `sql_err` method looks a lot cleaner than the first draft! Some comments...

src/error.rs
darkmmon: Originally I put 23000 since I saw it in the error testing file, but later I found that 23000 is the sqlstate for both unique/foreign key constraint violation. Therefore I cannot use sqlstate there. I would update the error codes again once I make a foreign key constraint test. 
Thank you for spotting this though!

src/error.rs
darkmmon: Thanks for all the comments! Integrated the later two into my code.

src/error.rs
```diff
@@ -139,3 +139,85 @@ where
...
+                let error_code = e.code().unwrap_or_default();
+                let _error_code_expanded = error_code.deref();
+                #[cfg(feature = "sqlx-mysql")]
+                if e.try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()
+                    .is_some()
+                {
+                    let error_number = e
+                        .try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()?
+                        .number();
+                    match error_number {
```
billy1624, line 178: Hey @darkmmon, I wonder why for MySQL we use `error_number` to determine the error type instead of using `_error_code_expanded` just like other databases?

darkmmon: @billy1624 
The reason is that the error code for MySQL is ambiguous. If I wanted to do it with .code() like for other databases, I would not be able to distinguish the violation between unique constraint and foreign key constraint (they both return 23000). 
I can only think of using the exact error number for distinguishing these errors, leading to the clumsy code. If there is an elegant way it would definitely be welcomed!

src/error.rs
```diff
@@ -144,18 +144,17 @@ where
...
 pub enum SqlErr {
-    /// error for inserting a record with a key that already exists in the table
-    #[error("Cannot have record with same key")]
-    UniqueConstraintViolation(),
-    /// error for Foreign key is not primary key
-    #[error("Cannot add non-primary key from other table")]
-    ForeignKeyConstraintViolation(),
+    /// error for duplicate record in unique field or primary key field
+    #[error("Unique Constraint Violated")]
+    UniqueConstraintViolation,
```
tyt2y3, line 149: My Question is why remove the `()`? It is a placeholder, in case some day, we'd like to add some content.

tyt2y3: Actually, I though we shipped this enum somewhere already.
But then can we actually shove a little more context inside them?

tests/sql_err_tests.rs
```diff
@@ -39,7 +42,21 @@ pub async fn test_error(db: &DatabaseConnection) {
...
+    } else if db.get_database_backend() == DbBackend::Sqlite {
+        error_message = "UNIQUE constraint failed: cake.id"
+    }
+
+    assert_eq!(
+        error.sql_err(),
+        Some(SqlErr::UniqueConstraintViolation(String::from(
+            error_message
+        )))
+    );
```
tyt2y3, line 59: The content is out of our control, so may be 

```rust
assert!(matches!(error.sql_err(), Some(SqlErr::UniqueConstraintViolation(_)));
```

would suffice.

src/error.rs
```diff
@@ -139,3 +139,105 @@ where
 {
     DbErr::Json(s.to_string())
 }
+
+/// An error from unsuccessful SQL query
+#[derive(Error, Debug, PartialEq, Eq)]
```
tyt2y3, line 144:
```suggestion
#[derive(Error, Debug, Clone, PartialEq, Eq)]
```

PR #1708, opened 2023-06-15
Make insert statement not panic when inserting nothing

darkmmon:
## PR Info

- Closes  https://github.com/SeaQL/sea-orm/issues/873

## New Features

- [ ]  A new wrapper struct for insert statement, allowing safe return when attempt to insert nothing

## Bug Fixes

- [ ]  

## Changes

- [ ]


tyt2y3 (commented):
Just a naming thing, may be we can rename `InsertAttempt` into `TryInsert`?

src/query/insert.rs
```diff
@@ -173,6 +222,7 @@ where
     /// };
     /// assert_eq!(
     ///     cake::Entity::insert(orange)
+    ///         .on_empty_do_nothing()
```
tyt2y3, line 225: umm, why do we need to change the examples here? I think this PR should have no changes to exsiting features

src/query/traits.rs
```diff
@@ -88,3 +91,95 @@ where
         QuerySelect::column_as(self, col, alias)
     }
 }
+
+/// Insert query Trait
+pub trait InsertTrait<A>: Sized
```
tyt2y3, line 96: why do we need a new Trait here? it is not strictly necessary?

src/query/insert.rs
```diff
@@ -148,18 +82,133 @@ where
...
+    /// ```
+    /// use sea_orm::{entity::*, query::*, sea_query::OnConflict, tests_cfg::cake, DbBackend};
+    ///
+    /// let orange = cake::ActiveModel {
+    ///     id: ActiveValue::set(2),
+    ///     name: ActiveValue::set("Orange".to_owned()),
+    /// };
+    /// assert_eq!(
+    ///     cake::Entity::insert(orange)
+    ///         .on_conflict(
```
tyt2y3, line 124: OH, I get it now. It was shifted. And I suppose the exsiting impl wasn't actually changed. If that's the case, it'd be fine.

tyt2y3:
Looks great so far! Please finish up this PR, and then make a follow up PR:

So, in https://github.com/SeaQL/sea-orm/blob/master/tests/upsert_tests.rs we have:

```rust
    let res = Entity::insert_many([..])
    .on_conflict(on_conflict)
    .exec(db)
    .await;

    assert_eq!(res.err(), Some(DbErr::RecordNotInserted));
```

Can we add a `do_nothing` method that converts `Insert` into `TryInsert`, which would then 'absorbs' the `DbErr::RecordNotInserted`?

Then, we'd have (in addition to the above):

```rust
    let res = Entity::insert_many([..])
    .on_conflict(on_conflict)
    .do_nothing()
    .exec(db)
    .await;

    assert_eq!(res, Ok(TryInsertResult::Conflicted));
```

src/query/traits.rs
billy1624: Not a must, we just want to share the common methods between `Insert` and `TryInsert`. Methods such as `one`, `many`, `add`...etc.

tyt2y3: They can all implement the same methods. I don't see `InsertTrait` being used in a type signature anywhere so I think it is not needed.

billy1624: My concern is code duplication. Implement the same method twice, for two struct, `Insert` and `TryInsert`.

We can use macro, but I think `InsertTrait` to define shared methods might be better. Both are okay for me.

tyt2y3: Yes. technically `TryInsert` contains an `Insert` so any implementation can simply be delegated. I am not seeing where's the potential saving in code duplication.

tyt2y3 (approved):
Still clean, isn't it

tests/empty_insert_tests.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    let empty_iterator = [bakery::ActiveModel {
+        name: Set("SeaSide Bakery".to_owned()),
+        profit_margin: Set(10.4),
+        ..Default::default()
+    }]
+    .into_iter()
+    .filter(|_| false);
+
+    let empty_insert = Bakery::insert_many(empty_iterator)
```
tyt2y3, line 48: Oh, there is the https://doc.rust-lang.org/std/iter/fn.empty.html

PR #1719, opened 2023-06-23
[CI] Fix example fmt

billy1624:


tyt2y3:
my observation is if we don't put `--all` it won't fmt the `entity` crates

billy1624:
> my observation is if we don't put `--all` it won't fmt the `entity` crates

Indeed!

Issue #1349, opened 2023-01-02
Derive macros for custom wrapper types


tyt2y3:
Right now, custom wrapper types kind of work with SeaORM, e.g. 1) https://github.com/SeaQL/sea-orm/blob/master/tests/common/features/event_trigger.rs 
2) https://github.com/nitnelave/lldap/pull/382/files#diff-27e26a3b85b610666d7dd0a3fa3e848582ab2d3999f7b1bdfed61052f82af741R96-R168
but require a lot of boilerplate, it'd be nice if we can make a `DeriveValueType` macro and remove some entropy.

To start, we can only accept struct in the form of `struct MyType(pub T)`. It would be great if we can also accept `Vec<T>`, where I think `Nullable` can also be implemented.

And may be a `column_type` attribute to allow customizing the column type.

billy1624:
The idea is to provide a derive macro to implement the necessary traits of a custom wrapper type. For simplicity, we only support `struct MyType(T)` for now.

Without derive macro, the trait implementations look like below.

```rs
use sea_orm::entity::prelude::*;

#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
#[sea_orm(table_name = "json_vec")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub str_vec: StringVec,
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
pub enum Relation {}

impl ActiveModelBehavior for ActiveModel {}

#[derive(Clone, Debug, PartialEq, Eq)]
pub struct StringVec(pub Vec<String>);

impl From<StringVec> for Value {
    fn from(source: StringVec) -> Self {
        source.0.into()
    }
}

impl sea_orm::TryGetable for StringVec {
    fn try_get_by<I: sea_orm::ColIdx>(res: &QueryResult, idx: I) -> Result<Self, sea_orm::TryGetError> {
        <Vec<String> as sea_orm::TryGetable>::try_get_by(res, idx).map(|v| StringVec(v))
    }
}

impl sea_query::ValueType for StringVec {
    fn try_from(v: Value) -> Result<Self, sea_query::ValueTypeErr> {
        <Vec<String> as sea_query::ValueType>::try_from(v).map(|v| StringVec(v))
    }

    fn type_name() -> String {
        stringify!(StringVec).to_owned()
    }

    fn array_type() -> sea_orm::sea_query::ArrayType {
[...]

PR #1713, opened 2023-06-20
Upgrading to syn v2 

darkmmon:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1703

## New Features

- [ ]  upgraded the codes to utilize syn v2

## Breaking Changes

- [ ] Drop the use of `#[sea_orm(table_name = "col_name")]` in `DeriveColumn`

## Changes

- [ ]

PR #1720, opened 2023-06-23
Derive value type

darkmmon:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1349

- Dependencies:
  - https://github.com/SeaQL/sea-orm/pull/1713 (need to check if syn v2 will break the new macro)

- Dependents:
  - 

## New Features

- [ ]  added macro to derive for simple value type wrapper

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]


sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,86 @@
...
+        let fields: Option<syn::punctuated::Punctuated<syn::Field, syn::token::Comma>> = match dat {
+            syn::Data::Struct(syn::DataStruct {
+                fields: syn::Fields::Unnamed(syn::FieldsUnnamed { unnamed, .. }),
+                ..
+            }) => Some(unnamed),
+            _ => None,
+        };
+        if fields.clone().expect("hello").into_iter().count() != 1 {
+            panic!()
+        };
```
billy1624, line 22: Hey @darkmmon, maybe we can include a friendly notice that we expect this derive macro to be used by a tuple struct which has only 1 value in the tuple? i.e. `struct Abc(T)`

> Btw... the `.expect("hello")` looks like a debug message to be, consider changing that as well?

tyt2y3:
Looks really good so far!

sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,239 @@
...
+                } else {
+                    // Reads the value expression to advance the parse stream.
+                    // Some parameters, such as `primary_key`, do not have any value,
+                    // so ignoring an error occurred here.
+                    let _: Option<Expr> = meta.value().and_then(|v| v.parse()).ok();
+                }
+
+                Ok(())
+            })
+            .expect("msg");
```
billy1624, line 65: debug message 1

sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,239 @@
...
+                        quote! { Binary(sea_orm::sea_query::BlobSize::Blob(None)) }
+                    }
+                    _ => {
+                        // Assumed it's ActiveEnum if none of the above type matches
+                        quote! {}
+                    }
+                };
+                if col_type.is_empty() {
+                    let field_span = field.span();
+                    let ty: Type = LitStr::new(field_type, field_span).parse().expect("g");
```
billy1624, line 116: debug message 2

sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,239 @@
...
+                    "Json" => quote! { Json },
+                    "Decimal" => quote! { Decimal },
+                    _ => {
+                        // Assumed it's ActiveEnum if none of the above type matches
+                        quote! {}
+                    }
+                };
+                if arr_type.is_empty() {
+                    let field_span = field.span();
+                    let ty: Type = LitStr::new(field_type, field_span).parse().expect("g");
```
billy1624, line 164: debug message 3

billy1624 (commented):
Seems we have 3 more debug messages to "fix"

sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,239 @@
...
+                }
+
+                fn array_type() -> sea_orm::sea_query::ArrayType {
+                    #array_type
+                }
+
+                fn column_type() -> sea_orm::sea_query::ColumnType {
+                    #column_type
+                }
+            }
```
billy1624, line 228: Please change the `sea_query::xxx` namespace to `sea_orm::sea_query::xxx`. User might not have `sea_query` crate on the root.

tests/common/features/custom_wrapper.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
+pub struct Integer(pub i32);
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
+pub struct StringVec(
+    #[sea_orm(column_type = "String(Some(1))", array_type = "String")] pub Vec<String>,
+);
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
+pub struct Boolbean(#[sea_orm(column_type = "Boolean", array_type = "Bool")] pub String);
```
billy1624, line 26: Input wanted. @tyt2y3 I think it's better to "tag" the proc_macro attribute to the struct

```rs
#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
#[sea_orm(column_type = "String(Some(1))", array_type = "String")] 
pub struct StringVec(pub Vec<String>);
```

tyt2y3: I agree

tests/common/features/value_type.rs
```diff
@@ -0,0 +1,26 @@
...
+pub enum Relation {}
+
+impl ActiveModelBehavior for ActiveModel {}
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
+pub struct Integer(pub i32);
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveValueType)]
+#[sea_orm(column_type = "String(Some(1))", array_type = "String")]
+pub struct StringVec(pub Vec<String>);
```
tyt2y3, line 22: Great work. I think it looks really good. One question, is `array_type` required here? What if we omit it?

Can we actually change the test case in `tests/common/features/json_vec.rs` to use this `DeriveValueType` instead of the manual implementation?

Just want to make sure we have had end to end tests.

sea-orm-macros/src/derives/value_type.rs
```diff
@@ -0,0 +1,240 @@
...
+            field_type.as_str()
+        };
+
+        let column_type = match col_type {
+            Some(t) => quote! { sea_orm::sea_query::ColumnType::#t },
+            None => {
+                let col_type = match field_type {
+                    "char" => quote! { Char(None) },
+                    "String" | "&str" => quote! { String(None) },
+                    "i8" => quote! { TinyInteger },
```
tyt2y3, line 81: Umm... we should not copy and paste here. Can we extract the function somewhere and share the same code among different macros?

tests/common/features/value_type.rs
```diff
@@ -0,0 +1,26 @@
+use sea_orm::entity::prelude::*;
+use sea_orm_macros::DeriveValueType;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "value_type")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub number: Integer,
```
tyt2y3, line 9: The strange bit here is that the StringVec is not included in the Model for testing.
Albeit it only works on Postgres.

darkmmon: Currently, the conversion of `DeriveValueType` rely on the pre-defined conversion from the field to `Value`.
I think `Vec<String>` currently does not have a defined conversion to any variant of `Value`, which I suppose it should.
(in json_vec, it convert itself to formatted String as `Value`, which I suppose isn't a good generic approach)
I'm wondering if I should implement `From<Vec<T>> for Value` or adjust the `DeriveValueType` implementation only.

tests/json_vec_tests.rs
```diff
@@ -5,11 +5,7 @@ use pretty_assertions::assert_eq;
 use sea_orm::{entity::prelude::*, entity::*, DatabaseConnection};
 
 #[sea_orm_macros::test]
-#[cfg(any(
-    feature = "sqlx-mysql",
-    feature = "sqlx-sqlite",
-    feature = "sqlx-postgres"
-))]
+#[cfg(feature = "sqlx-postgres")]
```
tyt2y3, line 8: json vec is supposed to work with all databases.

tests/common/features/schema.rs
```diff
@@ -312,7 +312,11 @@ pub async fn create_json_vec_table(db: &DbConn) -> Result<ExecResult, DbErr> {
                 .auto_increment()
                 .primary_key(),
         )
-        .col(ColumnDef::new(json_vec::Column::StrVec).string().not_null())
+        .col(
+            ColumnDef::new(json_vec::Column::StrVec)
+                .array(sea_query::ColumnType::String(None))
```
tyt2y3, line 319: why do we have to change the column type?

tests/common/features/json_vec.rs
```diff
@@ -15,41 +15,5 @@ pub enum Relation {}
...
 impl ActiveModelBehavior for ActiveModel {}
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct StringVec(pub Vec<String>);
-
-impl From<StringVec> for Value {
-    fn from(source: StringVec) -> Self {
-        Value::String(serde_json::to_string(&source).ok().map(Box::new))
-    }
-}
```
tyt2y3, line 25: this is the bit that should be preserved

tests/common/features/json_vec.rs
```diff
@@ -15,41 +15,5 @@ pub enum Relation {}
...
-        let json_str: String = res.try_get_by(idx).map_err(TryGetError::DbErr)?;
-        serde_json::from_str(&json_str).map_err(|e| TryGetError::DbErr(DbErr::Json(e.to_string())))
-    }
-}
-
-impl sea_query::ValueType for StringVec {
-    fn try_from(v: Value) -> Result<Self, sea_query::ValueTypeErr> {
-        match v {
-            Value::String(Some(x)) => Ok(StringVec(
-                serde_json::from_str(&x).map_err(|_| sea_query::ValueTypeErr)?,
```
tyt2y3, line 38: I see why. I suspect this is not handled by the macro?

tests/common/features/value_type.rs
tyt2y3: I see. Sadly then it means that we can't 'eat our own dog food' for this test case.

tyt2y3:
So perhaps it's better to have a separate test case to test `StringVec` end to end.

tyt2y3 (approved):
Perfect

examples/salvo_example/README.md
```diff
@@ -1,6 +1,6 @@
 ![screenshot](Screenshot.png)
 
-# Poem with SeaORM example app
+# Salvo with SeaORM example app
```
billy1624, line 3: laugh :P
Good eyesight!

billy1624 (approved):
Thanks!! @darkmmon awesome work!

PR #1723, opened 2023-06-25
Update readme temp

ccqpein:
## Description

After `sea-orm-cli migrate init`, the `migration/README.md` give several useful commands like `cargo run -- up`. 

But the `generate` new file isn't right. If I understand correctly, the `migrate` subcommand is for `sea-orm-cli`.

## Changes

- [x] update the readme of template


billy1624 (approved):
Thanks!! @ccqpein nice catch!

Issue #1242, opened 2022-11-22
Missing method "find_with_linked"


nitnelave:
## Motivation

We have `find_with_related` and `find_also_linked`, but there's no `find_with_linked`.
For instance, I want to get all users who are in a group whose name is "admin" (not primary key). I have a "membership" table for the N-N relation. I would want to write:

```rust
User::find().find_also_linked(Group).filter(Group::Column::Name.eq("admin")).all(db).await?;
```

Then there's also the question of inner join vs left join. I know the answer is probably going to be pointing me to https://www.sea-ql.org/SeaORM/docs/advanced-query/custom-select/, but it's just a seemingly missing part of the API.

nitnelave:
Somewhat related, but I'm looking to get a `Vec<(User, Vec<Group>)>`.

nitnelave:
Actually, I was looking at the custom-select page, and even with custom joins and so on, I can't figure out how to get a `Vec<(User, Vec<Group>)>`: You can't derive `FromQueryResult` for a struct with `Vec` inside, and I'm not sure how to aggregate across rows.

nitnelave:
So, from what I understand, the only way to do it given the current code (if we want to leverage the link) is to use `find_also_linked` and then post process the results: we'll get a `(User, Some(Group))` for each pair of (user, group), plus a `(User, None)` for users without groups. If we sort by user, they'll be correctly grouped and we can do some stream processing. Not the prettiest but it works. I still feel like this should be done by the framework.

tyt2y3:
I think this omission is not intentional and this can be added.
@billy1624 I think it should be a matter of refactoring into `prepare_select_two_for_linked` and construct a `SelectTwoMany`?

tyt2y3:
Closed by https://github.com/SeaQL/sea-orm/pull/1743

PR #1725, opened 2023-06-27
API DEMO: `find_with_linked`

billy1624:
## PR Info

- An API demo for https://github.com/SeaQL/sea-orm/issues/1242

### DIFF
--- a/tests/relational_tests.rs
+++ b/tests/relational_tests.rs
@@ -2,6 +2,7 @@ pub mod common;
 
 pub use chrono::offset::Utc;
 pub use common::{bakery_chain::*, setup::*, TestContext};
+use pretty_assertions::assert_eq;
 pub use rust_decimal::prelude::*;
 pub use rust_decimal_macros::dec;
 use sea_orm::{entity::*, query::*, DbErr, DerivePartialModel, FromQueryResult};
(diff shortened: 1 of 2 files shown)


tests/relational_tests.rs
```diff
@@ -747,6 +748,82 @@ pub async fn linked() -> Result<(), DbErr> {
...
+            "`baker`.`bakery_id` AS `A_bakery_id`,",
+            "`r4`.`id` AS `B_id`,",
+            "`r4`.`name` AS `B_name`,",
+            "`r4`.`notes` AS `B_notes`",
+            "FROM `baker`",
+            "INNER JOIN `cakes_bakers` AS `r0` ON `r0`.`baker_id` = `baker`.`id`",
+            "INNER JOIN `cake` AS `r1` ON `r1`.`id` = `r0`.`cake_id`",
+            "INNER JOIN `lineitem` AS `r2` ON `r2`.`cake_id` = `r1`.`id`",
+            "INNER JOIN `order` AS `r3` ON `r3`.`id` = `r2`.`order_id`",
+            "INNER JOIN `customer` AS `r4` ON `r4`.`id` = `r3`.`customer_id`",
```
darkmmon, line 771: Should this be a inner join or a left join?
I found that other find_with/find_also functions use left join instead of inner join

tests/relational_tests.rs
```diff
@@ -747,6 +748,82 @@ pub async fn linked() -> Result<(), DbErr> {
...
+                    },
+                    customer::Model {
+                        id: 1,
+                        name: "Kate".into(),
+                        notes: Some("Loves cheese cake".into()),
+                    },
+                    customer::Model {
+                        id: 2,
+                        name: "Kara".into(),
+                        notes: Some("Loves all cakes".into()),
```
darkmmon, line 820: Not sure if the order of these matters or not, since they return in reverse order for my implementation

tests/relational_tests.rs
```diff
@@ -747,6 +748,82 @@ pub async fn linked() -> Result<(), DbErr> {
...
+                },
+                vec![customer::Model {
+                    id: 2,
+                    name: "Kara".into(),
+                    notes: Some("Loves all cakes".into()),
+                }]
+            ),
+            (
+                baker::Model {
+                    id: 1,
```
darkmmon, line 799: I believe this should be a 2 instead of 1?

darkmmon (changes requested):
Asking for clarification.

tests/relational_tests.rs
billy1624: Note that I have have `ORDER BY` expression
```sql
ORDER BY `baker`.`id` ASC
```

billy1624:
I will closed this PR. @darkmmon please file your PR :P

tests/relational_tests.rs
billy1624: Seems this is a follow up question of https://github.com/SeaQL/sea-orm/pull/1725#discussion_r1246117258?

tests/relational_tests.rs
billy1624: Yes, it should be `LEFT JOIN`

PR #1726, opened 2023-06-27
DeriveDisplay macro for enum

darkmmon:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1255

## New Features

- [ ]  added a new macro DeriveDisplay to allow custom display value.

## Breaking Changes

- [ ]  removed impl Display trait for enum. Add DerivedDisplay if you want.

## Changes

- [ ]


sea-orm-macros/src/derives/active_enum_display.rs
```diff
@@ -0,0 +1,129 @@
...
+                    quote! { #string }
+                } else {
+                    let ident = &variant.ident;
+                    quote! { #ident }
+                }
+            })
+            .collect();
+
+        quote!(
+            impl #ident {
```
darkmmon, line 98: My guess on the source line for `cannot find value in scope` error
I guess the issue is that the ident was somehow moved and therefore cannot be used 
However, using Tokenstream to store the info directly seems to have solved it

sea-orm-macros/src/derives/active_enum.rs
```diff
@@ -101,6 +101,8 @@ impl ActiveEnum {
                     } else if meta.path.is_ident("num_value") {
                         is_int = true;
                         num_value = Some(meta.value()?.parse::<LitInt>()?);
+                    } else if meta.path.is_ident("display_value") {
+                        Some(meta.value()?.parse::<LitStr>()?);
```
billy1624, line 105: I believe this is not needed?

tyt2y3:
Hope I did not get your question wrong, but I think `Display` should not display the underlying string or numeric value, no matter what.

So the order of precedence can be: `display_value` -> enum variant.

sea-orm-macros/src/derives/active_enum.rs
darkmmon: Since the code for DeriveActiveEnum will panic if there is some unrecognised attributes, I added this so that deriving both Display and Active Enum would not make the code panic.

src/entity/active_enum.rs
```diff
@@ -390,8 +390,8 @@ mod tests {
 
                 assert_eq!($ident::db_type(), ColumnType::$col_def.def());
 
-                assert_eq!(format!("{}", $ident::Big), "1");
-                assert_eq!(format!("{}", $ident::Small), "0");
+                assert_eq!(format!("{}", $ident::Big), "'Big'");
+                assert_eq!(format!("{}", $ident::Small), "'Small'");
```
darkmmon, line 394: Currently the fallback enums doesn't pass this test as it weirdly doesn't have the single quotes 
Will try to find the reason behind it.

tyt2y3: We expect no single quote:

```rust
assert_eq!(format!("{}", Enum::Big), "Big");
```

src/entity/active_enum.rs
```diff
@@ -408,7 +408,7 @@ mod tests {
 
     #[test]
     fn escaped_non_uax31() {
-        #[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Copy)]
+        #[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Copy, DeriveDisplay)]
```
tyt2y3, line 411: Can we have a few test cases NOT deriving DeriveDisplay where it is not needed?

Just want to make sure we have test coverage for both cases.

tyt2y3 (approved):

tests/common/features/sea_orm_active_enums.rs
```diff
@@ -53,3 +53,12 @@ pub enum MediaType {
...
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, DeriveDisplay)]
+#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "tea")]
+pub enum DisplayTea {
+    #[sea_orm(string_value = "EverydayTea", display_value = "Everyday")]
+    EverydayTea,
+    #[sea_orm(string_value = "BreakfastTea", display_value = "Breakfast")]
+    BreakfastTea,
+}
```
tyt2y3, line 64: Oh wait, actually, did I miss the test case for this? There are no corresponding asserts?

As this is the only place the `display_value` attribute is used.

tyt2y3 (approved):

tyt2y3:
Awesome

tests/common/features/sea_orm_active_enums.rs
billy1624: Here is the test case.
https://github.com/SeaQL/sea-orm/pull/1726/files#diff-a85e90f626db076a34bb42800eafe047a773a6ef5fe19d05f3a284efb3c16f3aR802-R808

billy1624 (approved):
Thanks!! @darkmmon This is clean :)

PR #1728, opened 2023-06-29
Adding find_with_linked 

darkmmon:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/1242

## New Features

- [ ]  added find_with_linked

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/src/executor/select.rs
+++ b/src/executor/select.rs
@@ -992,102 +992,39 @@ where
 }
 
 fn consolidate_query_result<L, R>(
-    rows: Vec<(L::Model, Option<R::Model>)>,
+    mut rows: Vec<(L::Model, Option<R::Model>)>,
 ) -> Vec<(L::Model, Vec<R::Model>)>
 where
     L: EntityTrait,
     R: EntityTrait,
 {
-    {
-        let pkcol = <L::PrimaryKey as Iterable>::iter()
-            .next()
-            .expect("should have primary key")
-            .into_column();
+    //todo: could take not iter
+    let pkcol = <L::PrimaryKey as Iterable>::iter()
+        .next()
+        .expect("should have primary key")
+        .into_column();
 
-        let keys: Vec<Value> = rows.iter().map(|row| row.0.get(pkcol)).collect();
-        let mut unique_keys: Vec<Value> = Vec::new();
-        for key in keys {
-            let mut existed = false;
-            for existing_key in unique_keys.clone() {
-                if key == existing_key {
-                    existed = true;
-                    break;
-                }
-            }
-            if !existed {
-                unique_keys.push(key)
-            }
-        }
-
-        let key_values: Vec<(Value, L::Model, R::Model)> = rows
-            .iter()
-            .map(|row| {
-                (
-                    row.0.clone().get(pkcol),
-                    row.0.clone(),
-                    row.1.clone().expect("should have linked entity"),
-                )
-            })
-            .collect();
-
-        let mut hashmap: HashMap<Value, (L::Model, Vec<R::Model>)> = key_values.into_iter().fold(
-            HashMap::<Value, (L::Model, Vec<R::Model>)>::new(),
-            |mut acc: HashMap<Value, (L::Model, Vec<R::Model>)>,
-             key_value: (Value, L::Model, R::Model)| {
-                acc.insert(key_value.0, (key_value.1, Vec::new()));
-
-                acc
-            },
-        );
-
-        rows.into_iter().for_each(|row| {
+    let mut hashmap: HashMap<Value, Vec<R::Model>> = rows.iter_mut().fold(
+        HashMap::<Value, Vec<R::Model>>::new(),
+        |mut acc: HashMap<Value, Vec<R::Model>>, row: &mut (L::Model, Option<R::Model>)| {
             let key = row.0.get(pkcol);
+            let value = row.1.take().expect("should have a linked entity");
+            let vec: Option<&mut Vec<R::Model>> = acc.get_mut(&key);
+            if let Some(vec) = vec {
+                vec.push(value)
+            } else {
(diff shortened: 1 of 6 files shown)


src/query/combine.rs
```diff
@@ -119,11 +119,16 @@ where
...
+    }
+
+    pub(crate) fn new_without_prepare(query: SelectStatement) -> Self {
         Self {
             query,
             entity: PhantomData,
         }
-        .prepare_select()
         .prepare_order_by()
     }
```
billy1624, line 133: Did we invoke `prepare_order_by()` method twice? When we call `SelectTwoMany::new()` it will invoke `prepare_order_by()` method twice.

tyt2y3:
Just to clarify, the implementation should involve using a hash map, instead of relying on order by (which is theoretically wrong) to group elements.

The algorithm can be referenced in https://github.com/SeaQL/sea-orm/blob/4905dc5fde00a319ca64e63bee43b0eba900c6ca/src/query/loader.rs#L208

And actually, since now Value is already Hashable, we should change the implementation of loader as well.

We should also consider chaning the existing find_with_related impl as well, but may be, not in the same PR.

src/executor/select.rs
```diff
@@ -997,6 +998,27 @@ where
     L: EntityTrait,
     R: EntityTrait,
 {
+    // #[cfg(feature = "hashable-value")]
```
darkmmon, line 1001: I think sea_query::Value is not hashable by default and a specified feature is needed 
but then the tests would need to also have the feature on in the configuration
my thought for solution is to make sea_query::Value to be hashable by default

tyt2y3: Yes the 'hashable-value' feature is intended to be add by default.

src/executor/select.rs
```diff
@@ -997,31 +998,96 @@ where
...
-    for (l, r) in rows {
-        if let Some((last_l, last_r)) = acc.last_mut() {
-            let mut same_l = true;
-            for pk_col in <L::PrimaryKey as Iterable>::iter() {
-                let col = pk_col.into_column();
-                let val = l.get(col);
-                let last_val = last_l.get(col);
-                if !val.eq(&last_val) {
-                    same_l = false;
+    {
```
tyt2y3, line 1001: Why do we need a nested scope here?

src/executor/select.rs
```diff
@@ -997,31 +998,96 @@ where
...
+                    .get(&key)
+                    .cloned()
+                    .expect("should have key in hashmap")
+            })
+            .collect();
+        result
     }
-    acc
+
+    // let mut acc: Vec<(L::Model, Vec<R::Model>)> = Vec::new();
```
tyt2y3, line 1066: if you want to keep the old code for reference, we can put it in another function `consolidate_query_result_of_ordered_rows`

src/executor/select.rs
```diff
@@ -997,31 +998,96 @@ where
...
-                if !val.eq(&last_val) {
-                    same_l = false;
+    {
+        let pkcol = <L::PrimaryKey as Iterable>::iter()
+            .next()
+            .expect("should have primary key")
+            .into_column();
+
+        let keys: Vec<Value> = rows.iter().map(|row| row.0.get(pkcol)).collect();
+        let mut unique_keys: Vec<Value> = Vec::new();
```
tyt2y3, line 1008: We can have used a HashSet to find unique values. But I doubt this is even needed for the algorithm.

src/executor/select.rs
```diff
@@ -997,31 +998,96 @@ where
...
+            .map(|row| {
+                (
+                    row.0.clone().get(pkcol),
+                    row.0.clone(),
+                    row.1.clone().expect("should have linked entity"),
+                )
+            })
+            .collect();
+
+        let mut hashmap: HashMap<Value, (L::Model, Vec<R::Model>)> = key_values.into_iter().fold(
```
tyt2y3, line 1033: It could have been `HashMap<Value (PK of L), Vec<R::Model>>`

src/executor/select.rs
```diff
@@ -997,31 +998,96 @@ where
...
+
+            let vec = &mut hashmap
+                .get_mut(&key)
+                .expect("Failed at finding key on hashmap")
+                .1;
+
+            vec.push(row.1.expect("should have value in row"));
+        });
+
+        let result: Vec<(L::Model, Vec<R::Model>)> = unique_keys
```
tyt2y3, line 1054: unique_keys is equivalent to hashmap.key() ?

tyt2y3:
@darkmmon I believe you liked performance algorithms, so may be we can optimize the implementation, here we want to minimize allocation, copy and (to some extend) move.

We observe that `rows: Vec<(L::Model, Option<R::Model>)>` is owned. We can do some operations in place, by mutating it. We can use https://doc.rust-lang.org/std/option/enum.Option.html#method.take to take the R::Model out from the vec, leaving a None in place. As such we don't have to re-allocate Vec<L::Model>.

So the intermediate hashmap can simply be `L::PK -> Option<Vec<R::Model>>`

In the third and final step, we can re-assemble everything together. Something like:

```rust
row.into_iter().filter_map(|l_model, _| {
let l_pk = pk_of(l_model);
let r_models = hashmap.get_mut(l_pk).take();
if let Some(r_models) =r_models {
    Some((l_model, r_models))
} else {
    // it means we have seen this l_model already
    None
}
}).collect()
```

The key points here:

1) L::Model should **never** be cloned, and only needs to be moved **once**.
2) R::Model should **never** be cloned, and needs to be moved *twice*, first from input into the hashmap, and second from the hashmap into the output. 

Anything more might be unnecessary overhead.

tyt2y3:
GitHub somehow did not allow me to merge into a different branch. So I did manually.

PR #1765, opened 2023-07-19
* git2 dependency optional

Iskander508:
## PR Info

Recent changes in 0.12 RC2 that introduced the `git2` dependency break iOS target build in our project. (https://github.com/rust-lang/git2-rs/issues/830). This PR aims to restrict the `git2` dependency only for the `cli` feature, avoiding the `git2` dependency for library-only uses.

Maybe a better solution can be implemented to fix these issues, the code presented here could be one possible solution.

## Changes

* `git2` crate now only dependent on the `cli` feature
*  `git2` dependency version is not limited to `0.16.x` enabling updates when sea-orm used among other dependents on `git2`

### DIFF
--- a/sea-orm-cli/src/commands/migrate.rs
+++ b/sea-orm-cli/src/commands/migrate.rs
@@ -78,6 +78,7 @@ pub fn run_migrate_command(
     Ok(())
 }
 
+#[cfg(feature = "cli")]
 pub fn run_migrate_init(migration_dir: &str) -> Result<(), Box<dyn Error>> {
     let migration_dir = match migration_dir.ends_with('/') {
         true => migration_dir.to_string(),
--- a/sea-orm-cli/Cargo.toml
+++ b/sea-orm-cli/Cargo.toml
@@ -45,7 +45,7 @@ tracing = { version = "0.1", default-features = false }
 url = { version = "2.2", default-features = false }
 chrono = { version = "0.4.20", default-features = false, features = ["clock"] }
 regex = { version = "1", default-features = false }
-git2 = { version = "0.16", default-features = false }
+git2 = { version = ">= 0.16", default-features = false, optional = true }
 
 [dev-dependencies]
 smol = "1.2.5"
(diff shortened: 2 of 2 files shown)


sea-orm-cli/src/commands/migrate.rs
```diff
@@ -78,6 +78,7 @@ pub fn run_migrate_command(
     Ok(())
 }
 
+#[cfg(feature = "cli")]
 pub fn run_migrate_init(migration_dir: &str) -> Result<(), Box<dyn Error>> {
```
tyt2y3, line 82: @billy1624 is it appropriate to put `run_migrate_init` behind a feature guard?

strozynskiw: I feel like the main problem here is the fact that git2 depends on a binary library that, when shared has to be in the same version with might make some conflicts in projects involving sea-orm and some other crates that depend on git2. Plus sea-orm, as native in rust, can be compiled on any target. Such binary dependencies might break that as they did in that case. Can the code involving git2 library be replaced with non-dependency code? From what I see it only does:
```
    if git2::Repository::discover(Path::new(&migration_dir)).is_ok() {
        write_file!(".gitignore", "_gitignore");
    }
```

strozynskiw: Although if we can avoid CLI dependency with feature flags then I think it's just fine. In the end no one will use CLI on iOS.

tyt2y3:
Yes, it's still good to make dependencies optional.

tyt2y3:
It also seems like that all we needed was a `is_dir` check on `.git`

billy1624:
Replying https://github.com/SeaQL/sea-orm/pull/1765#discussion_r1268114098

> @billy1624 is it appropriate to put `run_migrate_init` behind the feature flag? I might be wrong, but it seems `run_migrate_generate` belongs to the `cli` feature too? And since they are the only `pub` symbols, it seems like the entire `migrate.rs` can be put behind the `cli` feature flag?

- it's okay to put `run_migrate_init` behind the feature flag
- yes, `run_migrate_generate` belongs to the `cli` feature too
- also true, the entire `migrate.rs` can be put behind the `cli` feature flag

billy1624:
My suggestions
- bump the version of `git2` to v0.17
- put the entire `migrate.rs` can be put behind the `cli` feature flag

PR #1775, opened 2023-07-23
Expose sea-orm feature flags in sea-orm-migration

aidalgol:
Some features of the sea-query language are unavailable in migrations with the default feature flags, so we need to be able to enable them in sea-orm-migration.

## PR Info

- Closes N/A

## New Features
N/A

## Bug Fixes
If you have a column definition like this in your table create statement,
```rust
    .col(
        ColumnDef::new(Release::Date)
            .date()
            .default(Value::ChronoDate(None)),
    )
```
It would fail to compile, because `Value::ChronoDate` comes from `sea-query`, and is behind the `with-chrono` feature flag.

As a workaround, at least when using SQLite, `Value::String` can be used instead, but then the migration is not quite correct.

## Breaking Changes
This should not break anything.

## Changes
N/A


tyt2y3 (approved):
Thank you for the contribution

PR #1792, opened 2023-07-31
Replace `dotenv` with `dotenvy`

billy1624:
## Housekeeping

- [ ] Depends on a actively maintained fork of `dotenv`


tyt2y3:
For Seaography example, we need to change Seaography itself.
We should not touch the old issues.

PR #1829, opened 2023-08-28
sea-orm-cli migrate init: don't overwrite files by default

chayleaf:
Don't overwrite files in existing directories with `sea-orm-cli migrate init`. Empty directories are still allowed by default, but not ones with any files. `--force` flag is provided to simulate the old behavior.


tyt2y3 (approved):
Thank you. I think this is fair default behaviour.

PR #1863, opened 2023-09-17
feat: implement IntoSimpleExpr for T: Into<SimpleExpr>

ho-229:
## Changes

- [ x] implement IntoSimpleExpr for T: Into<SimpleExpr>

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -106,6 +106,14 @@ impl IntoSimpleExpr for SimpleExpr {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Expurple (commented):
Hi! We are going to do this in SeaORM 2.0. There are a few more things that I'd like to do here. See https://github.com/SeaQL/sea-orm/discussions/2327#discussioncomment-13255559. If you're still interested, you can update the PR to include those things, and we'll merge it

PR #1834, opened 2023-09-01
Add Support For Decoding To Vec<T> For JSON Columns

anshap1719:
## PR Info
- Closes #1517

## New Features

- [x] Add ability to decode to concrete Vec types for JSON[] columns.

## Changes

- [x] Add `ActiveEnumOfJson` trait that is implemented by `ActiveEnum` and `JSON` types.
- [x] Auto implement `TryGetable` for any `T` where `T` implements `ActiveEnumOfJson`.
- [x] Auto implement `TryGetable` for any `Vec<T>` where `T` implements `ActiveEnumOfJson`, when array feature is enabled.
- [x] Add conditional clause in `FromJsonQueryResult` macro to implement the `NotU8` trait for the target struct/enum when `postgres-arrays` feature is enabled.
- [x] Update `DeriveActiveEnum` macro to implement `ActiveEnumOfJson` for ActiveEnums.
- [x] Add `ActiveEnumOfJson` implementation for JSON type.
- [x] Add tests

Issue #1517, opened 2023-03-01
Add support for root arrays in JSON in SeaORM


a14e:
## Description

Greetings! I've been trying to use jsonb with a root array object in SeaORM, but it appears that the current implementation doesn't support it.

## Steps to Reproduce

```rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "my_model")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: String,
    #[sea_orm(column_type = Json)]
    pub issue_tags: Vec<IssueTag>
}

#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
pub enum Relation {}

impl ActiveModelBehavior for ActiveModel {}

#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, FromJsonQueryResult)]
pub struct IssueTag {
    name: String,
    time: chrono::DateTime<Utc>,
}
```

### Expected Behavior
I expect the code to compile without errors.

### Actual Behavior
Instead, I encounter the following error:
```
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
   |                                   ^^^^^^^^^^^^^^^^^ the trait `From<Vec<IssueTag>>` is not implemented for `sea_orm::Value`
 ```
It would be great if SeaORM could support root array objects in JSON. Thank you for your attention to this matter!

**Comment by t348575** [comment]
I also want this, in the mean time this simple workaround works (I use postgres) @a14e :
Get stored in the db as if its a Vec<IssueTagData>
```rs
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, FromJsonQueryResult)]
pub struct IssueTag(pub Vec<IssueTagData>);

#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, FromJsonQueryResult)]
pub struct IssueTagData {
    pub name: String,
    pub time: chrono::DateTime<Utc>
}
```

anshap1719:
@a14e @t348575 

I was facing similar issue and created a PR today but I didn't realize this issue already exists. Updated my PR to reference this issue and also added support for root array objects. Can you also confirm that the PR works for your case by testing locally?

anshap1719:
@billy1624 @tyt2y3  If this is something you guys don't mind supporting, I'd be happy to create a PR.

tyt2y3:
[...]

PR #1898, opened 2023-10-04
Rework ActiveEnum to support TryGetable for Vec<Json>

tyt2y3:
Succeeds #1834

Step 1: remove the reliance on ValueVec


tyt2y3:
@anshap1719 I think I figured out something within the type system. At least it compiles, so it won't be too far away from complete. I can't really explain my course of thinking, but it wandered a bit and ah-ha!

If you have time, can you help verify and perhaps add back some of the test cases?

anshap1719:
@tyt2y3 Perfect! Thank you. My first approach was to try to get rid of ValueVec, but I didn't persue it because of lack of a holistic picture of the project.

In any case, I'll re-add the test files and fixtures this weekend if you haven't added them already by then.

sea-orm-macros/src/derives/active_enum.rs
```diff
@@ -337,6 +337,16 @@ impl ActiveEnum {
                 }
             }
 
+            #[automatically_derived]
+            impl sea_orm::TryGetableArray for #ident {
+                fn try_get_by<I: sea_orm::ColIdx>(res: &sea_orm::QueryResult, index: I) -> std::result::Result<Vec<Self>, sea_orm::TryGetError> {
+                    <<Self as sea_orm::ActiveEnum>::Value as sea_orm::ActiveEnumValue>::try_get_vec_by(res, index)?
+                        .into_iter()
+                        .map(|value| Self::try_from_value(&value).map_err(Into::into))
```
anshap1719, line 345: @tyt2y3 This needs to be changed from `Self::try_from_value(&value)` to `<Self as sea_orm::ActiveEnum>::try_from_value(&value)`

anshap1719 (changes requested):
@tyt2y3 I'm not sure how I feel about having to manually implement `NotU8` for types where I need to use them as root JSON arrays. It feels like that's an internal implementation detail that shouldn't be exposed to users of the crate.

Not sure what the solution would be either, in my mind the straightforward one would be to have a derive macro that either just implements NotU8 for target, or something like `FromJsonQueryArrayResult` that internally just derives `FromJsonQueryResult` as well as `NotU8`.

anshap1719:
Oh by the way, apart from things that I mentioned above, this does solve the original issue. I'm able to use root JSON arrays just fine with this branch.

sea-orm-macros/src/derives/active_enum.rs
tyt2y3: I see!

tyt2y3:
How cool is that! Let's get this done this few days.

anshap1719:
@tyt2y3 Created https://github.com/SeaQL/sea-orm/pull/1900 for adding tests and fixing issue I pointed above.

tyt2y3:
I figured we could make this work even on SQLite, provided we enable the `postgres-array` feature flag. Not idea, but Good enough for now.

anshap1719:
Hi @tyt2y3 Sorry for not being able to get back in time. The changes you had to make in `column.rs` was precisely why it was taking me time because insert tests weren't working for me locally.

Thank you for merging the PR and continuing on with necessary changes.

tyt2y3:
Is it going to pass this time? I am thrilled

tyt2y3:
image

anshap1719:
@IgnisDa Yes. I have already confirmed locally that it solves #1517

IgnisDa:
Awesome! Any ETA as to when it would be merged?

anshap1719 (approved):

anshap1719:
Thanks for all the efforts @tyt2y3

IgnisDa:
@anshap1719 What underlying data type are you using for this column?

tyt2y3:
Thank you everyone. I am happy for what we've achieved out of this collaboration!

anshap1719:
@tyt2y3 Do you know when we can expect the next release? Not trying to push for it, just want to understand the timeline. Thanks.

PR #1900, opened 2023-10-08
Add Tests And Fix Build Issue For "Rework active enum" PR

anshap1719:
## PR Info
This PR adds tests and fixes a build issue on top of #1898 that occurs when using the sea-orm crate in a project with the new code.

- Dependents:
  - #1898


tests/json_vec_tests.rs
```diff
@@ -5,16 +5,14 @@ use pretty_assertions::assert_eq;
 use sea_orm::{entity::prelude::*, entity::*, DatabaseConnection};
 
 #[sea_orm_macros::test]
-#[cfg(any(
-    feature = "sqlx-mysql",
-    feature = "sqlx-sqlite",
-    feature = "sqlx-postgres"
-))]
```
tyt2y3, line 12: I think we should be able to keep these test cases for all backend?

tests/json_vec_tests.rs
```diff
@@ -58,8 +56,37 @@ pub async fn insert_json_vec_derive(db: &DatabaseConnection) -> Result<(), DbErr
...
+        .filter(json_vec_derive::json_string_vec::Column::Id.eq(json_vec.id))
+        .one(db)
+        .await?;
+
+    assert_eq!(model, Some(json_vec));
+
+    Ok(())
+}
+
+pub async fn insert_json_struct_vec_derive(db: &DatabaseConnection) -> Result<(), DbErr> {
```
tyt2y3, line 69: did I miss it somewhere, but I don't see this function being used?

tyt2y3:
I'll try this

Issue #1931, opened 2023-10-24
Update entity generation to preserve customisations


dave42w:
Following a helpful discussion on Discord I would like to code an upgrade to sea-orm-cli generate entity, if the project is willing to merge a pull request.

My idea is that when an entity is regenerated the custom sections of code that we have written should not be lost.

My experience, so far, is only with modifying `impl ActiveModelBehavior for ActiveModel {}` but I imagine there will be other areas, such as Relations, where we will want modifications to be kept when the entity is regenerated.

I am not proposing adding great intelligence, so it will not remove the need for manual changes if the database schema changes. However, it would save manual work at present without enforcing fixed conventions around how the code base is organised.

I would modify `Sea-orm-cli generate`  to first read the pre-existing entities looking for specific comments used to mark sections to be preserved. These would be kept and after the new entity has been written these preserved sections would be inserted to replace the marked sections. 

As a fail safe preserved code would be appended if the correct place can't be found (and written to a new file if the original entity file has disappeared (renamed table for example).

So for example standard entity currently ends with:
```
impl ActiveModelBehavior for ActiveModel {}
```

To preserve changes a customised version of this would be marked by specific comments like this
```
// ** modified from generated code start ** impl ActiveModelBehavior for ActiveModel {}

use chrono::Utc;
use sea_orm::{prelude::async_trait::async_trait, ActiveValue};

#[async_trait]
impl ActiveModelBehavior for ActiveModel {
    async fn before_save<C: ConnectionTrait>(
        mut self,
        _db: &C,
        insert: bool,
    ) -> Result<Self, DbErr> {
        if insert {
            self.id = ActiveValue::NotSet;
            self.created_timestamp = ActiveValue::Set(Utc::now().to_string());
        };
        self.updated_timestamp = ActiveValue::Set(Utc::now().to_string());
        Ok(self)
    }
}

#[derive(DeriveIntoActiveModel, Serialize, Deserialize)]
pub struct BasicSong {
    pub id: i32,
    pub collection_id: i32,
    pub number: i32,
    pub title: String,
}

// ** modified from generated code end ** impl ActiveModelBehavior for ActiveModel {}
```

I am thinking that if the code after the start and end comment markers is the same then we are replacing a single line. If they differ then we start inserting our preserved code at the first and the look for the first occurrence of the end code.

eg to replace a single line with a section of code to be preserved
```
// ** modified from generated code start ** impl ActiveModelBehavior for ActiveModel {}
...
...
...
[...]

PR #1947, opened 2023-10-30
sea-orm-cli: allow skipping impl ActiveModelBehavior for generated entities

waynr:
This PR adds an option that enables users to skip generating the

```
```

by passing the `--gen-impl-active-model-behavior false` flag to `sea-orm-cli generate entity`.

This is intended to allow for an easy and idiomatic-for-rust alternative to https://github.com/SeaQL/sea-orm/issues/1931 which I described in more detail in [a related discussion post](https://github.com/SeaQL/sea-orm/discussions/1932#discussioncomment-7413992)

I don't necessarily think this should close out #1931, but I do think that the complexity of the solution described in that issue merits some consideration given the available alternative.

### DIFF
--- a/sea-orm-codegen/src/entity/writer.rs
+++ b/sea-orm-codegen/src/entity/writer.rs
@@ -50,6 +50,7 @@ pub struct EntityWriterContext {
     pub(crate) enum_extra_derives: TokenStream,
     pub(crate) enum_extra_attributes: TokenStream,
     pub(crate) seaography: bool,
+    pub(crate) gen_impl_active_model_behavior: bool,
 }
 
 impl WithSerde {
@@ -148,6 +149,7 @@ impl EntityWriterContext {
         enum_extra_derives: Vec<String>,
         enum_extra_attributes: Vec<String>,
         seaography: bool,
+        gen_impl_active_model_behavior: bool,
     ) -> Self {
         Self {
             expanded_format,
@@ -163,6 +165,7 @@ impl EntityWriterContext {
             enum_extra_derives: bonus_derive(enum_extra_derives),
             enum_extra_attributes: bonus_attributes(enum_extra_attributes),
             seaography,
+            gen_impl_active_model_behavior,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


waynr:
The current state of this PR only adds the CLI flag that maintains the current default behavior and updates `EntityWriter` test cases to include the new option. If maintainers are open to this approach I'd be happy to add some new test cases to validate the `false` code path.

dave42w:
I think this is an excellent solution.

waynr:
Something I didn't realize when I opened this PR is that it may be a little problematic since it's all-or-nothing whereas it may be desirable for the empty impl blocks to exist for some of the generated entities since some `ActiveModel` trait methods have `ActiveModelBehavior` as a trait bound: https://docs.rs/sea-orm/latest/sea_orm/entity/trait.ActiveModelTrait.html

tyt2y3:
The implementation looks good, and this feature is definitely useful!

jondot:
Hi this great, much needed, can we merge this?

tyt2y3:
@waynr do you have anything to amend?

LockedThread:
Any updates?

waynr:
Oh I don't have any additional changes. Sorry, I lost track of this after I started a new job earlier this year.

waynr:
I've rebased onto master branch and resolved the merge conflict so this should be ready to merge.

CloverOS:
When will it be updated? I really need this.

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -1667,6 +1706,29 @@ mod tests {
                 })
                 .to_string()
             );
+            assert_eq!(
+            >>>>>>> 4e74dde0 (sea-orm-cli: allow skipping impl ActiveModelBehavior for generated entities)
```
mccormickt, line 1710: There are several merge artifacts throughout this PR that should be cleaned up beforehand - but it would be great to see this merged in soon!

TheSamsa:
If nobody minds I created a copy of the original PR, since there was no progression.
I do not intend to steal anything, it is the work of @waynr he deserves the credit.
The reason why I copied this is, because there was no progress since october 2024.
Here is the PR: #2487

tyt2y3:
closed in favor of https://github.com/SeaQL/sea-orm/pull/2487 

thank you all

PR #1934, opened 2023-10-24
Support `--enum-extra-derives`

TmLev:
## PR Info

Doesn't close any issues (or I didn't find them). 

Supersedes #1392.

## New Features

- [x] New `--enum-extra-derives` CLI option allows for adding custom derives to active enums.

## Bug Fixes

None.

## Breaking Changes

Fully backward compatible for the end user.

## Changes

Made `bonus_derives` `pub(crate)` to use its functionality in the test.

PR #1952, opened 2023-11-01
Feature: Adds sea-orm-cli generate `--enum-extra-attributes` option

onefifth:
## PR Info

Adds an optional `--enum-extra-attributes` argument to sea-orm-cli generate entity. Very similar to the `--model-extra-attributes` option, it allows users to specify additional attributes to be applied to generated active enums.

Functionally related to, but not overlapping with #1934.

## New Features

- [x] Adds `--enum-extra-attributes` option to CLI `generate entity` allowing for additional attributes on generated active enums.

## Changes

- Made bonus_attributes pub(crate) so it could be used in tests.

### DIFF
--- a/sea-orm-cli/src/cli.rs
+++ b/sea-orm-cli/src/cli.rs
@@ -282,6 +282,13 @@ pub enum GenerateSubcommands {
 
         #[arg(
             long,
+            value_delimiter = ',',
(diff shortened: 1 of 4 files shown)


TmLev:
Just had to do this myself :)

tyt2y3:
I almost thought this is a duplicate of 1934. But no haha, I will resolve the conflict after merging 1934

PR #1954, opened 2023-11-02
Add #[sea_orm(skip)] for FromQueryResult macro

wyatt-herkamp:
```rust
#[derive(FromQueryResult)]
struct FromQueryAttributeTests {
    #[sea_orm(skip)]
    _foo: i32,
    _bar: String,
}
```
This will skip _foo then use Default to get the value. 

### Use Case
```rust
#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, FromQueryResult)]
pub struct PublicUser {
    pub id: i64,
    pub name: String,
    pub username: String,
    pub avatar: String,
    pub group: Group,
    pub banned: bool,
    #[serde(skip_serializing_if = "Vec::is_empty")]
    #[sea_orm(skip)]
    pub connections: UserConnections,
    pub created: DateTimeWithTimeZone,
}
```
So my UserConnections. That is a separate Vec. That I will grab later if needed. 

So if you have table that has many that relates. 
## New Features

- [x] Add the ability to skip a field in FromQueryResult

### DIFF
--- a/examples/rocket_example/api/Cargo.toml
+++ b/examples/rocket_example/api/Cargo.toml
@@ -11,12 +11,8 @@ async-trait = { version = "0.1" }
 rocket-example-service = { path = "../service" }
 futures = { version = "0.3" }
 futures-util = { version = "0.3" }
-rocket = { version = "0.5.0-rc.1", features = [
-  "json",
-] }
-rocket_dyn_templates = { version = "0.1.0-rc.1", features = [
-  "tera",
-] }
+rocket = { version = "0.5.0-rc.4", features = ["json"] }
+rocket_dyn_templates = { version = "0.1.0-rc.1", features = ["tera"] }
 serde_json = { version = "1" }
 entity = { path = "../entity" }
 migration = { path = "../migration" }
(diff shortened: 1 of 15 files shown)


tyt2y3:
This is very much welcomed!
Just skimmed through the implementation it looks pretty, but I will review it again.

wyatt-herkamp:
> Thank you for the contribution. This is very much welcomed! Just skimmed through the implementation it looks pretty, but I will review it again.

I don't know why the test is failing but I will be looking into it.

wyatt-herkamp:
The dependencies for the example rocket_okapi_example have not updated. So it cannot be updated.

wyatt-herkamp:
@tyt2y3 Hey I was wondering should I add `#[sea_orm(skip = "path::to_default::function")]` to this pr? It would be an optional value?

tyt2y3:
I think falling back onto `Default::default` is a good default. If user wants to customize the `default` function, I think it's better to have another attribute.

e.g. `#[sea_orm(skip_with="path::to_default::function")]`

or may be:

e.g. `#[sea_orm(skip, default="path::to_default::function")]`

I will merge this for now. Feel free to open another PR!

tyt2y3 (commented):
Thank you

PR #1979, opened 2023-11-21
`expr_as()`: accept and return `Self` by value

Expurple:
I was writing a complex select query and was annoyed by ownership errors forcing me to add noisy `.to_owned()` after every `.expr_as()`:

```rust
some::Entity::find()
    .select_only()
    .expr_as(..., ...)
    .to_owned()
    .column(...)
    .expr_as(..., ...)
    .to_owned()
    .column_as(..., ...)
    ...
```

This didn't look typical for the `sea-orm` API, so I dug into it, and it's clearly an accident. *All* other methods in that file accept `self` by value, except for `query()` accessor. That's the convention.

So I went and changed `expr_as()` to do that as well. It's a small change that may not be backwards-compatible in theory (couldn't find any info on that), but it didn't break any tests or any of my queries in practice. The old code still compiles as is, but can now be also shortened to

```rust
some::Entity::find()
    .select_only()
    .expr_as(..., ...)
    .column(...)
    .expr_as(..., ...)
    .column_as(..., ...)
    ...
```

### DIFF
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -511,7 +511,7 @@ pub trait QuerySelect: Sized {
     ///     "SELECT `cake`.`id`, `cake`.`name`, UPPER(`cake`.`name`) AS `name_upper` FROM `cake`"
     /// );
     /// ```
-    fn expr_as<T, A>(&mut self, expr: T, alias: A) -> &mut Self
+    fn expr_as<T, A>(mut self, expr: T, alias: A) -> Self
     where
         T: Into<SimpleExpr>,
         A: IntoIdentity,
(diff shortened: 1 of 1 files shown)


tyt2y3:
Thank you for catching this. But you're also right, this is a breaking change, which we shouldn't land right now. Here's what I can do: create a new method for now, and publish a minor release, so you can use it meanwhile.

tyt2y3:
Hope it help! https://github.com/SeaQL/sea-orm/releases/tag/0.12.7

Expurple:
LGTM, thanks. So, you gather some new features for `0.13.0`, go through `FIXME`s before the release, and land the breaking change in `0.13.0`?

tyt2y3:
Yup! that's the plan. I'd would batch several breaking changes in one release.

PR #1881, opened 2023-09-27
feat: Add proxy connection type

langyo:
## PR Info

- Closes #1875 

## New Features

- [ ] Add the proxy connection type.

## Changes

- [ ] Add a new feature called `proxy` that includes three sub features `proxy-mysql`, `proxy-postgresql` and `proxy-sqlite`.

PR #1980, opened 2023-11-24
refactor: Update GlueSQL and SurrealDB examples.

langyo:
## PR Info

- Dependencies:
  - #1881

## Changes

- [x] Bump gluesql to 0.15 for the proxy connection's example which uses gluesql.
- [x] For `GluesSQL` and `SurrealDB` example, use `SQLParser` dynamic parsing instead of hard-coded string replacement logic.

cc @tyt2y3 😉


tyt2y3:
Thanks

PR #1999, opened 2023-12-08
Add source annotations to errors

bouk:
## New Features

- [x] Add `#[source]` annotation to DbError and RuntimeErr to expose underlying errors


tyt2y3:
Thanks. Can you share a bit what you need it for? `.into()` ?

bouk:
See the [thiserror](https://github.com/dtolnay/thiserror) docs, this makes it so the Error.cause method exposes the underlying error.

PR #2002, opened 2023-12-09
Cast enums in `is_in` and `is_not_in` (#1527)

### LINKED ISSUE #1527: is_in and is_not_in don't use CAST when comparing enums.
## Description

`.is_in(...)` and `.is_not_in(...)` don't use CAST when comparing enums.

## Steps to Reproduce

1. Create a table in Postgres, that has a field with custom type as enum.
2. Make a query to that table, having a condition of this enum field IN vector.

### Expected Behavior

The same as with `.eq`, `.ne` operators. Resulting query should look something like this.
```sql
-- For .eq operator, it generates correctly
WHERE "table"."enum_field" = CAST($1 AS enum_type)
-- For .is_in operator, it generates a different query without CAST
WHERE "table"."enum_field" IN (CAST($1 AS enum_type), CAST($2 AS enum_type))
```

### Actual Behavior

Generated query looks something like this
```sql
WHERE "table"."enum_field" IN ($1, $2)
```
`PgDatabaseError` is returned.
```rust
{severity: Error, code: "42883", message: "operator does not exist: enum_type <> text", ...}
```

### Reproduces How Often

Always, I assume. But the bug was spoted when in subquery.

### Workarounds

Well, it is possible to replace `.is_in` operator with 
```rust
Condition::any()
    .add(table::Column::EnumField.eq(EnumType::Option1)
    .add(table::Column::EnumField.eq(EnumType::Option2)
```
This thing works just fine. The generated query looks like this
```sql
WHERE
      "table"."enum_field" = CAST($1 AS enum_type) 
      OR
      "table"."enum_field" = CAST($2 AS enum_type)
```

## Reproducible Example

I can do it by request, if anybody fails to reproduce the issue.

## Versions
sea-orm v0.11.0
postgres
macos Ventura

Expurple:
Fixes #1527

### DIFF
--- a/src/entity/column.rs
+++ b/src/entity/column.rs
@@ -61,7 +61,8 @@ macro_rules! bind_vec_func {
             V: Into<Value>,
             I: IntoIterator<Item = V>,
         {
-            Expr::col((self.entity_name(), *self)).$func(v)
+            let v_with_enum_cast = v.into_iter().map(|v| self.save_as(Expr::val(v)));
+            Expr::col((self.entity_name(), *self)).$func(v_with_enum_cast)
         }
     };
 }
--- a/tests/active_enum_tests.rs
+++ b/tests/active_enum_tests.rs
@@ -101,10 +101,7 @@ pub async fn insert_active_enum(db: &DatabaseConnection) -> Result<(), DbErr> {
     assert_eq!(
         model,
         Entity::find()
-            .filter(
-                Expr::col(Column::Tea)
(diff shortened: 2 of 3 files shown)


tyt2y3:
Thank you for you contribution!

Two things: can you add new test cases and keep the existing ones unchanged? More converage is better.
Can you also add some assertions against the generated SQL ?

Expurple:
Sure. Have another look now

tyt2y3 (approved):
Thank you

PR #2014, opened 2023-12-17
Cast enums in `lt()`, `lte()`, `gt()`, `gte()`

Expurple:
Revert 10f3de0, this was discussed under #1527. There weren't any tests for these operators in that initial PR (#258). Let me know if mine can't be merged without adding some. If that's the case, I'm not sure where they should go, bloating `tests::active_enum_tests::insert_active_enum` doesn't feel right.

### DIFF
--- a/src/entity/column.rs
+++ b/src/entity/column.rs
@@ -20,18 +20,6 @@ pub struct ColumnDef {
 }
 
 macro_rules! bind_oper {
-    ( $op: ident ) => {
-        #[allow(missing_docs)]
-        fn $op<V>(&self, v: V) -> SimpleExpr
-        where
-            V: Into<Value>,
-        {
-            Expr::col((self.entity_name(), *self)).$op(v)
-        }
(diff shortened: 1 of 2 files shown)


tyt2y3:
Thank you. Yes, tests are important. I guess the best way is probably for you to add a new table under `tests/common/features`.

Expurple:
Ok, so I tried to add a new table and use it for tests. Then I realized that comparison results depend on the enum's `db_type` annotation (whether it's an `Integer`, `String` or `Enum`) and that existing `ActiveEnum` table conveniently contains columns for all three enum representations. So I ended up using that.

Postgres and SQLite look fine, but I got surprising results for MySQL (see the code comments). If you don't mind, I would like to merge this as is to get Postgres working and create a separate issue for MySQL, which I don't want to dig into at the moment.

tyt2y3 (approved):
Happy new year and thank you!

PR #2026, opened 2023-12-19
Community: add Loco.rs framework

jondot:


tyt2y3:
Hi, That's very cool and wanted! I have always been day-dreaming of a "Rust on Rail" and your project seems promising!

If you have ideas that'd make the integration of SeaORM with Loco more seamless, feel free to let me know.

PR #2030, opened 2023-12-21
refactor: remove order by clause when counting rows to improve performance

shogo-nakano-desu:
## Purpose of this PR
To enhance performance. I have added a process to remove the order by clause in the `num_items` method of the `Paginator` trait, which is not needed to count rows.

## Performance Measurement Results
Upon executing queries on test data available to me, **a performance improvement of approximately three times** was observed when the order by clause was removed. (Unfortunately, I cannot provide the data used here, but I hope for your understanding.) Testing environment is PostgreSQL 14.7.

### DIFF
--- a/src/executor/paginator.rs
+++ b/src/executor/paginator.rs
@@ -71,7 +71,12 @@ where
             SelectStatement::new()
                 .expr(Expr::cust("COUNT(*) AS num_items"))
                 .from_subquery(
-                    self.query.clone().reset_limit().reset_offset().to_owned(),
+                    self.query
+                        .clone()
+                        .reset_limit()
+                        .reset_offset()
+                        .clear_order_by()
+                        .to_owned(),
                     Alias::new("sub_query"),
                 ),
         );


shogo-nakano-desu:
@tyt2y3 
First and foremost, I would like to express my gratitude for creating and maintaining such an excellent crate. Thank you very much.

At your convenience, could you please review this PR? We are currently using sea-orm in our production environment, and the changes proposed in this PR are expected to significantly contribute to resolving performance bottlenecks. Your assistance in reviewing would be greatly appreciated. Thank you for your time and consideration.

tyt2y3:
Thank you for the contribution. I think this is the right thing to do. Although just being curious, what performance gains you saw on a typical vs worst case query?

shogo-nakano-desu:
Thank you so much for your review. 

## Table info (Our testing environment)
- 14 columns
- 8M rows
- The field used in the ORDER BY clause does not have an index.
＊In our production environment, there are around 50M rows and the number is increasing rapidly.

## Result in our testing environment
### Worst case
Execution Time: 2740 ms -> 840 ms

### Typycal case
In the production environment, handling a significantly larger volume of data is expected. Therefore, it is anticipated that the worst-case scenario in terms of data quantity in the test environment would be equivalent to a typical case in the production environment. As a result, we have not measured the typical case in the test environment.

tyt2y3:
Interesting, good to know!

> The field used in the ORDER BY clause does not have an index.

That's usually the culprit haha

shogo-nakano-desu:
Now I am pushing a commit to fix Clippy error. It will be reflected when GitHub status is back to normal. https://www.githubstatus.com/

Screenshot 2024-01-09 at 22 38 25

tyt2y3:
GitHub is back now

shogo-nakano-desu:
Great, I pushed my commits. Now I think this PR will pass CI.

src/executor/paginator.rs
```diff
@@ -71,7 +71,12 @@ where
             SelectStatement::new()
                 .expr(Expr::cust("COUNT(*) AS num_items"))
                 .from_subquery(
-                    self.query.clone().reset_limit().reset_offset().to_owned(),
+                    self.query
+                        .clone()
+                        .reset_limit()
+                        .reset_offset()
+                        .clear_order_by()
+                        .to_owned(),
```
shogo-nakano-desu, line 79: Only these lines are the main change. Others are changes to fix clippy&compile errors.

shogo-nakano-desu:
I have to fix `error: associated function `accepts` is never used` clippy errors in `src/driver/sqlx_xxx.rs`

tyt2y3:
Sad, those clippy warnings seem to be very confusing

tyt2y3:
Can you reset hard this PR to 3ef07ab611a42a98 ? I will merge that first.
We can work on the clippy fixes later.

shogo-nakano-desu:
Sure, that's a good idea. I will create another PR to fix clippy errors.

PR #2037, opened 2023-12-29
Add cursor pagination desc order

ziimakc:
## New Features

Add cursor pagination desc order support

Usage:

`cursor_by().asc()` or `cursor_by().desc()`

---

Moved query builder logic inside `all` because it depends on sort order `asc/desc`

### DIFF
--- a/src/executor/cursor.rs
+++ b/src/executor/cursor.rs
@@ -2131,6 +2131,51 @@ mod tests {
     }
 
     #[smol_potat::test]
+    async fn related_composite_keys_1_desc() -> Result<(), DbErr> {
+        let db = MockDatabase::new(DbBackend::Postgres)
+            .append_query_results([[(
+                test_base_entity::Model {
+                    id: 1,
+                    name: "CAT".into(),
+                },
+                test_related_entity::Model {
+                    id: 1,
+                    name: "CATE".into(),
+                    test_id: 1,
+                },
+            )]])
+            .into_connection();
+
+        assert!(!test_base_entity::Entity::find()
+            .find_also_related(test_related_entity::Entity)
+            .cursor_by((test_base_entity::Column::Id, test_base_entity::Column::Name))
+            .last(1)
+            .desc()
+            .all(&db)
+            .await?
+            .is_empty());
+
+        assert_eq!(
+            db.into_transaction_log(),
+            [Transaction::many([Statement::from_sql_and_values(
+                DbBackend::Postgres,
+                [
+                    r#"SELECT "base"."id" AS "A_id", "base"."name" AS "A_name","#,
+                    r#""related"."id" AS "B_id", "related"."name" AS "B_name", "related"."test_id" AS "B_test_id""#,
+                    r#"FROM "base""#,
+                    r#"LEFT JOIN "related" ON "base"."id" = "related"."test_id""#,
+                    r#"ORDER BY "base"."id" ASC, "base"."name" ASC, "related"."id" ASC, "related"."name" ASC LIMIT $1"#,
+                ]
+                .join(" ")
+                .as_str(),
+                [1_u64.into()]
+            ),])]
+        );
+
+        Ok(())
+    }
+
+    #[smol_potat::test]
     async fn related_composite_keys_2() -> Result<(), DbErr> {
         let db = MockDatabase::new(DbBackend::Postgres)
             .append_query_results([[(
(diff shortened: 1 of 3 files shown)


tests/cursor_tests.rs
```diff
@@ -105,9 +106,25 @@ pub async fn cursor_pagination(db: &DatabaseConnection) -> Result<(), DbErr> {
...
+
+    cursor.before(5);
+
+    assert_eq!(
+        cursor.first(4).all(db).await?,
+        [
+            Model { id: 9 },
+            Model { id: 8 },
+            Model { id: 7 },
+            Model { id: 6 },
```
ziimakc, line 121: Not sure yet why it gives me `10-7` instead of `9-6`

ziimakc:
It seems more complicated then I expected, for example for the following cases

```
first=2,after=5,asc 6,7 is id>5 ASC
last=2,after=5,asc 9,10 is id>5 DESC + reverse result
first=2,before=5,asc 3,4 is id<5 DESC + reverse result
last=2,before=5,asc 1,2 is id<5 ASC

first=2,after=5,desc 4,3 is id<5 DESC
last=2,after=5,desc 2,1 is id<5 ASC + reverse result
first=2,before=5,desc 7,6 is id>5 ASC + reverse result
last=2,before=5,desc 10,9 is id>5 DESC
```

We can say rules is that:
- reverse order in sql query when `last`, `last+after` or `first+before` is used
- if order reversed, reverse results
- use `>` when `after+asc` or `before+desc` is used, otherwise `<`

But we need to somehow apply it to cases where both `after` and `before` is used:
```
first=2,after=5,before=9,asc 6,7 is id>5 and id<9 ASC
last=2,after=5,before=9,asc 7,8 is id>5 and id<9 DESC + reverse result
last=2,after=5,before=1,desc 3,2 is id<5 and id>1 ASC + reverse result
first=2,after=5,before=1,desc 4,3 is id<5 and id>1 DESC
```

ziimakc:
Ok rules is seems the following:
- reverse order in sql query when `last` `last+after` or `first+before` or `last+after+before` is used
- if order reversed, reverse results
- use `>` when `after+asc` or `before+desc` or `after+before+asc` is used, otherwise `<`

ziimakc:
Will add tests a bit later

tyt2y3 (commented):
Happy new year. Thank you! This is sure a lot of effort, and it seems to be working. I am not too concerned as long as our test coverage increases! (which leads to some comments below)

src/executor/cursor.rs
```diff
@@ -981,17 +1592,106 @@ mod tests {
...
+                ]
             ),])]
         );
 
         Ok(())
     }
 
     #[smol_potat::test]
-    async fn composite_keys_5() -> Result<(), DbErr> {
+    async fn composite_keys_5_desc() -> Result<(), DbErr> {
```
tyt2y3, line 1694: We should avoid changing existing test cases, since it shrinks our test coverage. Can you extract the common parts and make it a new test case?

ziimakc: I didn't change any existing tests, but duplicated all existing ones with `desc` order, it shows a bit incorrectly by github, I assume because names `composite_keys_5` and `composite_keys_5_desc` are similar

src/executor/cursor.rs
```diff
@@ -1116,14 +1939,15 @@ mod tests {
...
+                r#"ORDER BY "t"."col_1" DESC, "t"."col_2" DESC, "t"."col_3" DESC, "t"."col_4" DESC, "t"."col_5" DESC"#,
             ].join(" "))
         );
 
         assert_eq!(
             DbBackend::Postgres.build(&
                 Entity::find()
                 .cursor_by((Column::Col1, Column::Col2, Column::Col3, Column::Col4, Column::Col5, Column::Col6))
-                .after(("val_1", "val_2", "val_3", "val_4", "val_5", "val_6"))
+                .before(("val_1", "val_2", "val_3", "val_4", "val_5", "val_6")).desc().apply_limit().apply_order_by().apply_filters()
```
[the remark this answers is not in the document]
ziimakc: Same as my previous comment, it's a copy of `cursor_by_many` named as `cursor_by_many_desc`, no tests were removed or changed except additions like `.apply_limit().apply_order_by().apply_filters()` because filters applied when we call `all()`

src/executor/cursor.rs
tyt2y3: Oh thanks!

Issue #1855, opened 2023-09-11
DeriveValueType relies on a QueryResult import


nitnelave:
## Description

When using `#[derive(DeriveValueType)]`, the generated code refers to a naked `QueryType`, which breaks if it's not imported.

## Steps to Reproduce

```rust
use sea_orm::DeriveValueType;

#[derive(DeriveValueType)]
struct MyStruct;
```

### Expected Behavior

Compiles

### Actual Behavior

`cannot find type `QueryResult` in this scope`

### Workarounds

`use sea_orm::QueryResult;`

## Versions

12.0 (DeriveValueType was introduced in 12.0)

tyt2y3:
Thanks for the report! I think we should patch this

0xSh4dy:
Hey @tyt2y3. I'd like to work on this issue

PR #2054, opened 2024-01-05
fix: fully qualify types in DeriveValueType macro

bobogei81123:
`DeriveValueType` proc macro relies on several sea-orm types to be imported. This is partially fixed in #1855 by fully qualifying `sea_orm::QueryResult`, but others like `sea_orm::Value` is still not being qualified. Also, if use defined a type alias on `std::result::Result`, it can cause the proc macro to generate code that won't compile.

- Fully qualify `sea_orm::Value`, `std::string::String`, and `std::result::Result` in `DeriveValueType` proc macro.
- Add some simple integration tests for this.
- Add doc and doc test in `sea_orm_macro::derive_value_type`.

## Bug Fixes

- [x] Fix bugs that `DeriveValueType` proc macro relies on several sea-orm types to be imported.

### DIFF
--- a/sea-orm-macros/src/derives/value_type.rs
+++ b/sea-orm-macros/src/derives/value_type.rs
@@ -100,7 +100,7 @@ impl DeriveValueType {
 
         quote!(
             #[automatically_derived]
-            impl std::convert::From<#name> for Value {
+            impl std::convert::From<#name> for sea_orm::Value {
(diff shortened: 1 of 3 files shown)


tyt2y3 (approved):
Happy new year and thank you. Will merge after test suite passing. I have fixed clippy already, so no worries.

PR #2055, opened 2024-01-08
Add Ability To Configure SQLx Slow Statement Logging Settings

anshap1719:
## New Features

- [x] Add Ability To Configure SQLx Slow Statement Logging Settings

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -54,6 +54,10 @@ pub struct ConnectOptions {
     pub(crate) sqlx_logging: bool,
     /// SQLx statement logging level (ignored if `sqlx_logging` is false)
     pub(crate) sqlx_logging_level: log::LevelFilter,
+    /// SQLx slow statements logging level (ignored if `sqlx_logging` is false)
+    pub(crate) sqlx_slow_statements_logging_level: log::LevelFilter,
+    /// SQLx slow statements duration threshold (ignored if `sqlx_logging` is false)
+    pub(crate) sqlx_slow_statements_logging_threshold: Duration,
     /// set sqlcipher key
     pub(crate) sqlcipher_key: Option<Cow<'static, str>>,
     /// Schema search path (PostgreSQL only)
@@ -147,6 +151,8 @@ impl ConnectOptions {
             max_lifetime: None,
             sqlx_logging: true,
             sqlx_logging_level: log::LevelFilter::Info,
+            sqlx_slow_statements_logging_level: log::LevelFilter::Warn,
+            sqlx_slow_statements_logging_threshold: Duration::from_secs(1),
             sqlcipher_key: None,
             schema_search_path: None,
         }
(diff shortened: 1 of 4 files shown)


anshap1719:
Can you please confirm if I should fix that clippy failure? The error isn't coming from my changes.

tyt2y3:
If you wouldn't mind, that'll be great to fix clippy warnings!

anshap1719:
@tyt2y3 I don't mind at all, the reason I asked is because they look like re-exports and hence could be in use by crate users. So might not be the best idea to do that..

tyt2y3:
If clippy warns about unused re-export, then nothing actually have been exported.

anshap1719:
@tyt2y3 I see. Makes sense. Will fix them then

anshap1719:
Thank you @tyt2y3

tyt2y3:
We fixed them on #2057

src/database/mod.rs
```diff
@@ -147,6 +151,8 @@ impl ConnectOptions {
             max_lifetime: None,
             sqlx_logging: true,
             sqlx_logging_level: log::LevelFilter::Info,
+            sqlx_slow_statements_logging_level: log::LevelFilter::Warn,
```
tyt2y3, line 154: I think we should set the default level to `LevelFilter::Off`

src/driver/sqlx_mysql.rs
```diff
@@ -52,6 +52,10 @@ impl SqlxMySqlConnector {
             opt = opt.disable_statement_logging();
         } else {
             opt = opt.log_statements(options.sqlx_logging_level);
+            opt = opt.log_slow_statements(
```
tyt2y3, line 55: ... and only call this function if level filter is not Off.

src/database/mod.rs
```diff
@@ -269,11 +275,31 @@ impl ConnectOptions {
         self
     }
 
+    /// Set SQLx slow statements logging level and duration threshold (default INFO, 1s)
```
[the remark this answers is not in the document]
tyt2y3: We should not enable it by default

anshap1719:
@tyt2y3 I was wondering, should I put effort into being able to do this when user disables SQLx logging and uses sea-orm's debug-print instead? I will create a separate PR for it if it makes sense.

tyt2y3:
umm I don't think it can easily be done. disabling sqlx logging would disable slow log.

anshap1719:
@tyt2y3 Maybe I didn't word it clearly. I was just asking if "Slow statements logging" when `debug-print` is enabled, makes sense, completely independant of sqlx.

tyt2y3:
I think not necessary. Because I can imagine it'd take a lot of effort.

anshap1719:
@tyt2y3 Got it. Yup. Thanks for confirming.

tyt2y3 (approved):

tyt2y3:
Thank you. I'll merge now. Feel free to suggest other improvements!

anshap1719:
@tyt2y3 Will do. Thank you. Do you know yet when the next release will be, or are there no plans yet?

tyt2y3:
Probably today. Or the next few days.

anshap1719:
@tyt2y3 Perfect! Thank you.

PR #2056, opened 2024-01-10
Refactor/fix clippy errors

shogo-nakano-desu:
## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+allow-unwrap-in-tests = true
--- a/src/query/combine.rs
+++ b/src/query/combine.rs
@@ -2,7 +2,6 @@ use crate::{
     ColumnTrait, EntityTrait, IdenStatic, Iterable, QueryTrait, Select, SelectTwo, SelectTwoMany,
 };
 use core::marker::PhantomData;
-pub use sea_query::JoinType;
 use sea_query::{Alias, ColumnRef, Iden, Order, SeaRc, SelectExpr, SelectStatement, SimpleExpr};
 
 macro_rules! select_def {
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -15,7 +15,7 @@ pub use combine::{SelectA, SelectB};
 pub use delete::*;
 pub use helper::*;
 pub use insert::*;
-pub use join::*;
+
 #[cfg(feature = "with-json")]
 pub use json::*;
 pub use loader::*;
--- a/sea-orm-macros/src/strum/helpers/mod.rs
+++ b/sea-orm-macros/src/strum/helpers/mod.rs
@@ -1,4 +1,3 @@
-pub use self::case_style::CaseStyleHelpers;
 pub use self::type_props::HasTypeProperties;
 pub use self::variant_props::HasStrumVariantProperties;
 
--- a/src/driver/mock.rs
+++ b/src/driver/mock.rs
@@ -100,6 +100,8 @@ impl MockDatabaseConnector {
 }
 
 impl MockDatabaseConnection {
+    /// # Panics
+    ///
     /// Create a connection to the [MockDatabase]
     pub fn new<M: 'static>(m: M) -> Self
     where
@@ -116,6 +118,8 @@ impl MockDatabaseConnection {
         &self.mocker
     }
 
+    /// # Panics
+    ///
     /// Get the [DatabaseBackend](crate::DatabaseBackend) being used by the [MockDatabase]
     pub fn get_database_backend(&self) -> DbBackend {
         self.mocker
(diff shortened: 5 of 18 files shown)


clippy.toml
```diff
@@ -0,0 +1 @@
+allow-unwrap-in-tests = true
```
shogo-nakano-desu, line 1: I have added this file to configure clippy.

src/lib.rs
```diff
@@ -345,7 +345,6 @@ pub mod tests_cfg;
 mod util;
 
 pub use database::*;
-pub use driver::*;
```
tyt2y3, line 348: Instead of removing this line, I think we should add one `#[allow(clippy::)]`, because the content of the re-export depends on feature flags

src/query/combine.rs
```diff
@@ -2,7 +2,6 @@ use crate::{
     ColumnTrait, EntityTrait, IdenStatic, Iterable, QueryTrait, Select, SelectTwo, SelectTwoMany,
 };
 use core::marker::PhantomData;
-pub use sea_query::JoinType;
```
tyt2y3, line 5: can confirm this is already exported in `query.rs`

src/query/mod.rs
```diff
@@ -15,7 +15,6 @@ pub use combine::{SelectA, SelectB};
 pub use delete::*;
 pub use helper::*;
 pub use insert::*;
-pub use join::*;
```
tyt2y3, line 18: The file contains only impl, so nothing to export

src/driver/sqlx_mysql.rs
```diff
@@ -36,6 +36,7 @@ impl std::fmt::Debug for SqlxMySqlPoolConnection {
 
 impl SqlxMySqlConnector {
     /// Check if the URI provided corresponds to `mysql://` for a MySQL database
+    #[cfg(feature = "mock")]
```
tyt2y3, line 39: this is what confuses me, this `accepts` method should be in the public API, why does clippy think it's redundant?

clippy.toml
tyt2y3: May be a `#![allow(clippy::unwrap_used, clippy::expect_used)]` directive in test files would be better?

src/query/mod.rs
shogo-nakano-desu: In that case, do we need to keep `pub use join::*;`?

clippy.toml
shogo-nakano-desu: We can do that. It might be better because clippy config file like `clippy.toml` is unstable.

src/driver/sqlx_mysql.rs
shogo-nakano-desu: I think that is because the `accepts` method is not used from anywhere in this workspace except for `mock`. But if we want this method to be public, we need `#[allow(unused_variables)]` instead so I will replace them.

clippy.toml
shogo-nakano-desu: Fixed
[889d250](https://github.com/SeaQL/sea-orm/pull/2056/commits/889d250d54663f0b35fcdb3c1981ba1b820a252c)

src/lib.rs
shogo-nakano-desu: Fixed 
[138b9dc](https://github.com/SeaQL/sea-orm/pull/2056/commits/138b9dc75a006e5fcf2809455ff3aea534662d03)

I will add `#[allow(clippy::unused_xxx)]` later because I could not see clippy error in my local somehow...

src/query/combine.rs
shogo-nakano-desu: Oh, it is not exported so I have reverted this change. [e4f4a12](https://github.com/SeaQL/sea-orm/pull/2056/commits/e4f4a129b9a1220ddc41fc5de2dff9866aaf9388)

src/driver/sqlx_mysql.rs
shogo-nakano-desu: Remove feature flag and I put `#[allow(unused_variables)]` instead
[09f845a](https://github.com/SeaQL/sea-orm/pull/2056/commits/09f845ad52fc6637ec25f66633ed6f04b6bb2153)

src/lib.rs
shogo-nakano-desu: I have added `#[allow(unused_imports)]`
[20dc881](https://github.com/SeaQL/sea-orm/pull/2056/commits/20dc881d81bc7951c6efbfe4b8ba863f41b3f8a0)

tyt2y3 (commented):
Thank you. I will make some tweaks

shogo-nakano-desu:
Thank you so much for your review. I will check your PR and would like to learn from it:)

PR #2067, opened 2024-01-17
Expose PostgreSQL Application Name Option Config

anshap1719:
## Changes

- [x] Expose PostgreSQL Application Name Option Config
- [ ] Fix tests


tyt2y3:
Thank you. I think it's fine to add new config. However this PR shouldn't have to change the `Cargo.toml` files (and well now there's a number of conflicts).

anshap1719:
@tyt2y3 It wasn't intentional. I believe it only happened because PR is too old. I will fix.

PR #2070, opened 2024-01-18
Enable sqlite returning with feature flag

### LINKED ISSUE #1885: Support RETURNING for SQLite
@billy1624 I'm not sure what the original problem was, but could RETURNING be supported for SQLite now?

## Motivation

I have a query that has an `ON CONFLICT` and I want to use `exec_with_returning`, however `last_insert_rowid` is not set for SQLite when there's a conflict, so there's no way to figure out which row was updated. My query looks like:

```rust
Model::insert().on_conflict(...).exec_with_returning(db)
```

This works fine on Postgres but not SQLite

## Proposed Solutions

Get RETURNING working for SQLite

## Current Workarounds

You can do an extra SELECT query instead of using ON CONFLICT

acrrd:
Following the discussion in #654 and this comment https://github.com/SeaQL/sea-orm/issues/1885#issuecomment-1741173714 this PR adds a feature flag to mark SQLite as a returning backend.

For my use case, this means to be able to use other types as primary keys with SQLite.

## PR Info

- Closes #1885 #653

## New Features

This PR adds a new feature: `returning_clauses_for_sqlite_3_35`.
This enables the SQLite backend to be a returning backend.
With this feature, the user must ensure that the application links to SQLite version >= 3.35.


tyt2y3:
Thank you for the contribution. I think at runtime, upon Pool creation, we can do a `select sqlite_version()` and check for the version number.

We can enable this flag by default at some point in the future.

acrrd:
I added the check. Let me know if you like the approach and if there is a better way to return the errors.

.github/workflows/rust.yml
```diff
@@ -308,6 +308,7 @@ jobs:
       matrix:
         runtime: [async-std]
         tls: [native-tls, rustls]
+        sqlite_flavor: ["", returning_clauses_for_sqlite_3_35]
```
tyt2y3, line 311: You yml skills are very good!

src/driver/sqlx_sqlite.rs
```diff
@@ -68,12 +69,20 @@ impl SqlxSqliteConnector {
...
                     pool,
                     metric_callback: None,
-                },
-            )),
+                };
+
+                #[cfg(feature = "returning_clauses_for_sqlite_3_35")]
+                {
+                    let version = get_version(&pool).await?;
+                    enure_returning_version(&version)?;
```
tyt2y3, line 81:
```suggestion
                    ensure_returning_version(&version)?;
```

src/driver/sqlx_sqlite.rs
```diff
@@ -268,3 +277,78 @@ pub(crate) async fn set_transaction_config(
...
+        .await?
+        .ok_or_else(|| {
+            DbErr::Conn(RuntimeErr::Internal(
+                "Error reading SQLite version".to_string(),
+            ))
+        })?
+        .try_get_by(0)
+}
+
+fn enure_returning_version(version: &str) -> Result<(), DbErr> {
```
tyt2y3, line 297:
```suggestion
fn ensure_returning_version(version: &str) -> Result<(), DbErr> {
```

src/driver/sqlx_sqlite.rs
```diff
@@ -268,3 +277,78 @@ pub(crate) async fn set_transaction_config(
...
+    };
+
+    let major = extract_next()?;
+    let minor = extract_next()?;
+
+    if major > 3 || (major == 3 && minor >= 35) {
+        Ok(())
+    } else {
+        Err(DbErr::Conn(RuntimeErr::Internal(
+            "SQLite library does not support returning".to_string(),
```
tyt2y3, line 321:
```suggestion
            "SQLite version does not support returning".to_string(),
```

tyt2y3 (approved):
A few clippy warnings but otherwise great.
Actually I'd rename the feature to `sqlite-use-returning-for-3_35`

tyt2y3:
Thank you

PR #2081, opened 2024-01-26
Generate select_as and save_as attributes for custom column types.

### LINKED ISSUE #1643: `sea-orm-cli generate entity` migration `select_as`, `save_as` support
## Motivation
I was trying to make Postgresql `citext` column with migration and generate entity with `sea-orm-cli generate entity`.

Migration up:
```rust
manager
    .get_connection()
    .execute_unprepared("CREATE EXTENSION IF NOT EXISTS citext")
    .await?;

manager
    .create_table(
        Table::create()
            .table(User::Table)
            .if_not_exists()
            .col(ColumnDef::new(User::Id).uuid().not_null().primary_key())
             .col(
                 ColumnDef::new(User::Email)
                    .custom(Alias::new("citext"))
                    .not_null()
                    .unique_key(),
              )
              .to_owned(),
          )
          .await?;
```

`sea-orm-cli generate entity -o src/entity` generates:
```rust
//! `SeaORM` Entity. Generated by sea-orm-codegen 0.11.3

#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
#[sea_orm(table_name = "user")]
pub struct Model {
    pub id: Uuid,
    #[sea_orm(column_type = "custom(\"citext\")", unique)]
    pub email: String,
}
```

When I try to select the user, this error occurs:
```
DatabaseError occurred: Query Error: error occurred while decoding column "email": mismatched types; Rust type `core::option::Option<alloc::string::String>` (as SQL type `TEXT`) is not compatible with SQL type `citext`
```

If I edit the entity to this the error doesn't occurs:
```rust
#[sea_orm(column_type = "custom(\"citext\")", select_as = "text", save_as = "citext", unique)]
pub email: String
```

## Proposed Solutions
Able to set `select_as`, `save_as` in migration.

## Current Workarounds
Edit the entity manually when I generate it.

lucasmerlin:
## PR Info

- Fixes #1643

## New Features

- Generate select_as and save_as attributes to automatically cast custom types
Before, using custom types in postgres resulted in a error like this: 
```
DatabaseError occurred: Query Error: error occurred while decoding column "email": mismatched types; Rust type `core::option::Option<alloc::string::String>` (as SQL type `TEXT`) is not compatible with SQL type `citext`
```
Now it will be casted correctly.

I am not sure what the implications of adding select_as and save_as are for other database backends, since I believe this is only needed for postgres? So maybe it will need to be made configurable or only be enabled for postgres. Some feedback on this would be appreciated.

I've tested these changes in my codebase where it works great.

## Breaking Changes

- Custom types will now have the select_as and save_as attributes generated. But since they wouldn't have been usable before, at least with postgres, I'm not sure if this is a breaking change.

### DIFF
--- a/sea-orm-codegen/src/entity/writer.rs
+++ b/sea-orm-codegen/src/entity/writer.rs
@@ -3,6 +3,7 @@ use heck::ToUpperCamelCase;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
 use std::{collections::BTreeMap, str::FromStr};
+use sea_query::ColumnType;
 use syn::{punctuated::Punctuated, token::Comma};
 use tracing::info;
 
@@ -754,6 +755,11 @@ impl EntityWriter {
                         attrs.push(quote! { nullable });
                     }
                 };
+                if let ColumnType::Custom(typ) = &col.col_type {
+                    let type_string = typ.to_string();
+                    attrs.push(quote! { select_as = "text" });
+                    attrs.push(quote! { save_as = #type_string });
+                }
                 if col.unique {
                     attrs.push(quote! { unique });
                 }


marcusbuffett:
I keep shooting myself in the foot with this issue - sea-orm-cli will overwrite the change to my users.rs file, which includes `save_as = "citext"`, then I accidentally deploy the site without that `save_as`, and user lookups take 5 seconds instead of .01 seconds, since it's doing a full scan when querying citext columns as TEXT. Was also quite difficult to figure out this is what was going on, at first.

TL;DR: would love for this to be merged

PR #2109, opened 2024-02-13
Add mock feature flag docs

AngelOnFira:
This adds some better docs for why `DatabaseConnection` might not be clone. From a discussion on Discord: https://discord.com/channels/873880840487206962/1205866461935964182


AngelOnFira:
I'm not sure if this is the only doc fix that is needed, since many examples seem to include `mock` by default, meaning that people who copy them as templates will likely run into this problem.

tyt2y3 (approved):
Thank you

PR #2111, opened 2024-02-15
DerivePartialModel: Allow specification of Entity via a Path in addition to as an Ident 

jreppnow:
## PR Info

- Content: With the current implementation when deriving `DerivePartialModel`, one must specify the `Entity` whose columns should be used as an `Ident` - meaning that one must import it into the current scope
- This is somewhat unwieldy, especially when dealing with generated code, where one would prefer to just `use` the containing module, not the contents themselves
- This PR instead allows one to specify a `Path`, meaning that no direct `use` is required anymore

- Ticket: Sorry, should have probably created a ticket first - does not exist atm, but can create one if necessary. 

## Bug Fixes

- [x] allow the `Entity` to be specified as a `Path`, not just an `Ident`, alleviating the need to import the Entity struct whenever you want to use the DerivePartialModel macro

## Breaking Changes

- The changes should be backwards-compatible/a pure extension, since `Path`s are a superset of `Ident`s

## Changes

- [x] fixed a typo in the changed file ("specific" to "specified") in a couple of places

### DIFF
--- a/sea-orm-macros/src/derives/partial_model.rs
+++ b/sea-orm-macros/src/derives/partial_model.rs
@@ -16,7 +16,7 @@ use self::util::GetAsKVMeta;
 #[derive(Debug)]
 enum Error {
     InputNotStruct,
-    EntityNotSpecific,
+    EntityNotSpecified,
     NotSupportGeneric(Span),
     BothFromColAndFromExpr(Span),
     Syn(syn::Error),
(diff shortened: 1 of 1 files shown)


tyt2y3:
I think the intent of this PR is mostly already covered by https://github.com/SeaQL/sea-orm/pull/2137 
If you have still want to add something, please rebase onto master, thanks!

PR #2136, opened 2024-03-03
Bring back rocket_okapi support

### LINKED ISSUE #2135: Missing okapi support
## Description

Raising this as a bug, both because there is a broken example on the repo, and because the breakage in question hasn't been communicated as a major version bump/info in changelog.

I can see that the support to rocket_okapi has been added in #1071, and then removed in c3b725662291b17142a7b5ab4b8ab022f1d0d78e. This has been included as part of 0.12.5 release ([0.12.4...0.12.5](https://github.com/SeaQL/sea-orm/compare/0.12.4...0.12.5)), but its changelog doesn't mention this (breaking) change.

It would be nice to get the rocket_okapi support back or get a reason for removing the support and in this case, removal of the example and feature flag from the repository. Now it's misleading, as the example suggests that the support is there, while it doesn't work.

## Steps to Reproduce

1. Clone the sea-orm repo
2. `cd examples/rocket_okapi_example`
3. `cargo run`

### Expected Behavior

The example compiles and runs.

### Actual Behavior

It fails with `failed to resolve: use of undeclared crate or module rocket_okapi` (among other errors).

### Reproduces How Often

Always

### Workarounds

I looked at the git history and found out the support has been silently removed in c3b725662291b17142a7b5ab4b8ab022f1d0d78e.

## Versions

* sea-orm v1.0.0-rc.1
* sea-orm-rocket v0.5.4
* DB: SQLite (although it doesn't really matter, as the issue happens at the compile time)
* OS: Arch Linux

m4tx:
## PR Info

This brings back `rocket_okapi` support - see the issue linked below for details.

In addition to that, it modifies the rocket_okapi_example slightly by changing to SQLite by default (to make running the example a bit easier) and updates the `rocket_cors` dependency.

- Closes #2135

- Dependencies:
  - none

- Dependents:
  - none

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [x] Bring back `rocket_okapi` support, originally added in #1071, then removed in c3b725662291b17142a7b5ab4b8ab022f1d0d78e because of missing support for Rocket 0.5 in `rocket_okapi` at the time

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -243,7 +243,7 @@ jobs:
           examples/poem_example,
           examples/proxy_gluesql_example,
           examples/rocket_example,
-          # examples/rocket_okapi_example,
+          examples/rocket_okapi_example,
           examples/salvo_example,
           examples/seaography_example,
           examples/tonic_example,
--- a/examples/rocket_okapi_example/README.md
+++ b/examples/rocket_okapi_example/README.md
@@ -1,8 +1,8 @@
 # Rocket and Rocket-API with SeaORM example app
 
-1. Modify the `url` var in `api/Rocket.toml` to point to your chosen database
+1. Modify the `url` var in `api/Rocket.toml` to point to your chosen database (or leave it as-is to use in-memory SQLite)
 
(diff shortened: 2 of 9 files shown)


tyt2y3 (commented):
Thank you for the contribution!

Issue #2113, opened 2024-02-18
[BUG] `DerivePartialModel` attr `entity` not support `syn::Path` token


Goodjooy:
<!--

Welcome! Thank you for reporting bugs!

First of all, please star our repo. Your support is vital to the continued maintenance of SeaORM.

Want to ask a question? You can reach us via:

- Discord: https://discord.com/invite/uCPdDXzbdv
- GitHub Discussions: https://github.com/SeaQL/sea-orm/discussions/new

Please make sure that you are not asking for a missing feature; a bug is incorrect behavior -
either in the feature specification or implementation. Feature requests should be first raised in discussions.

Please also make sure your description is clear and precise - maintainers don't have access to your
code and can't see what you have seen. Please avoid vague descriptions like "they are different"
or "the program crashes" - in either case, provide exact information.

If you are certain there is a bug, please provide a reproducible example, which helps the investigator
to pin-point the bug and the implementor to verify that a solution is satisfactory. Bug reports without 
reproducible example may be closed immediately or dangle forever.

Finally, please search for existing issues and discussions before submission. Feel free to revive old
threads if you have new information to add, but please don't ask for ETA or "+1".

-->

## Description

`DerivePartialModel` will produce error when set `entity` as `path::to::entity::Entity`


## Steps to Reproduce

1. prepare an Entity, using `cake` for example
2. define the partial model like following
```rust
#[derive(FromQueryResult, DerivePartialModel)]
#[sea_orm(entity="cake::Entity")]
struct CakePartial{
   .. // fields
}
```
3. get the macro error : 
```
error: unexpected token                                                                                                                                                                                                            
  --> src\service\child_statical.rs:48:45
   |
48 | #[derive(FromQueryResult, DerivePartialModel)]
   |                           ^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the derive macro `DerivePartialModel` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Expected Behavior

the macro should accept the entity as a `Path` token

### Actual Behavior

The macro only accept `Ident` token

### Workarounds

1. using the `Cake` in the `src/entities/prelude` ( the `src/entities` is the out put dir of the command `sea-orm-cli generate entity`)
2. using a local mod use the entity, like following
```rust
mod local_quiz {
            use crate::entities::quizes::Entity;
            use sea_orm::DerivePartialModel;
            use sea_orm::FromQueryResult;
[...]

PR #2137, opened 2024-03-03
Fix `DerivePartialModel` attr `entity` not support syn::Type token

Goodjooy:
## PR Info

- #2113 

## Bug Fixes

- [x] fix `DerivePartialModel` attr argument `entity` not accept `syn::Type`


tyt2y3 (approved):
Thank you!

PR #2140, opened 2024-03-03
actix example bug fix

### LINKED ISSUE #2122: Incorrect Route Matching Behavior in Actix Example Code
## Description

I discovered an issue with the route matching behavior in the example code located in the `examples/actix_example` directory. When accessing routes that do not match the specified patterns like `/xxx`, instead of returning a 404 Not Found page, the application returns an error message `can not parse 'xxx' to a i32`.

## Steps to Reproduce

1. Navigate to the `examples/actix_example` directory.
2. Run `cargo run` to start the application.
3. Access a route that does not match any specified patterns, for example, `/xxx`.
4. Observe the application's response and note that it does not behave as expected.

### Expected Behavior

I expect the application to handle routes that do not match any specified patterns appropriately, such as returning a 404 page.

### Actual Behavior

Instead of returning a 404 page, the application returns an error message `can not parse 'xxx' to a i32`.

### Reproduces How Often

This issue is reproducible every time.

### Workarounds

The issue seems to be caused by `/xxx` matching the `#[get("/{id}")]` route in `examples/actix_example/api/src/lib.rs`. This leads to the execution of the `async fn edit(data: web::Data<AppState>, id: web::Path<i32>) -> Result<HttpResponse, Error>` handler function.

ikka0426:
## PR Info

- Closes #2122

## Bug Fixes

- [x] Fixed a routing bug where non-numeric one-level paths were not handled correctly.
- [x] Fixed an issue where data was not returned when the ID represented by a number did not exist in the database.

## Changes

- [ ] Modified the routing logic in `examples/actix_example/api/src/lib.rs` to return a 404 Not Found page when non-numeric paths are entered.

### DIFF
--- a/examples/actix_example/api/src/lib.rs
+++ b/examples/actix_example/api/src/lib.rs
@@ -89,24 +89,34 @@ async fn create(
         .finish())
 }
 
-#[get("/{id}")]
+#[get(r#"/{id:\d+}"#)]
 async fn edit(data: web::Data<AppState>, id: web::Path<i32>) -> Result<HttpResponse, Error> {
     let conn = &data.conn;
     let template = &data.templates;
     let id = id.into_inner();
 
-    let post: post::Model = Query::find_post_by_id(conn, id)
+    let post: Option<post::Model> = Query::find_post_by_id(conn, id)
         .await
-        .expect("could not find post")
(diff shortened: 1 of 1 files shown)


tyt2y3 (approved):
Thank you

PR #2148, opened 2024-03-10
Add method for retrieiving column names

jharrilim:
tl;dr: Add a method for retrieving column names from query results so that applications can use them alongside retrieved rows 

## Background

For creating some database tooling, I'd like to be able to retrieve database column names that come with the query results. This would be very dynamic because the queries are either generated at runtime from the user, or from introspection. Ideally I'd be able to get to a point where I have something like this:

```rust
let results: Vec<QueryResult> = db // DatabaseConnection
    .query_all(Statement::from_string(db.db_type.into(), query))
    .await?;
let headers = results
    .first()
    .and_then(|r| Some(r.columns())) // This PR addresses this part
    .unwrap_or_default();

// This doesn't work yet. Ideally there's a way to convert all row values
// to some enum for which I can match against and implement std::fmt::Display for
let rows = results
    .iter()
    .flat_map(|r| r.try_get_many_by_index::<serde_json::Value>())
    .map(|r| r.to_string())
    .collect::<Vec<_>>();
```

so that I can display headers in a table and the rows underneath them. Not so different from the output that you'd see in `psql`.

## Aside

- I've considered using sqlx directly, but I can't map the inner types to rust types easily https://github.com/launchbadge/sqlx/issues/1369, and that seems to be covered in sea-orm.

- My current implementation is really basic. I'm not sure if it would be better to return an `impl Iterator` for the column names, or return actual column structs with their metadata instead.

- I wasn't too sure how to write and execute tests for this. I had initially wrote one in the `query.rs` file like:
    ```rust
    #[cfg(feature = "sqlx-dep")]
    #[test]
    fn column_names_from_query_result() {
        let mut values = BTreeMap::new();
        values.insert("id", Value::Int(Some(1)));
        values.insert("name", Value::String(Some("Abc".to_owned())));
        let query_result = QueryResult {
            row: QueryResultRow::Mock(crate::MockRow { values }),
        };
        assert_eq!(
            query_result.columns(),
            vec!["id".to_owned(), "name".to_owned()]
        );
    }
    ```
  And tried to run it with `cargo t --test query_tests --features sqlx-dep`, but I get a bunch of compile errors saying `use of undeclared crate or module `sqlx`. Maybe there's a better way?

---

## New Features

- [x] Adds a method for retrieving column names from a query result

## Breaking Changes

- [x] None

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -116,22 +116,36 @@ impl QueryResult {
         Ok(T::try_get_many_by_index(self)?)
     }
 
-    /// Gets all column names in the order they were written.
-    #[cfg(feature = "sqlx-dep")]
-    pub fn columns(&self) -> Vec<String> {
+    /// Retrieves the names of the columns in the result set
+    pub fn column_names(&self) -> Vec<String> {
+        #[cfg(feature = "sqlx-dep")]
         use sqlx::Column;
 
         match &self.row {
             #[cfg(feature = "sqlx-mysql")]
-            QueryResultRow::SqlxMySql(row) => row.columns().iter().map(|c| c.name().to_string()).collect(),
+            QueryResultRow::SqlxMySql(row) => {
+                row.columns().iter().map(|c| c.name().to_string()).collect()
+            }
             #[cfg(feature = "sqlx-postgres")]
-            QueryResultRow::SqlxPostgres(row) => row.columns().iter().map(|c| c.name().to_string()).collect(),
+            QueryResultRow::SqlxPostgres(row) => {
+                row.columns().iter().map(|c| c.name().to_string()).collect()
+            }
             #[cfg(feature = "sqlx-sqlite")]
-            QueryResultRow::SqlxSqlite(row) => row.columns().iter().map(|c| c.name().to_string()).collect(),
+            QueryResultRow::SqlxSqlite(row) => {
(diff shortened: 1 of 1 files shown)


src/executor/query.rs
```diff
@@ -115,6 +115,27 @@ impl QueryResult {
     {
         Ok(T::try_get_many_by_index(self)?)
     }
+
+    /// Gets all column names in the order they were written.
+    #[cfg(feature = "sqlx-dep")]
```
tyt2y3, line 120: I think if we move this line to directly above `use sqlx::Column`, this can still compile?

tyt2y3:
> And tried to run it with cargo t --test query_tests --features sqlx-dep, but I get a bunch of compile errors saying use of undeclared crate or module sqlx

sqlx-dep is a 'transitive  dependency'. You have to specify a specific backend, e.g. `sqlx-postgres`

tyt2y3:
> I'm not sure if it would be better to return an impl Iterator for the column names, or return actual column structs with their metadata instead.

Whatever suits your need is good enough! Although I'd rename the method to `column_names` so that we can implement the advanced version in the future.

jharrilim:
Alright, I've made the requested changes and added a test!

tyt2y3 (approved):
Nice work, thanks!

PR #2164, opened 2024-03-19
feat: allow an easier use of limit/offset

ctron:
This is a draft PR and it lacks some things to fix up before merging. I just want to understand if that's going the right direction.

## PR Info

- Closes https://github.com/SeaQL/sea-orm/discussions/2163

## New Features

- [x] Add an easier way working with limit & offset


billy1624 (commented):
Hey @ctron, thanks for contributing!!

Sorry, I don't quite understand the exact usage of the Limiter after reading https://github.com/SeaQL/sea-orm/discussions/2163 and this PR. Could you please elaborate?

ctron:
Right now, one can use `page_size` and `start_page`. However, there seems to be no simple way to use `limit` and `offset`.

The idea is to provide a simple way to set `limit` and `offset`, which is a more basic version. That could then (not part of this PR yet) be used to implement pagination on top of it.

PR #2168, opened 2024-03-21
Upstream Changes - 4

billy1624:


sea-orm-cli/src/commands/generate.rs
```diff
@@ -171,7 +171,7 @@ pub async fn run_generate_command(
                         .filter(|schema| filter_skip_tables(&schema.info.name))
                         .map(|schema| schema.write())
                         .collect();
-                    (Some(schema.schema), table_stmts)
+                    (database_schema, table_stmts)
```
billy1624, line 174: we should only pass through non-default schema here

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -818,16 +818,9 @@ impl EntityWriter {
...
-        match schema_name {
-            Some(schema_name) => {
-                if schema_name != "public" {
-                    Some(quote! { #schema_name })
-                } else {
-                    None
-                }
-            }
-            None => None,
-        }
```
billy1624, line 830: Then we can remove this checking at codegen. This checking should be part of CLI anyway.

sea-orm-cli/src/commands/generate.rs
tyt2y3: OK

tyt2y3 (approved):

PR #2171, opened 2024-03-24
Set runtime features for sqlx.

bamidev:
When you add sea-orm simply like this:
`sea-orm = { version = "0.12.15", features = ["runtime-tokio", "sqlx-sqlite"] }`
It will complain about no runtime being selected:
```
thread 'main' panicked at /home/bamilab/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sqlx-core-0.7.4/src/pool/inner.rs:52:24:
either the `runtime-async-std` or `runtime-tokio` feature must be enabled
```
Even though I use the "runtime-tokio" feature.
It seems like it isn't being passed down to the `sqlx` crate.

For the moment I use the workaround of adding this to my Cargo.toml:
`sqlx = { features = ["runtime-tokio"] }`
But it is just a workaround.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,7 @@ sqlx-mysql = ["sqlx-dep", "sea-query-binder/sqlx-mysql", "sqlx/mysql"]
 sqlx-postgres = ["sqlx-dep", "sea-query-binder/sqlx-postgres", "sqlx/postgres", "postgres-array"]
 sqlx-sqlite = ["sqlx-dep", "sea-query-binder/sqlx-sqlite", "sqlx/sqlite"]
 sqlite-use-returning-for-3_35 = []
-runtime-async-std = ["sqlx/runtime-async-std"]
+runtime-async-std = ["sqlx?/runtime-async-std"]
 runtime-async-std-native-tls = [
     "sqlx?/runtime-async-std-native-tls",
     "sea-query-binder?/runtime-async-std-native-tls",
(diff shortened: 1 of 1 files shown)


tyt2y3:
You're right. The original design is that you have to specify a runtime+tls feature flag , e.g. `runtime-tokio-native-tls`.
The `runtime-tokio` is transitive, and is simply used to signify that tokio has been enabled in sea-orm.

Cargo.toml
```diff
@@ -113,7 +113,7 @@ runtime-actix-rustls = [
     "sea-query-binder?/runtime-actix-rustls",
     "runtime-actix",
 ]
-runtime-tokio = []
+runtime-tokio = ["sqlx/runtime-tokio"]
```
tyt2y3, line 116:
```suggestion
runtime-tokio = ["sqlx?/runtime-tokio"]
```

Cargo.toml
```diff
@@ -91,7 +91,7 @@ sqlx-mysql = ["sqlx-dep", "sea-query-binder/sqlx-mysql", "sqlx/mysql"]
 sqlx-postgres = ["sqlx-dep", "sea-query-binder/sqlx-postgres", "sqlx/postgres", "postgres-array"]
 sqlx-sqlite = ["sqlx-dep", "sea-query-binder/sqlx-sqlite", "sqlx/sqlite"]
 sqlite-use-returning-for-3_35 = []
-runtime-async-std = []
+runtime-async-std = ["sqlx/runtime-async-std"]
```
tyt2y3, line 94:
```suggestion
runtime-async-std = ["sqlx?/runtime-async-std"]
```

PR #2177, opened 2024-03-28
add #![cfg(any(feature ...))], to reduce warning in cargo test

Hmikihiro:
## Changes

- [x] I add `#![cfg()]`to the beginning of files that have `#[sea_orm_macros::test]` macros.  

 When I run cargo test, Cl makes many warning such as unused_import. Those imports were compiled all time, but test function is compiled when future have "sqlx- ...".
`#[sea_orm_macros::test]` attributes is expand to `#[cfg(any(
    feature = "sqlx-mysql",
    feature = "sqlx-sqlite",
    feature = "sqlx-postgres"
))]`. So, I think that to reduce `cargo test` warnings, conditional compilation with the same condition that `#[sea_orm_macros::test]` is needed for each file.

### DIFF
--- a/tests/connection_tests.rs
+++ b/tests/connection_tests.rs
@@ -1,3 +1,8 @@
+#![cfg(any(
+    feature = "sqlx-mysql",
+    feature = "sqlx-sqlite",
+    feature = "sqlx-postgres"
+))]
 pub mod common;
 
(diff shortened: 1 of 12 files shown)


tyt2y3 (approved):
Thanks!

PR #1716, opened 2023-06-21
Implement "Nest" for `DerivePartialModel` and From `QueryResult`

Goodjooy:
## PR Info

-  #1715

## New Features

- [x] `FromQueryResult` support `flatten`
- [x] `DerivePartialModel` support `flatten`
- [x] document update
- [x] test case

## Changes

- [x] `FromQueryResult` has attributes `sea_orm`

PR #2167, opened 2024-03-21
Feat/reppnj/partial model skip

jreppnow:
## New Features

- [ ] allow for usage of `skip` when deriving `PartialModelTrait`, analogous to `FromQueryResult`
- [ ] adjust the related error messages accordingly

## Changes

- [ ] fixed some typos in the related error messages/types

PR #2179, opened 2024-03-29
feat: Allow `nest`ing of structs deriving `FromQueryResult` (and `DerivePartialModel`)

jreppnow:
## PR Info

Hi, this is a separate implementation of #1716 , which seems to have somewhat stalled. 
Normally I would not cut in from the side like this, but this feature would allow us to cut down on code duplication **massively** and avoid some quite annoying bugs from re-occurring in our code bases, so we would appreciate if could be merged in the near future. 

## New Features

- [x] allows for usage of the `nested` attribute in both `FromQueryResult` and `DerivePartialModel`
- [x] I took special care to preserve the behavior otherwise, to allow for this being integrated as a minor fix 

## Breaking Changes

- [x] hopefully none (although I would really like to change the `FromQueryResult` trait, but that could cause breakage in dependent crates..)

## Changes

- [x] fixed some typos in the compile time messages for the derive macros

### DIFF
--- a/sea-orm-macros/src/derives/from_query_result.rs
+++ b/sea-orm-macros/src/derives/from_query_result.rs
@@ -3,49 +3,53 @@ use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::{
     ext::IdentExt, punctuated::Punctuated, token::Comma, Data, DataStruct, Fields, Generics, Meta,
+    Type,
 };
 
-struct FromQueryResultItem {
-    pub skip: bool,
-    pub ident: Ident,
+enum ItemType {
+    Normal,
+    Skipped,
+    Nested,
 }
 
-impl ToTokens for FromQueryResultItem {
-    fn to_tokens(&self, tokens: &mut TokenStream) {
-        let Self { ident, skip } = self;
-        if *skip {
-            tokens.extend(quote! {
-                #ident: std::default::Default::default(),
-            });
-        } else {
-            let name = ident.unraw().to_string();
-            tokens.extend(quote! {
-                #ident: row.try_get(pre, #name)?,
-            });
-        }
-    }
+struct FromQueryResultItem {
+    pub typ: ItemType,
+    pub ident: Ident,
 }
 
 struct TryFromQueryResultCheck<'a>(&'a FromQueryResultItem);
 
 impl<'a> ToTokens for TryFromQueryResultCheck<'a> {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        let FromQueryResultItem { ident, skip } = self.0;
-        if *skip {
-            tokens.extend(quote! {
-                let #ident = std::default::Default::default();
-            });
-        } else {
-            let name = ident.unraw().to_string();
-            tokens.extend(quote! {
-                let #ident = match row.try_get_nullable(pre, #name) {
-                    std::result::Result::Err(sea_orm::TryGetError::DbErr(err)) => {
-                        return Err(err);
-                    }
-                    std::result::Result::Err(sea_orm::TryGetError::Null(_)) =>  std::option::Option::None,
-                    std::result::Result::Ok(v) => std::option::Option::Some(v),
-                };
-            });
+        let FromQueryResultItem { ident, typ } = self.0;
+
+        match typ {
+            ItemType::Normal => {
+                let name = ident.unraw().to_string();
+                tokens.extend(quote! {
+                    let #ident = match row.try_get_nullable(pre, #name) {
+                         Err(v @ sea_orm::TryGetError::DbErr(_)) => {
+                             return Err(v);
+                         }
+                         v => v,
+                     };
+                });
+            }
+            ItemType::Skipped => {
+                tokens.extend(quote! {
+                    let #ident = std::default::Default::default();
+                });
+            }
+            ItemType::Nested => {
+                tokens.extend(quote! {
+                    let #ident = match sea_orm::FromQueryResult::from_query_result_nullable(row, pre) {
+                        Err(v @ sea_orm::TryGetError::DbErr(_)) => {
+                            return Err(v);
+                        }
+                        v => v,
+                    };
+                });
+            }
         }
     }
 }
@@ -54,20 +58,19 @@ struct TryFromQueryResultAssignment<'a>(&'a FromQueryResultItem);
 
 impl<'a> ToTokens for TryFromQueryResultAssignment<'a> {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        let FromQueryResultItem { ident, skip } = self.0;
-        if *skip {
-            tokens.extend(quote! {
-                #ident,
-            });
-        } else {
-            tokens.extend(quote! {
-                #ident: match #ident {
-                    std::option::Option::Some(v) => v,
-                    std::option::Option::None => {
-                        return std::result::Result::Ok(std::option::Option::None);
-                    }
-                },
-            });
+        let FromQueryResultItem { ident, typ, .. } = self.0;
+
+        match typ {
+            ItemType::Normal | ItemType::Nested => {
+                tokens.extend(quote! {
+                    #ident: #ident?,
+                });
+            }
+            ItemType::Skipped => {
+                tokens.extend(quote! {
+                    #ident,
+                });
+            }
         }
     }
 }
@@ -92,19 +95,23 @@ pub fn expand_derive_from_query_result(
 
     let mut fields = Vec::with_capacity(parsed_fields.len());
     for parsed_field in parsed_fields.into_iter() {
-        let mut skip = false;
+        let mut typ = ItemType::Normal;
         for attr in parsed_field.attrs.iter() {
             if !attr.path().is_ident("sea_orm") {
                 continue;
             }
             if let Ok(list) = attr.parse_args_with(Punctuated::<Meta, Comma>::parse_terminated) {
                 for meta in list.iter() {
-                    skip = meta.exists("skip");
+                    if meta.exists("skip") {
+                        typ = ItemType::Skipped;
+                    } else if meta.exists("nested") {
+                        typ = ItemType::Nested;
+                    }
                 }
             }
         }
         let ident = format_ident!("{}", parsed_field.ident.unwrap().to_string());
-        fields.push(FromQueryResultItem { skip, ident });
+        fields.push(FromQueryResultItem { typ, ident });
     }
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
@@ -114,22 +121,21 @@ pub fn expand_derive_from_query_result(
     Ok(quote!(
         #[automatically_derived]
         impl #impl_generics sea_orm::FromQueryResult for #ident #ty_generics #where_clause {
-            fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> std::result::Result<Self, sea_orm::DbErr> {
-                Ok(Self {
-                    #(#fields)*
-                })
+            fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> Result<Self, sea_orm::DbErr> {
+                Ok(Self::from_query_result_nullable(row, pre)?)
             }
 
-            fn from_query_result_optional(row: &sea_orm::QueryResult, pre: &str) -> std::result::Result<Option<Self>, sea_orm::DbErr> {
+            fn from_query_result_nullable(row: &sea_orm::QueryResult, pre: &str) -> Result<Self, sea_orm::TryGetError> {
                 #(#ident_try_init)*
 
-                std::result::Result::Ok(std::option::Option::Some(Self {
+                Ok(Self {
                     #(#ident_try_assign)*
-                }))
+                })
             }
         }
     ))
 }
+
 mod util {
     use syn::Meta;
 
--- a/sea-orm-macros/src/derives/from_query_result.rs
+++ b/sea-orm-macros/src/derives/from_query_result.rs
@@ -3,7 +3,6 @@ use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::{
     ext::IdentExt, punctuated::Punctuated, token::Comma, Data, DataStruct, Fields, Generics, Meta,
-    Type,
 };
 
 enum ItemType {
--- a/sea-orm-macros/src/derives/from_query_result.rs
+++ b/sea-orm-macros/src/derives/from_query_result.rs
@@ -42,7 +42,7 @@ impl<'a> ToTokens for TryFromQueryResultCheck<'a> {
             ItemType::Nested => {
                 let name = ident.unraw().to_string();
                 tokens.extend(quote! {
-                    let #ident = match sea_orm::FromQueryResult::from_query_result_nullable(row, &format!("{pre}{}_", #name)) {
+                    let #ident = match sea_orm::FromQueryResult::from_query_result_nullable(row, &format!("{pre}{}-", #name)) {
                         Err(v @ sea_orm::TryGetError::DbErr(_)) => {
                             return Err(v);
                         }
--- a/sea-orm-macros/src/derives/from_query_result.rs
+++ b/sea-orm-macros/src/derives/from_query_result.rs
@@ -16,6 +16,15 @@ struct FromQueryResultItem {
     pub ident: Ident,
 }
 
+/// Initially, we try to obtain the value for each field and check if it is an ordinary DB error
+/// (which we return immediatly), or a null error.
+///
+/// ### Background
+///
+/// Null errors do not necessarily mean that the deserialization as a whole fails,
+/// since structs embedding the current one might have wrapped the current one in an `Option`.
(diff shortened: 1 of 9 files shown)


Goodjooy:
The nest feature you implements seems great, should I keep the PR I opened opening?

tests/partial_model_tests.rs
tyt2y3, line 1: Nice to see an example. Can you include some integration tests where we actually put this `Nest` into select queries? This can server as both example and testcase. Ideally, we'd have a hand-unrolled implementation of the macro and being able to compare it against the derive macro generated version.

tyt2y3 (commented):
I would really love to accept this PR and make a patch release. I wish we can have more throughout tests and examples.
For example, when I include this in changelog or documentation, how should I describe this feature? Actually, may be it'd be helpful to first show an example of the problem you are trying to solve?

Like, without this macro extension, I'd have to...
And now, with this feature, we can simply...

jreppnow:
@tyt2y3 Hi, thanks for the comments! I'll get on to addressing them now. 

Two things:

1) I will write an integration test, which should also serve as a good example for why this is so useful. What it allows you do is basically have different structs for subqueries of a larger join query, and re-use this in various places. Say I am selecting over table A, left-joining it with table B. In order to only get the fields I need, until now I would have to add all the columns of B into the struct I define for A (rendering it non-reusable in the process), and make them all optional (due to left-join), even if the column is NOT NULL in B. 

2) There is potentially a different way to implement this, without needing to actually specify the `nested` attribute. However, this would most likely involve doing something like `impl<T: TryGet> FromQueryResult for T { ... }`, which would potentially be a lot more intrusive and definitely a breaking change (as both traits are pub and not sealed..).

jreppnow:
Ah, also, I would appreciate if we could also get #2167 in the same release (however, that one is a lot more minor and has an easy workaround).

jreppnow:
I added a couple of test cases, but stumbled over a really annoying issue - when you nest a struct into another which both refer to columns of the same name, but from different tables, sqlx just overwrites the value of the one deserialized first with the second one.. 

The user can do a workaround (renaming the fields), but this is honestly really ugly and above all, very error-prone (as values actually get overwritten without warning in the "good" case). A better solution would be to do what other ORMs do and rename the columns in the query based on some random, unique string per FromQueryResult struct, but that would also involve changing the non-partial Model logic.

Possible solutions: 

1. Keep DerivePartialModel and FromQueryResult together 
   1. My solution would probably be to generate a compile-time random identifier per struct that implements FromQueryResult, and ensure that that is prefixed before every AS in the response query. 
   2. Involves changing a lot of other stuff. 
2. Introduce a new `DerivePartialModel` derive macro (maybe just `PartialModel`?), which would then produce an independent implementation of `FromQueryResult` and be incompatible with deriving `FromQueryResult`. Easiest solution, but involves some duplication + the old version stays in the API for a while, users would have to switch to the new version. 
3. Remove/deprecate `DerivePartialModel` and incorporate its functionality into `FromQueryResult`, which would then behave differently based on some flag.

tests/partial_model_tests.rs
```diff
@@ -66,3 +69,51 @@ struct NestOption {
...
 }
+
+#[sea_orm_macros::test]
+async fn partial_model_left_join_does_not_exist() {
+    use common::bakery_chain::*;
+
+    #[derive(FromQueryResult, DerivePartialModel)]
+    #[sea_orm(entity = "bakery::Entity")]
+    struct Bakery {
+        id: i32,
```
jreppnow, line 80: The issue occurs because this `id` field (and the `name` field below) have the same name as the ones in the `Cake` struct. SQLx does not differentiate between these two and just overrides the ones it encounters first (the ones for `Cake`) with the ones in it encounters next (the ones for `Bakery`). The "good" case (where the associated `Bakery` exists) is arguably the most broken, as the `id` and `name` values of the `Cake` get silently replaced with the ones of the `Bakery`.

jreppnow:
Think I figured out a workaround, need a little bit more time though.. I will notify you when I have something.

jreppnow:
Okay, I think I found a solution which will change slightly how some queries are generated, but should not affect the API otherwise (possible via patch release I would think). Basically, in order to combat the issue of the same id showing up twice in the same query, I use `AS` directives and and prefix the fields with the field they belong to in the parent struct, ONLY if they are nested (i.e., the behavior is the same if nesting is not used..).

jreppnow:
One caveat: I decided to also do an `AS` for the members of the lowest in the hierarchy. This changes existing queries, although in what I believe is a compatible way, since the rename is just to name that is expected anyway.

tyt2y3:
Thanks for the massive update! I am going through them

jreppnow:
@tyt2y3 Just an FYI, this PR (specifically a 0.12.x backport) has been in production use with us for about a month now, without any problems and without breaking any existing queries.

seijikun:
Cool stuff, thanks for working on this @jreppnow!
Since Sea-ORM's relationship handling is unfortunately rather weak; something I'm wondering when I see your integration test examples is, whether this supports:
- Multiple nested structs (in example: `Order` -> `{Product, Customer}`)
- Recursively nested structs (in example: `Order` -> `Customer` -> `Address`)

Example:
```mermaid
erDiagram
    Product ||--o{ Order : "id"
    Customer ||--o{ Order : "customer_id"
    Order ||--o{ Product : "product_id"
    Order {
        id UUID
        product_id UUID
        customer_id UUID
    }
    Address {
        id UUID
    }
    Customer {
        id UUID
        address_id UUID
    }
    Product {
        id UUID
    }

    Address ||--o{ Customer : "id"
```
```sql
CREATE TABLE "Address" (
    id UUID NOT NULL;
);
CREATE TABLE "Customer" (
    id UUID NOT NULL;
    address_id UUID NOT NULL REFERENCES Address("id");
);
CREATE TABLE "Order" (
    id UUID NOT NULL;
    product_id UUID NOT NULL REFERENCES Product("id");
    customer_id UUID NOT NULL REFERENCES Customer("id");
);
CREATE TABLE "Product" (
    id UUID NOT NULL;
);
```
```rust
struct Product; // [...]
struct Address; // [...]

#[derive(Debug, FromQueryResult, DerivePartialModel)]
#[sea_orm(entity = "Order")]
struct Order {
	id: Uuid,
	#[sea_orm(nested)]
	product: Product,
	#[sea_orm(nested)]
	customer: Customer,
}

#[derive(Debug, FromQueryResult, DerivePartialModel)]
#[sea_orm(entity = "Customer")]
struct Customer {
	id: Uuid,
	#[sea_orm(nested)]
	product: Address
}
```

I'd like to get a list of `Order`s, each with `Product`, `Customer` and `Customer -> Address` joined in one query.

jreppnow:
@seijikun Thanks! Both should work perfectly well (we have multiple layers of recursion as well as multiple nested structs within various places in our code). Should you try this out and it unexpectedly does not work for some reason, I would consider this a bug and try to fix it if you let me know. 

As you mentioned, relation handling is rather loose, so you need to write the joins by hand and keep them in sync with the structs.

seijikun:
@jreppnow I just tested your branch and I have to say ... it's fabulous!

Seems to work flawlessly. Also with a lot of column name collisions between all the joined entities.
I originally planned to use this to circumvent the limited `SelectTwo<>` stuff, but now I just directly load from the database into my rest/json view models for most cases.

Thanks again for working on this!

Alt-iOS:
Hi, this seems incredible useful!, Are there any plan to stabilize this into  `1.1.0-rc.x`?

Jacob-32587:
I would love to see this stabilized soon too! Coming from other ORM's this is one of the biggest things I miss.

plusls:
any progress？

jreppnow:
This PR is feature-complete since beginning of May. If I can get a review and some feedback (positive or negative), I will rebase it and get it ready-to-merge. 

@tyt2y3 I really don't like to put pressure on open source maintainers and have intentionally refrained from explicitly asking for another review/feedback, but this PR has been waiting for quite a while now, with a few people deeming it useful/desirable. It's still in use in production for us and has not caused any problems so far. Would you mind having another look (or, at least state that this PR does not match your vision and will not be merged)? Sorry!

CC @billy1624

ns-sjorgedeaguiar:
@jreppnow the way maintenance is being conducted is this project is really disappointing. I myself closed a PR for this exactly reason - lack of any input whatsoever from maintainers. On discord, help is exclusively being provided by other users and here the issues and PRs are mostly ignored. It's a shame to see the community being abandoned like that. This has actually led me to move to sqlx and giving up on SeaORM altogether in fear for the future.

Being busy and facing (personal) problems is completely understandable but there is no need for this lack of response. I'm sure there is a handful of people willing to get aboard and help. I hope that maintainers can eventually come clean about the situation and, if needed, welcome more members into the team. Fingers crossed that your PR does get merged and that your work does not go to waste.

Jacob-32587:
I have started to use this in my code base, merged the most recent version of `sea_orm` here: https://github.com/Jacob-32587/sea-orm-with-nested-structs. This is a huge win for simplifying my code, before I was mapping from the raw `Vec<QueryResults>` which was not fun. I wish nested `Vec<T>` was supported but understand why it isn't right now.

For all of us that only use `sea_query` for selecting entities from the database I have written a select statement extension that makes this much more seamless. If you have any suggestions/improvements let me know! Hope this gets approved soon! 

```rust
fn column_ref_into_alias_str(col_ref: &ColumnRef) -> Result<String, DbErr> {
    const UNSUPPORTED_VARIANT_ERR_MSG: &str =
        "Can not build alias unless column names and table name is known";
    match col_ref {
        ColumnRef::Column(_) => Err(DbErr::Custom(String::from(UNSUPPORTED_VARIANT_ERR_MSG))),
        ColumnRef::TableColumn(t, c) => {
            let mut temp = t.to_string();
            temp.push('-');
            temp.push_str(&c.to_string());
            Ok(temp)
        }
        ColumnRef::SchemaTableColumn(s, t, c) => {
            let mut temp = s.to_string();
            temp.push('-');
            temp.push_str(&t.to_string());
            temp.push('-');
            temp.push_str(&c.to_string());
            Ok(temp)
        }
        ColumnRef::Asterisk => Err(DbErr::Custom(String::from(UNSUPPORTED_VARIANT_ERR_MSG))),
        ColumnRef::TableAsterisk(_) => {
            Err(DbErr::Custom(String::from(UNSUPPORTED_VARIANT_ERR_MSG)))
        }
    }
}

pub trait SelectStatementExtensions {
    fn nested_alias<C, I>(&mut self, cols: I) -> &mut Self
    where
        C: IntoColumnRef,
        I: IntoIterator<Item = C>;
}

impl SelectStatementExtensions for SelectStatement {
    fn nested_alias<C, I>(&mut self, cols: I) -> &mut Self
    where
        C: IntoColumnRef,
        I: IntoIterator<Item = C>,
    {
        self.exprs(
            cols.into_iter()
                .map(|x| {
                    let col_ref = x.into_column_ref();
                    SelectExpr {
                        alias: Some(
                            Alias::new(column_ref_into_alias_str(&col_ref).unwrap()).into_iden(),
                        ),
                        expr: SimpleExpr::Column(col_ref),
                        window: None,
                    }
                })
                .collect::<Vec<SelectExpr>>(),
        )
    }
}
```
Example:
```rust
        ...
        Query::select()
            .columns(ingredient_get_by_id::Column::iter().map(|x| (Entity, x)))
            .nested_alias(only_name::Column::iter().map(|x| (serving_size_unit::Entity, x)))
            .nested_alias(brand_owner::Column::iter().map(|x| (brand_owner::Entity, x)))
            .from(Entity)
            ...
```

Sytten:
This would be useful for sure to reduce code duplication

sea-orm-macros/src/derives/from_query_result.rs
```diff
@@ -42,40 +101,54 @@ pub fn expand_derive_from_query_result(
...
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
+    let ident_try_init: Vec<_> = fields.iter().map(TryFromQueryResultCheck).collect();
+    let ident_try_assign: Vec<_> = fields.iter().map(TryFromQueryResultAssignment).collect();
+
     Ok(quote!(
         #[automatically_derived]
         impl #impl_generics sea_orm::FromQueryResult for #ident #ty_generics #where_clause {
-            fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> std::result::Result<Self, sea_orm::DbErr> {
+            fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> Result<Self, sea_orm::DbErr> {
```
Sytten, line 133: please use std::result::Result

sea-orm-macros/src/derives/from_query_result.rs
```diff
@@ -42,40 +101,54 @@ pub fn expand_derive_from_query_result(
...
+                Ok(Self::from_query_result_nullable(row, pre)?)
+            }
+
+            fn from_query_result_nullable(row: &sea_orm::QueryResult, pre: &str) -> Result<Self, sea_orm::TryGetError> {
```
Sytten, line 137: Same here, use std::result::Result

Sytten (changes requested):
Two small bugs

sea-orm-macros/src/derives/from_query_result.rs
```diff
@@ -5,22 +5,81 @@ use syn::{
...
+            }
+            ItemType::Skipped => {
+                tokens.extend(quote! {
+                    let #ident = std::default::Default::default();
+                });
+            }
+            ItemType::Nested => {
+                let name = ident.unraw().to_string();
+                tokens.extend(quote! {
+                    let #ident = match sea_orm::FromQueryResult::from_query_result_nullable(row, &format!("{pre}{}-", #name)) {
```
Sytten, line 54: Adding the package name to the prefix makes the whole thing unusable unless you are using it with a `DerivePartialModel` because the base model don't add the entity name in front of the column. An integration test would have caught that easily.

Sytten (changes requested):

sea-orm-macros/src/derives/from_query_result.rs
jreppnow: 0. This is not the package name but a prefix computed based on the location (property name) of the nested struct in the parent struct. 
1. There are integration tests.
2. `nested` only makes sense when used with `DerivePartialModel`. 
3. (I have specified this in the comments explicitly as well) This is required to avoid name collisions.

sea-orm-macros/src/derives/from_query_result.rs
jreppnow: Valid criticism. The correct way is to use `::std::..` or `::sea_orm::..` everywhere, so I will adjust it to that.

sea-orm-macros/src/derives/from_query_result.rs
jreppnow: Ah, this is about using a struct with `nested` in it in `into_model::<..>()`, I guess? Yup, that is most likely going to break, but due to the name collision thing above there is not really anything I can do about it. The relationship between `FromQueryResult` and `DerivePartialModel` is weird and they should arguably be mutually exclusive, but that would involve copying the entire code from `FromQueryResult` into `DerivePartialModel`. That would also mean doing things like the `skip` I proposed elsewhere twice:

tyt2y3:
sorry for the late reply, it seems like rebasing on top of master would fix the test failures?

sea-orm-macros/src/derives/from_query_result.rs
Sytten: I dont remember what this was really but it didnt work when I wanted to do:
```
#[derive(FromQueryResult)]
pub struct PackageWithDownloads {
   #[sea_orm(nested)]
   pub package: plugin_packages::Model,
   pub downloads: i64,
}
```
I fixed it in my fork by removing the format,see https://github.com/caido/dependency-sea-orm/commit/30cdea1babec257073e160b7927fcd1255fe8f92#diff-50509553d7ac1acce8ee99595b98afc8307096fe6fa1a85056e00a29eba829e1R53

tyt2y3:
Thank you all. I'll take a closer look with @billy1624

sea-orm-macros/src/derives/from_query_result.rs
jreppnow: Unfortunately, this change does not work with the plan model structs (and I don't know an easy way to make it work). You need the prefixing because otherwise you get name collisions between common names like "id" and such, so the `format` cannot be removed.

jreppnow:
Thank you. I'm on vacation right now, but I can rebase again when I get back.

sea-orm-macros/src/derives/from_query_result.rs
tyt2y3: I think I might get it now. so there are two scenarios:

1. only FromQueryResult. and we don't need prefix
2. both FromQueryResult and DerivePartialModel. we need to put prefix

tyt2y3: I think the solution is for `DerivePartialModel` to also assumes the function of `FromQueryResult`.
So we will no longer recommend using both macros together.

tyt2y3:
I think I've got it working in https://github.com/SeaQL/sea-orm/pull/2508

tyt2y3:
Done!

PR #2180, opened 2024-03-29
Introduce ARITY to PrimaryKeyTrait

tyt2y3:
This is one of the last breaking changes I want to ship in 1.0

### DIFF
--- a/src/executor/insert.rs
+++ b/src/executor/insert.rs
@@ -116,7 +116,7 @@ where
     {
         // so that self is dropped before entering await
         let mut query = self.query;
-        if db.support_returning() && <A::Entity as EntityTrait>::PrimaryKey::iter().count() > 0 {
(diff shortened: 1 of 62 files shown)


tyt2y3:
@billy1624 can you help looking at the test failure?

src/executor/insert.rs
```diff
@@ -116,7 +116,7 @@ where
     {
         // so that self is dropped before entering await
         let mut query = self.query;
-        if db.support_returning() && <A::Entity as EntityTrait>::PrimaryKey::iter().count() > 0 {
```
tyt2y3, line 119: I believe all our primary keys has len > 0, so this check is redundant?

billy1624: Redundant indeed!

tyt2y3:
Closed in favour of #2185

PR #2184, opened 2024-04-02
Simpler and ergonomics commit and rollback.

sirlordt:
When you use commit or rollback the methods are self consuming. This is very confusing for new rust users who do not understand that the transaction variable can no longer be used after the commit or rollback method is called. Especially if you come from environments like Java or C#, where the commit or the rollback don't have this behavior.

Additionally, these self-consuming methods are more difficult to use with Arc<Mutex<DatabaseTransaction>> because it is necessary to extract the value of the Arc and the Mutex. To be able to call the Commit or the Rollback. That is much more confusing and unergonomic.

That is why alternative methods commit_by_ref and rollback_by_ref are proposed that allow commit and rollback to be used in a simpler way.

Thanks.

## New Features

- [commit_by_ref and rollback_by_ref methods to the transaction object]

## Bug Fixes

- [none]

## Breaking Changes

- [none]

## Changes

- [none]

### DIFF
--- a/src/database/transaction.rs
+++ b/src/database/transaction.rs
@@ -112,6 +112,12 @@ impl DatabaseTransaction {
     #[instrument(level = "trace")]
     #[allow(unreachable_code, unused_mut)]
     pub async fn commit(mut self) -> Result<(), DbErr> {
+        self.commit_by_ref().await
+    }
+
+    #[instrument(level = "trace")]
+    #[allow(unreachable_code, unused_mut)]
+    pub async fn commit_by_ref(&mut self) -> Result<(), DbErr> {
         match *self.conn.lock().await {
             #[cfg(feature = "sqlx-mysql")]
             InnerConnection::MySql(ref mut c) => {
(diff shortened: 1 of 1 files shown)


jreppnow:
(Opinion as a fellow library user) Using the transaction after either the rollback or the commit is *always* an error and explicitly allowing that seems like a really bad idea. Speaking as someone working with majority Rust beginners, enabling error-prone code is certainly not the way to teach Rust (and make them appreciate it). For my beginner colleagues' sake as well I would much rather this stays not an option.

Also, cannot speak for your code of course, but having the transaction behind a mutex smells like an anti-pattern. Even then if you must, turning it into an `Mutex<Option<Transaction>>` is really easy and illegal state stays unrepresentable as it should.

tyt2y3:
Not trying to start a debate, but I agree with @jreppnow . Reusing the object afterwards is always an error. And the [drop](https://doc.rust-lang.org/std/mem/fn.drop.html) function is magically defined as a 'self consuming' function.

But thank you for your sharing all-in-all

tyt2y3:
I am really glad that you have found success with Rust / SeaORM. Just my two cents, I'd probably do message passing and have one single `ServiceHandler` in the app. There are many good mpsc channel options in Rust.

PR #2197, opened 2024-04-14
Add `ActiveValue::try_as_ref()`

Expurple:
Another helper method that I use very often and would like to upstream. Basically, a non-panicking version of [ActiveValue::as_ref](https://docs.rs/sea-orm/latest/sea_orm/entity/enum.ActiveValue.html#method.as_ref).

## PR Info

- Closes

- Dependencies:

- Dependents:

## New Features

- `ActiveValue::try_as_ref()`

## Changes

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -849,6 +849,26 @@ where
             None => ActiveValue::NotSet,
         };
     }
+
+    /// Get the inner value, unless `self` is [NotSet][ActiveValue::NotSet].
+    ///
+    /// There's also a panicking version: [ActiveValue::as_ref].
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// # use sea_orm::ActiveValue;
+    /// #
+    /// assert_eq!(ActiveValue::Unchanged(42).try_as_ref(), Some(&42));
+    /// assert_eq!(ActiveValue::Set(42).try_as_ref(), Some(&42));
+    /// assert_eq!(ActiveValue::NotSet.try_as_ref(), None::<&i32>);
+    /// ```
+    pub fn try_as_ref(&self) -> Option<&V> {
+        match self {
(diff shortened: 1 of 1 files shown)


tyt2y3:
Thanks. But is `try_as_ref` the best name? I am thinking of something similar in the Rust ecosystem

Expurple:
I think, it's a good name because it's consistent with the existing `as_ref`, and there's also a third party [try_as_traits::TryAsRef::try_as_ref](https://docs.rs/try_as_traits/latest/try_as_traits/trait.TryAsRef.html#tymethod.try_as_ref) for using this pattern generically. I've decided against adding an extra dependency because I don't need to be abstract & generic, and inherent methods are more discoverable.

In my dependent codebase, this helper is called `ActiveValueExt::value`. But when I was upstreaming it, I discovered that this name would be confusing in combination with existing `into_value` and `into_wrapped_value` which return `Value` rather than `V`.

Overall, it's up to you. I'm ok with other names, as long as they don't contain `value`. E.g. `as_ref` (if it wasn't already taken), `borrow` (you may not like it because it sounds like `Borrow::borrow` but has a different signature), `get`...

billy1624:
Hey @Expurple, thanks for contributing!!

Maybe `value` or `as_value`? We already have `unwrap`, `into_value` and `into_wrapped_value` methods.

https://github.com/SeaQL/sea-orm/blob/c59e28f8c0319accc7177d3c37fe7bd2a8892b82/src/entity/active_model.rs#L815-L842

Expurple:
Just choose any one name that you agree to merge, then I make the edit and you merge it. The final approve decision is up to you anyway. I've already listed some options and explained why I prefer `try_as_ref`/`borrow`/`get` to `value`/`as_value`, but all of these will do the job and I'm not interested in bikeshedding

billy1624:
Hey @tyt2y3, perhaps `value` / `as_value`?

tyt2y3 (approved):
Thank you for the explanation!

PR #2199, opened 2024-04-17
use @generated mark in files created by `sea-orm-codegen`

RafDevX:
Changes the doc header on generated files to include `@generated`.  This makes it easier for external tools to identify these files as generated and treat them differently according to their own configuration, e.g., code formatters might ignore them. This is the *de facto* way to indicate a file has been automatically generated, and even `Cargo.lock` uses it.

This can perhaps be considered a breaking change if someone is relying on the current exact header, so an alternative would be to keep the current comment and add another line after it with `// @generated`. On the other hand, that would mean almost anything at all changing this crate's writers would too be a breaking change, which is debatable -- this PR does not introduce any changes to e.g. function signatures.

## Breaking Changes

- [x] Different doc header with `@generated`


tyt2y3 (approved):
That's neat. Thanks!

tyt2y3:
The RC period is for breaking changes :D

PR #2230, opened 2024-05-20
Getting Started with Loco & SeaORM

billy1624:


examples/loco_starter/src/models/_entities/prelude.rs
```diff
@@ -0,0 +1,3 @@
+//! `SeaORM` Entity. Generated by sea-orm-codegen 0.12.4
```
chris-numeus, line 1: Um... we should use the 1.0 codegen?

examples/loco_starter/src/models/_entities/files.rs
```diff
@@ -0,0 +1,31 @@
+//! `SeaORM` Entity. Generated by sea-orm-codegen 0.12.4
```
chris-numeus, line 1: We should use the 1.0 codegen

tyt2y3, line 1: We should use the 1.0 codegen

tyt2y3:
Would appreciate a screenshot. For the postman collection, may be we can commit it too

examples/loco_starter/src/models/_entities/files.rs
billy1624: I removed these lines https://github.com/SeaQL/sea-orm/pull/2230/commits/fea1725b8bee19acd97920895652c384e3d24acf

PR #2233, opened 2024-05-22
Exposing connection pools for sqlite

ys928:
## Changes

- Provide a function to expose sqlite connection pools

### DIFF
--- a/src/driver/sqlx_sqlite.rs
+++ b/src/driver/sqlx_sqlite.rs
@@ -239,6 +239,11 @@ impl SqlxSqlitePoolConnection {
         self.pool.close().await;
         Ok(())
     }
+
+    /// Get SqlitePool
+    pub fn get_pool(&self) -> &SqlitePool {
+        &self.pool
+    }
 }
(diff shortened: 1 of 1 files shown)


tyt2y3:
I think this might be what you're looking for? https://github.com/SeaQL/sea-orm/blob/010433bf1633dd387f8493071c0b3838a7fbb575/src/database/db_connection.rs#L537

ys928:
thanks, I didn't find it beforehand, turns out I also needed to add a sea-orm-internal feature to show this function in vscode's autocompletion

tyt2y3:
I think I will expose it by default moving forward

ys928:
That's great, but I think this function should be included in the sqlx-sqlite feature.After all, it might not make sense to be able to use it without enabling this feature.

tyt2y3:
Yes, the method itself is guarded by `sqlx-sqlite`

ys928:
Oh, indeed, I was wrong, everything is fine.

Issue #1884, opened 2023-09-28
`has_one` relation is not recognized with primary foreign key (class table inheritance)


Expurple:
## Description

When I use class table inheritance, sea-orm generates a `has_many` relation from Parent to Child instead of `has_one`. This issue is somewhat similar to #1393, but my foreign key is also a primary key and this issue happens with both PostgreSQL and SQLite. PostgreSQL is used by the actual codebase where I first encountered this, and I chose SQLite for the reproducer to make it more simple and portable.

## Steps to Reproduce

1. [Set up](https://www.sea-ql.org/sea-orm-tutorial/ch01-02-migration-cli.html) a dummy sea-orm project with migrations and with SQLite driver enabled.
2. Paste the migration code from the [Reproducible Example](#reproducible-example) section.
3. Generate entities:
    ```bash
    touch db.sqlite3
    DATABASE_URL="sqlite://./db.sqlite3" sea-orm-cli migrate fresh
    sea-orm-cli generate entity -u "sqlite://./db.sqlite3" -o src/entities
    ```
4. Open `src/entities/parent.rs`

### Expected Behavior

I expected to see `has_one` instead of `has_many`.

### Actual Behavior

```rust
// In generated src/entities/parent.rs

#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
pub enum Relation {
    #[sea_orm(has_many = "super::child::Entity")]
    Child,
}
```

### Reproduces How Often

Always.

### Workarounds

In the actual codebase with PostgreSQL and different migration code, if I turn the Child's `.primary_key()` into a `.unique_key()`, `has_one` is generated as expected. However, the Child entity doesn't compile afterwards because it must have a primary key. In the reproducer migration with SQLite this doesn't happen (it still generates `has_many`). I didn't investigate the reason yet. Ping me if this is important.

## Reproducible Example

```rust
use sea_orm_migration::prelude::*;

#[derive(DeriveMigrationName)]
pub struct Migration;

#[async_trait::async_trait]
impl MigrationTrait for Migration {
    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
        manager
            .create_table(
                Table::create()
                    .table(Parent::Table)
                    .if_not_exists()
                    .col(
                        ColumnDef::new(Parent::Id)
                            .integer()
                            .not_null()
                            .auto_increment()
                            .primary_key(),
                    )
                    .to_owned(),
            )
            .await?;
        manager
            .create_table(
[...]

PR #1629, opened 2023-05-03
primary key also is not null and unique key

jm-observer:
## Bug Fixes

- primary key also is not null and unique key

PR #2236, opened 2024-05-24
Mark primary key columns as unique in sea-orm-codegen

SZenglein:
## PR Info

This is a relatively simple change where sea-orm-cli did not mark primary key columns as unique, resulting in a has_many relation where a has_one would be correct. Issue #1884 offers a complete example where the problem occurs.

I only tested this for SQLite, but it should be pretty universal.

Maybe related: https://github.com/SeaQL/sea-orm/pull/1629
Fixes: #1884 

## New Features
None

## Bug Fixes

sea-orm-cli no longer incorrectly assumes primary key columns as non-unique.

## Breaking Changes

Code generation will be different in some cases (has_many replaced by has_one).

### DIFF
--- a/sea-orm-codegen/src/entity/transformer.rs
+++ b/sea-orm-codegen/src/entity/transformer.rs
@@ -50,7 +50,7 @@ impl EntityTransformer {
                     col.unique = table_create
                         .get_indexes()
                         .iter()
-                        .filter(|index| index.is_unique_key())
+                        .filter(|index| index.is_unique_key() || index.is_primary_key())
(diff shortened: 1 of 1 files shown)


billy1624 (commented):
Hey @SZenglein, thanks for the PR. Sorry, I don't think this solve the root cause of the issue. I just made a PR, https://github.com/SeaQL/sea-orm/pull/2254, would you like to take a look and try it?

Install `sea-orm-cli` and try it on your local machine:

```sh
cargo install sea-orm-cli --force --git https://github.com/SeaQL/sea-orm --branch cli-gen-has-one-relation
```

PR #2265, opened 2024-06-20
Access the underlying row type of `QueryResult`

Expurple:
I have a use case with dynamic queries, similar to what's described in #2148. Using an existing SeaORM connection (switching to another library is not an option), I need to execute dynamic queries and then process the result by inspecting the returned columns, their types, type names, values in rows...

At the moment, [QueryResult](https://docs.rs/sea-orm/latest/sea_orm/struct.QueryResult.html) doesn't provide all necessary info to do that. [sqlx::Row](https://docs.rs/sqlx-core/0.7.3/sqlx_core/row/trait.Row.html) trait does, but it's not implemented for [QueryResult](https://docs.rs/sea-orm/latest/sea_orm/struct.QueryResult.html). In theory, I could've tried to implement it. But I couldn't immediately tell if it's possible or how long it's going to take, given that [MockRow](https://docs.rs/sea-orm/latest/sea_orm/struct.MockRow.html) and [ProxyRow](https://docs.rs/sea-orm/latest/sea_orm/struct.ProxyRow.html) don't implement it either. And I don't actually care about those, because I only ever use real Postgres!

So I went with an approach that's quick and guaranteed to provide all necessary info. I simply added direct access to the underlying [sqlx::PgRow](https://docs.rs/sqlx-postgres/latest/sqlx_postgres/struct.PgRow.html) which already implements [sqlx::Row](https://docs.rs/sqlx-core/0.7.3/sqlx_core/row/trait.Row.html). I also reexported `sea_orm::sqlx` in order to have guaranteed access to `sqlx` types at correct versions and feature flags. This worked really well and I already depend on [my fork of 0.12.15](https://github.com/Expurple/sea-orm/tree/get-sqlx-row) in my project.

I'd like to upstream it. To make a better contribution, here I also added equivalent methods for all other databases.

Tests don't seem to be necessary here. These accessor methods are trivial and I'm not even sure how to manually construct a `QueryResult` in a public doctest.

## PR Info

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- `QueryResult::try_as_mysql_row`
- `QueryResult::try_as_pg_row`
- `QueryResult::try_as_sqlite_row`
- `QueryResult::try_as_mock_row`
- `QueryResult::try_as_proxy_row`
- reexported `sea_orm::sqlx`

## Changes

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -150,6 +150,56 @@ impl QueryResult {
             _ => unreachable!(),
         }
     }
+
+    /// Access the underlying `MySqlRow` if we use the MySQL backend.
+    #[cfg(feature = "sqlx-mysql")]
+    pub fn try_as_mysql_row(&self) -> Option<&sqlx::mysql::MySqlRow> {
+        match &self.row {
+            QueryResultRow::SqlxMySql(mysql_row) => Some(mysql_row),
+            #[allow(unreachable_patterns)]
+            _ => None,
+        }
+    }
+
+    /// Access the underlying `PgRow` if we use the Postgres backend.
+    #[cfg(feature = "sqlx-postgres")]
+    pub fn try_as_pg_row(&self) -> Option<&sqlx::postgres::PgRow> {
+        match &self.row {
+            QueryResultRow::SqlxPostgres(pg_row) => Some(pg_row),
+            #[allow(unreachable_patterns)]
+            _ => None,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Expurple:
@tyt2y3 @billy1624 is there any chance this can get merged? I'm still using the fork :(

tyt2y3 (approved):
Sorry for the delay

PR #2271, opened 2024-06-25
Case-sensitive enum type casts

niklaskorz:
## PR Info

- Closes https://github.com/SeaQL/sea-query/issues/725
- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/789

## Bug Fixes

- [x] Filtering on enum fields whose type name was declared case sensitive no longer crashes

## Breaking Changes

- [x] Enum names are now case sensitive, i.e. `#[sea_orm(enum_name = "MyType")]` will result in `"MyType"` instead of `mytype` for PostgreSQL.


niklaskorz:
Tests are only failing because they depend on https://github.com/SeaQL/sea-query/issues/725

PR #2284, opened 2024-07-11
Custom migration dir

ns-sjorgedeaguiar:
## New Features

Allow customizing the migration directory in `sea-orm-migration` CLI. This allow embedding migrations to an existing crate (and/or inside a subdirectory) instead of enforcing a separate crate.

This also aligns the behavior of the tool with `sea-orm-cli`, where the migration directory is already customizable.

## Changes

The entrypoint was adjusted as to facilitate testing, and tests are added for the `migrate` command.
The migration directory can be specified via command line argument or environment variable, but the same default value (`./`) is still used if unspecified.

### DIFF
--- a/sea-orm-migration/Cargo.toml
+++ b/sea-orm-migration/Cargo.toml
@@ -32,6 +32,7 @@ futures = { version = "0.3", default-features = false, features = ["std"] }
 
 [dev-dependencies]
 async-std = { version = "1", features = ["attributes", "tokio1"] }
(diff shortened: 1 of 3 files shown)


ns-sjorgedeaguiar:
@tyt2y3 @billy1624 could you check this one please?

ns-sjorgedeaguiar:
Closing due to no response / lack of interest

DenuxPlays:
thats so sad :/

PR #2321, opened 2024-08-07
call `EnumIter::get` using fully qualified syntax

jcreekmore:
## Bug Fixes

`itertools` version 0.13.0 added a `get` method that takes precedence over the `EnumIter::get` method. The `EnumIter` derive should be calling the method fully-qualified to ensure that the local version is called rather than the method that is provided by the `Itertools` trait. This was fixed in `strum` 0.26.x, but since this is a forked/adapted version of the `EnumIter` derive, it requires the same fix .


Expurple:
Oh, so that's what has been preventing me from updating to `itertools` 0.13! Thank you.

jcreekmore:
Is there something I am missing or need to do to get a review?

Expurple:
For the last few months, maintainers pay very little attention to community issues/PRs :(

It's weird, given that they still release new versions, blog posts, etc. They're probably focusing on their own checklist for `sea-orm` 1.0 and then `sea-orm` 1.1.0 with `sqlx` 0.8, although it's hard to say for sure.

billy1624 (commented):
Hey @jcreekmore, sorry for the delay! I will merge this today and publish a new RC with this.

PR #2344, opened 2024-08-30
Fix lib panic on insert_many

j-mendez:
This removes the panic that happens on insert many. 
We should not have panics in a library production code.


tyt2y3:
sorry, I don't think this is a proper fix, yes it removes the panic, but the resulting query will still be wrong.
ideally, we'd have to loop over the models a second time with a complete list of all columns to be constucted in the insert statement

PR #2348, opened 2024-09-03
`DatabaseConnection` directly from `PgPool`

fosskers:
### PR Info

This allows sqlx-based connection pools that were already on-hand to be easily converted into SeaORM types.

### New Features

Two new `From` impls which allow a native `PgPool` to be easily converted into SeaORM wrapper types that can then be used for SeaORM logic.


nicolas-vivot (approved):

billy1624 (commented):
Hey @fosskers, good idea! I also implemented for `MySqlPool` and `SqlitePool`

PR #2349, opened 2024-09-03
Use input ident instead of associated type projections in ActiveModel derive

svix-jplatte:
## Overview

Currently, sea-orm's `ActiveModel` derive generates impls like `impl From<<Entity as sea_orm::EntityTrait>::Model> for ActiveModel`. This prevents any other `From<LocalType> for ActiveModel` impls from being added from other crates that depend on the crate that defines the `ActiveModel`. This is a bug in rustc: https://github.com/rust-lang/rust/issues/85898

This PR works around that by not using associated type projections for these impls. I couldn't figure out why this was done in the first place, as it seems like the real type this would resolve to would always be the input type of the derive macro. Maybe I'm wrong about that.

I found that the associated type projection was introduced in https://github.com/SeaQL/sea-orm/commit/890464f9137fcf8e8f3f28c62a6c818ea37a45ac, and an example was updated but at the same time but I fail to see how it would be impacted by this change. Maybe CI will tell.

## Changes

- [ ] Simplify `ActiveModel` derive output to work around a compiler bug

### DIFF
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -22,8 +22,8 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
     }
     .into_iter();
 
-    let derive_active_model = derive_active_model(all_fields.clone())?;
-    let derive_into_model = derive_into_model(all_fields)?;
+    let derive_active_model = derive_active_model(&ident, all_fields.clone())?;
+    let derive_into_model = derive_into_model(&ident, all_fields)?;
 
     Ok(quote!(
         #derive_active_model
@@ -31,7 +31,7 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
     ))
 }
 
-fn derive_active_model(all_fields: IntoIter<Field>) -> syn::Result<TokenStream> {
+fn derive_active_model(ident: &Ident, all_fields: IntoIter<Field>) -> syn::Result<TokenStream> {
     let fields = all_fields.filter(field_not_ignored);
 
     let field: Vec<Ident> = fields.clone().map(format_field_ident).collect();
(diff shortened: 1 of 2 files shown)


svix-jplatte:
Gentle ping @tyt2y3

tyt2y3:
GitHub seems to have skipped the button to let us run CI

sea-orm-macros/src/derives/active_model.rs
```diff
@@ -22,16 +22,16 @@ pub fn expand_derive_active_model(ident: Ident, data: Data) -> syn::Result<Token
     }
     .into_iter();
 
-    let derive_active_model = derive_active_model(all_fields.clone())?;
-    let derive_into_model = derive_into_model(all_fields)?;
+    let derive_active_model = derive_active_model(&ident, all_fields.clone())?;
```
tyt2y3, line 25:
```suggestion
    // test
    let derive_active_model = derive_active_model(&ident, all_fields.clone())?;
```

tyt2y3:
Now it's running

tyt2y3:
> This prevents any other From<LocalType> for ActiveModel impls from being added from other crates that depend on the crate that defines the ActiveModel

@svix-jplatte can you also include a test case for this? I want to understand what did not compile for you, and be sure that we'd don't break it in the future.

svix-jplatte:
This issue only manifests with multiple crates involved. Should I create two dummy crates that are just a regular part of the workspace and compiled as part of it? Where would I put them, `tests/`? They would be different from the other tests in that they're not actually test binaries, but individual cargo packages with their own `Cargo.toml`s.

tyt2y3:
Can you put the entity in https://github.com/SeaQL/sea-orm/tree/master/src/tests_cfg and then use it in integration tests?
I think the crate boundary would work here?

svix-jplatte:
Ping @tyt2y3 looks like CI has succeeded :)

PR #2353, opened 2024-09-09
Update `ouroboros`

jaymell:
0.17.x has a dependency on `proc-macro-error`, which is unmaintained 
per https://rustsec.org/advisories/RUSTSEC-2024-0370


atezet:
Any update on this? Would be great if it could be included in 1.1

tyt2y3:
Sure! thank you

billy1624 (approved):
Thanks!! @jaymell

PR #2371, opened 2024-09-29
Update sqlx version to 0.8.2

### LINKED ISSUE #2361: Sqlx vulnerability according to Dependabot
Should upgrade sqlx to v0.8.1

---

The following presentation at this year's DEF CON was brought to our attention on the SQLx Discord:

SQL Injection isn't Dead: Smuggling Queries at the Protocol Level
http://web.archive.org/web/20240812130923/https://media.defcon.org/DEF%20CON%2032/DEF%20CON%2032%20presentations/DEF%20CON%2032%20-%20Paul%20Gerste%20-%20SQL%20Injection%20Isn't%20Dead%20Smuggling%20Queries%20at%20the%20Protocol%20Level.pdf
(Archive link for posterity.)
Essentially, encoding a value larger than 4GiB can cause the length prefix in the protocol to overflow,
causing the server to interpret the rest of the string as binary protocol commands or other data.

It appears SQLx does perform truncating casts in a way that could be problematic,
for example: https://github.com/launchbadge/sqlx/blob/6f2905695b9606b5f51b40ce10af63ac9e696bb8/sqlx-postgres/src/arguments.rs#L163

This code has existed essentially since the beginning,
so it is reasonable to assume that all published versions <= 0.8.0 are affected.

ethandmd:
## PR Info

- Closes #2361

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ] Bumps `sqlx` from `0.8` to `0.8.2` to mitigate security vulnerability concern.


billy1624 (approved):
Thanks!! @ethandmd

PR #2390, opened 2024-10-10
Add support to client SSL PEM cert and key and root CA PEM cert

elichai:
## PR Info
Currently if you want to use SSL client certificate and key you can currently do it via a few options:
1. Manually create a `PgPool` or `MySqlPool` and create a `DatabaseConnection`.
2. For postgres:
  A. You can set `sslrootcert=PATH`, `sslkey=PATH`, `sslca=PATH` in the URI params.
  B. You can set env variables `PGSSLCERT=PATH`, `PGSSLKEY=PATH`, `PGSSLROOTCERT=PATH`
3. For mysql: You can set `ssl-cert`, `ssl-key=PATH`, `ssl-ca=PATH` in the URI params

Both `3` and `2` require you to mount the keys/certs to a volume and pass a PATH to that.
Only 1 allows you to pass the certs/keys themselves which allows nicer integrations with KMS/Secret Manager but comes at the cost of building the pools yourself,
Which is why I only added the `_pem` variants of `sqlx`

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -63,6 +63,14 @@ pub struct ConnectOptions {
     /// Schema search path (PostgreSQL only)
     pub(crate) schema_search_path: Option<String>,
     pub(crate) test_before_acquire: bool,
+    /// Sets whether or with what priority a secure SSL TCP/IP connection will be negotiated
+    pub(crate) ssl_mode: Option<SslMode>,
+    /// Sets the SSL client certificate as a PEM-encoded byte slice.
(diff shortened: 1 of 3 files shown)


tyt2y3:
Small nit, can you use the casing `SslMode`?

elichai:
> Small nit, can you use the casing `SslMode`?

Sure! Will fix :)

elichai:
@tyt2y3 Sorry that it took so long, but rebased and fixed

PR #2393, opened 2024-10-14
Let `last_insert_id` optional.

langyo:
## PR Info

- Dependencies: #1875 

I'm sorry for submitting such a PR so presumptuously. I found that the type of `last_insert_id` is flawed.

- For Postgres, a valid `last_insert_id` will not be returned anyway, because Postgres may directly use a non-integer primary key.
- The primary key specified by the table may not be able to auto-increment. For cases where auto-increment is not possible, we should not return a specific ID.
- For proxy connection type databases, whether to return a valid `last_insert_id` is entirely up to the user. If the user cannot get a valid ID from the actual database transferred (for example, for databases such as `surrealdb`, their IDs are UUIDs rather than integers), it is unreasonable to return a 0 directly.

This change is likely to affect a considerable number of downstream libraries, so it is a breaking change. Whether to make such a change requires discussion and decision.

## Breaking Changes

- [x] Let `last_insert_id` optional.

### DIFF
--- a/src/database/proxy.rs
+++ b/src/database/proxy.rs
@@ -31,14 +31,14 @@ pub trait ProxyDatabaseTrait: Send + Sync + std::fmt::Debug {
 #[derive(Clone, Debug, Default, serde::Serialize, serde::Deserialize)]
 pub struct ProxyExecResult {
     /// The last inserted id on auto-increment
-    pub last_insert_id: u64,
+    pub last_insert_id: Option<u64>,
     /// The number of rows affected by the database operation
     pub rows_affected: u64,
 }
 
 impl ProxyExecResult {
     /// Create a new [ProxyExecResult] from the last inserted id and the number of rows affected
-    pub fn new(last_insert_id: u64, rows_affected: u64) -> Self {
+    pub fn new(last_insert_id: Option<u64>, rows_affected: u64) -> Self {
         Self {
             last_insert_id,
             rows_affected,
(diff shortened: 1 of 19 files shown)


langyo:
cc @billy1624 @tyt2y3 
Please take a look at whether this modification is feasible if you have anytime..

tyt2y3:
hi, thank you for your contribution. this is such a big breaking change. may I ask what's the original usecase for it?

can you provide an example?

langyo:
> hi, thank you for your contribution. this is such a big breaking change. may I ask what's the usecase for it?
> 
> can you provide an example?
> 
> we have a [TryInsert](https://docs.rs/sea-orm/latest/sea_orm/query/struct.TryInsert.html) struct, may be it is more suitable for what you're trying to do?

The example revision is already included in this PR.

By the way, I took a look at the implementation of `TryInsert`. I found that the return value of [`exec_without_result`](https://docs.rs/sea-orm/latest/sea_orm/query/struct.TryInsert.html#method.exec_without_returning) might be unreasonable. Specifically, even if the database does not necessarily need to return the object just inserted after execution, the index ID returned by the database is not necessarily of the `u64` type, but needs to follow the `<dyn EntityTrait>::PrimaryKey` type.

I am considering whether I need to make further improvements to the question I just asked..

tyt2y3:
@langyo thanks again for your contribution. I agree with most of your points. sorry for the late action, but can you take a look and see if #2628 addresses your concerns?

langyo:
> @langyo thanks again for your contribution. I agree with most of your points. sorry for the late action, but can you take a look and see if #2628 addresses your concerns?

I took a quick look at the new PR. It basically meets the criteria I mentioned before. Thanks! I hope these new changes can be seen in the new version as soon as possible. I'll close the old PR.

PR #2397, opened 2024-10-19
Allow modifying the connection in migrations

Lythenas:
## New Features

Add a function to allow modifying the database connection in migrations.

This allows configuring the database connection like you do in your real application. I'm currently using this to register sqlite functions.

Usage would be something like this:

```rs
#[async_std::main]
async fn main() {
    cli::run_cli_with_connection(migration::Migrator, |connect_options| async {
        let db = Database::connect(connect_options)
             .await?;
        if db.get_database_backend() == DatabaseBackend::Sqlite {
            db::register_sqlite_functions(&db).await;
        }
        return Ok(db);
    }).await;
}
```

Note: I didn't see a test for `run_cli` except in the examples, so I wasn't sure if/where to add a test.

I also looked at adding it as a function to `MigratorTrait` but there we could at most accept a `&DbConn` (which would allow my use case) but I think this way is more flexible. The closure is accepting `ConnectOptions` so we can continue to use the same cli options and `DATABASE_URL` var. But it still allows to modify it before creating the connection or throw it away and create your own, if needed.

### DIFF
--- a/sea-orm-migration/src/cli.rs
+++ b/sea-orm-migration/src/cli.rs
@@ -19,7 +19,7 @@ where
     run_cli_with_connection(migrator, Database::connect).await;
 }
 
-/// Same as [`run_cli`] but you crate the [`DbConn`] yourself.
+/// Same as [`run_cli`] where you provide the function to create the [`DbConn`].
 ///
 /// This allows configuring the database connection as you see fit.
 /// E.g. you can change settings in [`ConnectOptions`] or you can load sqlite
--- a/sea-orm-migration/src/cli.rs
+++ b/sea-orm-migration/src/cli.rs
@@ -16,24 +16,7 @@ pub async fn run_cli<M>(migrator: M)
 where
     M: MigratorTrait,
 {
-    dotenv().ok();
-    let cli = Cli::parse();
-
-    let url = cli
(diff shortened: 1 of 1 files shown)


tyt2y3 (commented):
thank you for your contribution. I think this is a nice idea, however can you make `run_cli` delegate to `run_cli_with_connection` ? I assume the following would work?

```rust
async fn run_cli(migrator: M) {
    run_cli_with_connection(migrator, Database::connect)
}

**Comment by Lythenas** [comment]
Yes thanks, I didn't think about that.

File: sea-orm-migration/src/cli.rs [hunk 1]
```diff
@@ -13,6 +15,20 @@ const MIGRATION_DIR: &str = "./";
 pub async fn run_cli<M>(migrator: M)
 where
     M: MigratorTrait,
+{
+    run_cli_with_connection(migrator, Database::connect).await;
+}
+
+/// Same as [`run_cli`] but you crate the [`DbConn`] yourself.
```
tyt2y3, line 22:
```suggestion
/// Same as [`run_cli`] where you provide the function to create the [`DbConn`].
```

tyt2y3 (approved):

PR #2400, opened 2024-10-25
Failed attempt to fix `--model-extra-derives` works with `::` in Derive path.

Jack-NimbleTron:
## Bug Fixes

- [ ] When using `sea-orm-cli generate entity --model-extra-derives` with a value such as `async_graphql::SimpleObject` the code no longer generates spaces around the double colon.

## Testing
- [ ] cargo test --manifest-path sea-orm-codegen/Cargo.toml ... failing

### DIFF
--- a/sea-orm-codegen/src/entity/writer.rs
+++ b/sea-orm-codegen/src/entity/writer.rs
@@ -1,9 +1,10 @@
 use crate::{util::escape_rust_keyword, ActiveEnum, Entity};
 use heck::ToUpperCamelCase;
-use proc_macro2::TokenStream;
-use quote::{format_ident, quote};
+use proc_macro2::{Literal, Span, TokenStream, TokenTree};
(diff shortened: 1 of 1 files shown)


noppej:
After a couple of tries, I've come to the conclusion that there is no elegant way to remove the extra spaces around the double colons in the extra derives.

To overcome this, I used a `sed` command in my `build.rs` (see below). It is a bit of a hack, but until someone comes up with something better, it allows my workflow of generating entities to succeed without human intervention.

```
    Command::new("find")
        .arg(&format!("{target_dir}/src"))
        .arg("-name")
        .arg("*.rs")
        .arg("-exec")
        .arg("sed")
        .arg("-i")
        .arg("/^#\\[derive(/s/ *:: */::/g")
        .arg("{}")
        .arg("+")
        .status()
        .expect("Failed to fix Sea ORM entities!");
```

PR #2401, opened 2024-10-27
Add unique getter to ColumnDef

elichai:
This is very useful for building generic getters, especially useful with `.on_conflict(OnConflict::columns(...))` in order to get proper "unique constraint failed" errors

### DIFF
--- a/src/entity/column.rs
+++ b/src/entity/column.rs
@@ -398,6 +398,10 @@ impl ColumnDef {
     pub fn is_null(&self) -> bool {
         self.null
     }
+    /// Returns true if the column is unique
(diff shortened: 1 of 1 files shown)


tyt2y3 (approved):
thank you. of course, not sure why we never added this!

src/entity/column.rs
```diff
@@ -398,6 +398,10 @@ impl ColumnDef {
     pub fn is_null(&self) -> bool {
         self.null
     }
+    /// Returns true if the column is unique
```
tyt2y3, line 401:
```suggestion

    /// Returns true if the column is unique
```

PR #2419, opened 2024-11-14
Use MIGRATION_DIR environment variable

sjunepark:
I'm not sure whether this was intended or not, but the current cli command does not use a environment variable for the `migrate` command.

In my case using a `MIGRATION_DIR` was more easier to use, as I had to pass `-d` for every command I run if I'm not using the default directory.

If this seems unnecessary or misleading, just close the PR thanks.


tyt2y3 (approved):
thank you, it's indeed a nice addition!

Issue #1407, opened 2023-01-20
Panic in `insert_many` if `ActiveModels` have different columns set


GeorgeHahn:
## Description

`insert_many` panics with the message "columns mismatch" if the entities to be stored have values in a mix of columns.

## Steps to Reproduce

See minimal reproduction below

### Expected Behavior

This seems like it should be supported or explicitly disallowed.

### Actual Behavior

Panic internal to the library

### Reproduces How Often

100%

## Versions

Tested with 0.10.7. Reproduces on Windows 10 and Ubuntu 20.04.

## Additional Information

Minimal reproduction:

```rust
// Cargo.toml:
// [package]
// name = "sql-repr"
// version = "0.1.0"
// edition = "2021"
//
// [dependencies]
// sea-orm = { version = "0.10.7", default-features = false, features = [
//     "sqlx-sqlite",
//     "runtime-tokio-rustls",
//     "with-time",
// ] }
// sea-orm-migration = "0.10.7"
// tokio = { version = "1.23.0", features = ["full"] }
//
// [profile.dev]
// panic = "abort"

use migration::Migrator;
use sea_orm::*;
use sea_orm_migration::MigratorTrait;

mod entity {
    use super::*;
    pub mod varies {
        use super::*;

        #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, DeriveEntityModel)]
        #[sea_orm(table_name = "varies")]
        pub struct Model {
            #[sea_orm(primary_key)]
            pub id: i32,
            pub a: Option<i32>,
            pub b: Option<String>,
        }

        impl ActiveModelBehavior for ActiveModel {}

        #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
        pub enum Relation {}
    }
}

mod migration {
    use sea_orm_migration::prelude::*;

    pub struct Migrator;
    #[async_trait::async_trait]
    impl MigratorTrait for Migrator {
        fn migrations() -> Vec<Box<dyn MigrationTrait>> {
            vec![Box::new(Create)]
        }
    }

    #[derive(DeriveMigrationName)]
    struct Create;

    #[async_trait::async_trait]
    impl MigrationTrait for Create {
        async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
            manager
                .create_table(
                    Table::create()
                        .table(Varies::Table)
                        .if_not_exists()
                        .col(
                            ColumnDef::new(Varies::Id)
                                .integer()
[...]

PR #2433, opened 2024-12-01
Insert many allow active models to have different column set

tyt2y3:
Fixes https://github.com/SeaQL/sea-orm/issues/1407

This requires a new feature in sea-query, [Value::as_null](https://docs.rs/sea-query/latest/sea_query/value/enum.Value.html#method.as_null)

### DIFF
--- a/src/query/insert.rs
+++ b/src/query/insert.rs
@@ -175,18 +175,19 @@ where
                 let av = am.take(col);
                 match av {
                     ActiveValue::Set(value) | ActiveValue::Unchanged(value) => {
-                        columns[idx] = Some(col);
-                        null_value[idx] = Some(value.as_null());
-                        values.push(col.save_as(Expr::val(value)));
+                        columns[idx] = Some(col); // mark the column as used
+                        null_value[idx] = Some(value.as_null()); // store the null value with the correct type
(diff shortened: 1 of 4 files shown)


tyt2y3:
I believe this should have solved the problem once and for all. Since this is changing a core part of the library, we should test it more to prevent regression.

The safest way might be to have a new method `insert_many_nopanic`, but of course ideally we should replace the old method transparently.

src/query/insert.rs
```diff
@@ -155,9 +155,67 @@ where
...
         }
+
+        if !all_values.is_empty() {
+            // filter only used column
+            self.query.columns(columns.iter().cloned().flatten());
+
+            // flag used column
+            for col in columns.iter() {
+                self.columns.push(col.is_some());
+            }
```
tyt2y3, line 197: thanks. self.columns is really for use with add(). so should we do `self.columns.clear()` first?

billy1624: Okay, reset (clear) it before altering it seems safer.

billy1624: Or simply replace the columns vector, isn't this more readable?

```rs
self.columns = columns.iter().map(Option::is_some).collect();
```

PR #2459, opened 2024-12-24
Fix TryInsert as well

tyt2y3:
Cont'd https://github.com/SeaQL/sea-orm/pull/2433

I realised `TryInsert::add_many` doesn't actually delegate to `Insert::add_many`


tyt2y3:
🤦

Issue #2458, opened 2024-12-20
macro DeriveActiveEnum is conflict with crate utoipa's macro ToSchema


799189288:
https://github.com/juhaku/utoipa/issues/1156

tyt2y3:
PR is welcomed if you want to give it a try

PR #2460, opened 2024-12-24
Fix: use fully qualified syntax when call fn name in trait ActiveEnum.

799189288:
## PR Info

https://github.com/SeaQL/sea-orm/issues/2458


tyt2y3:
I think it will be good, thanks!

Issue #555, opened 2022-02-23
Failed to generate entities for large schema


milesgranger:
## Description

When testing out the project, specifically generating entities for a schema we use which has nearly 1,000 tables, the following results:

```
$ sea-orm-cli generate entity -u $CONNECTION_URI -s big-schema -o src/models/replication/big-schema --compact-format
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PoolTimedOut', /home/milesg/.cargo/registry/src/github.com-1ecc6299db9ec823/sea-schema-0.5.1/src/postgres/discovery/executor/real.rs:30:55
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Steps to Reproduce

Difficult, but generating a lot of tables, which may have hundreds of columns can reproduce this error.

### Expected Behavior

Same behavior (works) on small schemas with ~20 tables.

### Actual Behavior

`PoolTimedOut`

### Reproduces How Often

Every time.

## Versions

```
└── sea-orm v0.6.0
    ├── sea-orm-macros v0.6.0 (proc-macro)
    ├── sea-query v0.21.0
    │   ├── sea-query-derive v0.2.0 (proc-macro)
    ├── sea-strum v0.23.0
    │   └── sea-strum_macros v0.23.0 (proc-macro)
```

OS: Ubuntu 18
Database: PostgreSQL 12.7

billy1624:
Hey @milesgranger, thanks for the report! Actually, we haven't try to run `sea-orm-cli generate entity` on a huge database. I think there are some bottleneck when querying the metadata from the information schemas. We would try to reproduce the errors with a large database then identify the bottleneck.

tyt2y3:
The cause of failure may or may not be "large". It could be because of one table using an unsupported feature. So I think at least a minimum-reproducible schema be provided for this to be actionable. Please reopen if you have any updates

milesgranger:
I think at least, a reasonable error should be produced if a table is using an unsupported feature, would you agree?

tyt2y3:
Indeed

PR #2461, opened 2024-12-24
Resolve connect timeout in schema discovery

ctaque:
## PR Info

This PR resolves the timeout in schema discovery in larger databases when generating entities with `sea-orm-cli`

- Closes https://github.com/SeaQL/sea-orm/issues/555

## Bug Fixes

- [ ] This PR increases the acquire timeout of the connection of the schema discovery in Postgresql from 30s to 240s.

The discovery of my database now work with this value. IMO it could be parametrized in the cli.

## Changes

- [ ] The connection is now an sqlx pool with an `acquire_timeout` of 240s

### DIFF
--- a/sea-orm-cli/src/cli.rs
+++ b/sea-orm-cli/src/cli.rs
@@ -203,6 +203,13 @@ pub enum GenerateSubcommands {
         max_connections: u32,
 
         #[arg(
+            long,
+            default_value = "30",
+            long_help = "Acquire timeout in seconds of the connection used for schema discovery"
+        )]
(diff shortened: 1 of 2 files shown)


tyt2y3:
thanks, I agree with that it could be parametrized in the cli.
would you be able to work on it?

ctaque:
I added this commit 006fc1e7c6bba4d57f8683d319b7a1367f40ff72 to add the option to the cli and use it in Mysql, Postgresql and Sqlite

tyt2y3 (approved):
Thank you

PR #2466, opened 2025-01-04
Remove futures dependency

jplatte:
## PR Info

Reduces dependencies and compile time by removing `futures` as a dependency.

## Changes

- [ ] Use `futures-util` directly rather than through `futures` re-exports


tyt2y3 (approved):
Thank you!

PR #2471, opened 2025-01-09
Respect on_condition when batch loading relations

georgeboot:
## Bug Fixes

- [ ] When using batch loading (`LoaderTrait`), the `on_condition` of a relation was not applied.


tyt2y3 (commented):
Thank you! This seems to be a valid fix, can you also add a test case, so we'd not break it in the future?

tyt2y3:
ah sorry so master is currently undergoing a number of breaking changes. you can target 1.1.x so there should be no conflicts.
and can you add a test case to illustrate what we're trying to fix?

georgeboot:
Sure

PR #2475, opened 2025-01-15
fix: seaography::register_active_enums macro

ouertani:
## PR Info

fixes: the macro enum generation for seaography otherwise the generated code will not compile for multiple enums

## Bug Fixes

- [x] fixes mod.rs for an array of enum


billy1624 (commented):
Hey @ouertani, thanks for the fix! I will add a few test cases for this.

tyt2y3:
Thank you all! I will release this soon

PR #2480, opened 2025-01-24
Add find_linked_recursive method to ModelTrait

rumpuslabs:
## PR Info

- Dependencies:
  -  SeaQL/sea-query#859

## New Features

- [ ] if a `Linked<From = M, To = M>` exists, then `model.find_linked_recursive` will generate a CTE to follow the links recursively and return all ancestors or descendants of the model.

## Bug Fixes

- [ ] `RelationDef.rev()` did not swap the arguments to its `on_condition`

## Changes

- [ ] I added `#[ignore]` to `join_16` and `join_18` tests in `src/query/join.rs` because the generated queries make no sense, I suggest removing them (and `entity_linked::JoinWithoutReverse`)


src/query/join.rs
```diff
@@ -245,8 +207,7 @@ mod tests {
                 .to_string(),
             [
                 "SELECT `fruit`.`id`, `fruit`.`name`, `fruit`.`cake_id` FROM `fruit`",
-                "INNER JOIN `cake` ON `cake`.`id` = `fruit`.`cake_id`",
-                "WHERE `cake`.`id` = 12",
```
tyt2y3, line 249: Oh wow this is such a smart optimization

src/tests_cfg/entity_linked.rs
```diff
@@ -46,8 +46,8 @@ impl Linked for CheeseCakeToFillingVendor {
         vec![
             super::cake_filling::Relation::Cake
                 .def()
-                .on_condition(|left, _right| {
-                    Expr::col((left, super::cake::Column::Name))
+                .on_condition(|_cake_filling, cake| {
+                    Expr::col((cake, super::cake::Column::Name))
```
tyt2y3, line 50: why is the left and right swapped?

tyt2y3:
> RelationDef.rev() did not swap the arguments to its on_condition

I understand your point that this is a bug, but this will be a breaking change to fix this at this point, sadly.

Can you add some test cases to demonstrate the recursive use cases?

rumpuslabs:
Removed breaking change to `RelationDef.rev()` and added recursive unit test cases (`CakeToCakeViaFilling`).

tyt2y3:
https://github.com/SeaQL/sea-orm/actions/runs/19209449625/job/54909548118
`RelationDef should refer to valid column name`
@rumpuslabs I've encountered a problem in which I don't understand
is this an assumption being violated? I think the problem stems from `RelationDef.rev()`

rumpuslabs:
The `entity_linked::JoinWithoutReverse` relation used in `test_16` and `test_18` is actually invalid since it is missing the call to `.rev()` on `super::cake_filling::Relation::Cake` so the link is the wrong way around. I need to convert from an `Identity` to a `<Self::Entity as EntityTrait>::Column` and this conversion fails if the column named doesn't exist in the table. You could change those tests to `should_panic` perhaps?

tyt2y3:
@rumpuslabs thank you for your explanation. yes I can confirm that those test cases are wrong, I've removed them now.
however I think the "one less join" optimzation is affecting too many places, and it's more complex for relations with `via` defined. I've rebased the linked_recursive features in 1a21ae28186e5142044f42323e06762338d19c79 and I will cherry-pick this to master too.

thank you again for your contribution

PR #2485, opened 2025-01-31
Fix UUID parsing issue due to hyphenated format in MariaDB

rito528:
## PR Info
This pull request resolves an issue where Uuid values retrieved from MariaDB (I am using 10.11.10) were not correctly parsed, leading to runtime errors such as:

```
Query Error: error occurred while decoding column "column": invalid length: expected 16 bytes, found 36
```

The problem occurred because [MariaDB's UUID type](https://mariadb.com/kb/en/uuid-data-type/) stores UUIDs as hyphenated strings (e.g., `f8aa-ed66-1a1b-11ec-ab4e-f859-713e-4be4`), while `uuid::Uuid` in Rust expects a non-hyphenated byte format. The current implementation attempted to directly convert the database value into `Uuid`, which caused errors due to this mismatch in format and byte length. The behavior on other database systems or versions has not been tested.

## Changes
- Modified the handling of UUID values retrieved from MariaDB to account for the fact that MariaDB stores UUIDs as hyphenated strings.
- The `try_get` method now retrieves the UUID value as an `Option<Vec<u8>>`, because we cannot determine whether the database is MariaDB or MySQL and need a more general approach.
- If the UUID cannot be directly parsed, the value is first converted to a string and then parsed using `uuid::Uuid::parse_str`, ensuring correct handling of the hyphenated UUID format.

This change ensures that UUID values retrieved from MariaDB are handled correctly, preventing runtime errors related to the incorrect length or format, while maintaining compatibility with the database's UUID storage format.

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -627,14 +627,12 @@ macro_rules! try_getable_uuid {
                         })?,
                     #[cfg(feature = "sqlx-postgres")]
                     QueryResultRow::SqlxPostgres(row) => row
-                        .try_get::<Option<String>, _>(idx.as_sqlx_postgres_index())
-                        .map(|opt| opt.and_then(|s| uuid::Uuid::parse_str(&s).ok()))
+                        .try_get::<Option<uuid::Uuid>, _>(idx.as_sqlx_postgres_index())
                         .map_err(|e| sqlx_error_to_query_err(e).into())
                         .and_then(|opt| opt.ok_or_else(|| err_null_idx_col(idx))),
                     #[cfg(feature = "sqlx-sqlite")]
                     QueryResultRow::SqlxSqlite(row) => row
-                        .try_get::<Option<String>, _>(idx.as_sqlx_sqlite_index())
-                        .map(|opt| opt.and_then(|s| uuid::Uuid::parse_str(&s).ok()))
+                        .try_get::<Option<uuid::Uuid>, _>(idx.as_sqlx_postgres_index())
                         .map_err(|e| sqlx_error_to_query_err(e).into())
                         .and_then(|opt| opt.ok_or_else(|| err_null_idx_col(idx))),
                     #[cfg(feature = "mock")]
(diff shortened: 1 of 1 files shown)


rito528:
Ready for review!

tyt2y3:
thank you for your investigation. I think there must be a better solution.
I think we might be able to do:

1. try `Uuid` as usual
2. if that fails, try [Hyphenated](https://docs.rs/uuid/latest/uuid/fmt/struct.Hyphenated.html)
3. if succeeded, convert back to `Uuid`
4. if failed, return error

may be worth a try?

rito528:
@tyt2y3 

If the proposed approach were feasible, I believe it would make the intent clearer. However, there is no direct way to convert to `uuid::fmt::Hyphenated` without first going through `uuid::Uuid`.  

I tried modifying the code as follows, but encountered a decoding error:  
`Query Error: error occurred while decoding column "column": mismatched types; Rust type core::option::Option<uuid::fmt::Hyphenated> (as SQL type VARCHAR) is not compatible with SQL type BINARY`  

```rust
QueryResultRow::SqlxMySql(row) => {
    let uuid_parse_result = row
        .try_get::<Option<uuid::Uuid>, _>(idx.as_sqlx_mysql_index())
        .map_err(|e| sqlx_error_to_query_err(e).into())
        .and_then(|opt| opt.ok_or_else(|| err_null_idx_col(idx)));

    match uuid_parse_result {
        Ok(uuid) => Ok(uuid),
        Err(e) => row
            .try_get::<Option<uuid::fmt::Hyphenated>, _>(idx.as_sqlx_mysql_index())
            .map_err(|e| sqlx_error_to_query_err(e).into())
            .and_then(|opt| opt.ok_or_else(|| err_null_idx_col(idx)))
            .map(uuid::fmt::Hyphenated::into_uuid)
    }
}
```

Given this, I believe the only feasible approach is the one currently implemented in the pull request. Does this approach seem reasonable to you?

tyt2y3:
I see. So we’re basically blocked on https://github.com/launchbadge/sqlx/issues/3039

We should think of a workaround with less overhead.
________________________________

tyt2y3:
I see. seems viable. can you update the PR?

rito528:
@tyt2y3 
I updated it. Ready for review!

tyt2y3 (approved):
thank you

PR #2487, opened 2025-02-03
feat: implement a skip generate ActiveModelBehaviour

TheSamsa:
sea-orm-cli generate has a new option where generation of ActiveModelBehaviour can be skipped to not overwrite custom logic

This is a duplicate of: #1947
I have only copied this work, to enable this feature. I merely copied it, since there was no progress since october 2024.

## PR Info

- Closes N/A

## New Features

- [ ] introduces a `--gen-impl-active-model-behavior` flag to disable generation


tyt2y3:
I would remove the `gen_` prefix though

PR #2503, opened 2025-02-21
Add trait bound for binary PostgreSQL array

pantsforcats:
## PR Info

This adds a missing trait bound that allows `Vec<u8>` to be stored in a PostgreSQL array.

## Bug Fixes

The CLI will currently happily generate an entity that contains a `Vec<Vec<u8>>` when encountering a `bytea[]` column in PostgreSQL. This entity will however fail to compile, as the `TryGetable` trait bound is missing for this type.

This PR adds the missing trait bound.


jippeholwerda:
Anything that can be done to help move this forward?

tyt2y3:
not much really, please come back if there are issues

PR #2509, opened 2025-02-23
impl_active_model_behavior's default value must be false

zer0tim3r:
impl_active_model_behavior's default value must be false, becuase it cannot change to flase on cli


tyt2y3:
Well, I just jumped down the rabbit hole

https://github.com/clap-rs/clap/issues/1649

tyt2y3:
Not sure how, but it works 1052f366db55717f23feb2e16be027175d1cfee9 !

PR #2511, opened 2025-02-24
Close connection by reference

### LINKED ISSUE #2510: Cannot close `DatabaseConnection` by reference
## Description

Hi, so I use a `DatabaseConnection` in my service that is shared across different tasks by reference (`Arc`, in particular). I've recently  encountered a problem similar to this (https://discord.com/channels/873880840487206962/1059720897939918878) and I need to manually `close` the connection.

However, I can't just easily do this because `close` requires an owned value. Also, I've noticed that [all branches](https://docs.rs/sea-orm/latest/src/sea_orm/database/db_connection.rs.html#479-499) of `close` actually just need a reference (if traced to `sqlx`). So, I think it makes sense to update the implementation to `&self` instead, considering that other 'write' operations (to the db) are performed by reference and this is the way sqlx does it.

## Steps to Reproduce
```
Arc::new(connection).close().await;
```

### Workarounds

I know that under the hood the sqlx pools have `Arc`, so it makes sense to try and use owned value (with `clone`ing as needed). However, it all breaks once `mock` feature is enabled. I don't want to restrict the usage of `mock`.

## Versions

latest

bragov4ik:
## PR Info

- Closes #2510 

## Bug Fixes

- [x] Allow to close connection by reference

## Changes

- [x] Add `close_by_ref` methods to close connections with references

### DIFF
--- a/src/database/db_connection.rs
+++ b/src/database/db_connection.rs
@@ -476,7 +476,7 @@ impl DatabaseConnection {
     }
 
     /// Explicitly close the database connection
-    pub async fn close(self) -> Result<(), DbErr> {
+    pub async fn close(&self) -> Result<(), DbErr> {
         match self {
             #[cfg(feature = "sqlx-mysql")]
             DatabaseConnection::SqlxMySqlPoolConnection(conn) => conn.close().await,
--- a/src/database/db_connection.rs
+++ b/src/database/db_connection.rs
@@ -475,15 +475,21 @@ impl DatabaseConnection {
         }
     }
 
+    /// Explicitly close the database connection.
+    /// See [`Self::close_by_ref`] for usage with references.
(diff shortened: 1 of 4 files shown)


tyt2y3:
This is a breaking change. But we can add a new set of methods called `close_by_ref`

spencewenski:
Hmm, yeah, I agree this is probably technically a breaking change. However, this would only break when using the fully qualified syntax, e.g. `DatabaseConnection::close(connection)`. This case also isn't specifically mentioned in https://doc.rust-lang.org/cargo/reference/semver.html. So, it may be okay to introduce this change IMO, but that's just my two cents and I'll leave it up to the maintainers to decide.

tyt2y3 (approved):

PR #2524, opened 2025-03-10
Replace `once_cell` crate with `std` equivalent

FalkWoldmann:
## Changes

- [x] Replace `once_cell` crate used in tests with `std` equivalent


tyt2y3 (approved):
thank you!

PR #2549, opened 2025-03-30
Allow to specify an escape character in `like` and `not_like`

Expurple:
## PR Info

- Closes #2546
- Dependencies: none
- Dependents: none

## Changes

- Allow to specify an escape character in `like` and `not_like`.

The old signatures were insufficiently generic and didn't accept `LikeExpr`. See the linked discussion.

Changing the parameter type from `T: Into<String>` to `T: IntoLikeExpr` is a backwards-compatible, because the former is already generic and is a subtype of the latter ([`IntoLikeExpr` is implemented for  `T: Into<String>`](https://docs.rs/sea-query/0.32.3/sea_query/types/trait.IntoLikeExpr.html#impl-IntoLikeExpr-for-T)).

It's also consistent with the corresponding [`ExprTrait` methods](https://docs.rs/sea-query/0.32.3/sea_query/expr/trait.ExprTrait.html#method.like).

### DIFF
--- a/src/entity/column.rs
+++ b/src/entity/column.rs
@@ -1,6 +1,7 @@
 use crate::{DbBackend, EntityName, Iden, IdenStatic, IntoSimpleExpr, Iterable};
 use sea_query::{
-    Alias, BinOper, DynIden, Expr, IntoIden, SeaRc, SelectStatement, SimpleExpr, Value,
+    Alias, BinOper, DynIden, Expr, IntoIden, IntoLikeExpr, SeaRc, SelectStatement, SimpleExpr,
+    Value,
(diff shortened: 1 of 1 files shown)


Expurple:
The CI failures seem unrelated to my change:
```txt
cannot find type `IpNetwork` in this scope
```

Expurple:
Rebased to pick up the [CI fix](https://github.com/SeaQL/sea-orm/commit/a9eb30d66704235b1092d72857fc127a2ca8f699) and trigger a re-run

tyt2y3 (approved):
thanks!

PR #2552, opened 2025-04-01
fix: use fully-qualified syntax for ActiveEnum associated type

omjadas:
## PR Info

Currently if you try and derive `ActiveEnum` on an enum with a variant named `Value` you run into [this](https://doc.rust-lang.org/rustc/lints/listing/deny-by-default.html#ambiguous-associated-items) deny by default lint.

## Bug Fixes

- [x] use fully-qualified syntax for ActiveEnum associated type


tyt2y3 (approved):
Nice catch, thank you

Issue #2063, opened 2024-01-14
Entity generation fails if `model-extra-attributes` include a comma surrounded by parentheses


naskya:
## Description

We can't seem to pass a string containing a comma surrounded by parentheses as `model-extra-attributes` in sea-orm-cli. This becomes a problem when you want to pass `cfg_attr(predicate, attribute)`, for instance.

## Steps to Reproduce

```sh
$ sea-orm-cli generate entity --model-extra-attributes 'test(a, b)' -u 'postgres://username:password@localhost/dbname'
```

### Expected Behavior

Since the following does work and produces `Model`s with `#[test(a)]`, I guess `test(a, b)` should also work.

```sh
$ sea-orm-cli generate entity --model-extra-attributes 'test(a)' -u 'postgres://username:password@localhost/dbname'
```

### Actual Behavior

```sh
$ RUST_BACKTRACE=1 sea-orm-cli generate entity --model-extra-attributes 'test(a, b)' -u 'postgres://username:password@localhost/dbname'

Connecting to Postgres ...
Discovering schema ...
... discovered.
thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sea-orm-codegen-0.12.10/src/entity/writer.rs:108:57:
called `Result::unwrap()` on an `Err` value: LexError { span: Span }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/result.rs:1652:5
   3: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   4: sea_orm_codegen::entity::writer::EntityWriterContext::new
   5: sea_orm_cli::commands::generate::run_generate_command::{{closure}}
   6: async_std::task::task_locals_wrapper::TaskLocalsWrapper::set_current
   7: <futures_lite::future::Or<F1,F2> as core::future::future::Future>::poll
   8: async_io::driver::block_on
   9: async_global_executor::executor::block_on
  10: async_std::task::task_locals_wrapper::TaskLocalsWrapper::set_current
[...]

PR #2560, opened 2025-04-13
[FIX] codegen/entity: Support multi argument extra attribute

macadelic80:
## PR Info

- Closes 
https://github.com/SeaQL/sea-orm/issues/2063

## Bug Fixes

 Before this commit, using `--*-extra-attributes` with an attribute containing a comma (several arguments) did not work (example `--model-extra-attributes=“graphql(complex, name=”users“`)

## Changes

- [ ] Add support of multi-argument attributes by ensuring proper parsing of comma-separated values in the TokenStream

### DIFF
--- a/sea-orm-codegen/src/entity/active_enum.rs
+++ b/sea-orm-codegen/src/entity/active_enum.rs
@@ -184,7 +184,7 @@ mod tests {
                 &WithSerde::None,
                 true,
                 &TokenStream::new(),
-                &bonus_attributes([r#"serde(rename_all = "camelCase")"#])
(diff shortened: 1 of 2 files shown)


Huliiiiii:
I have to revert this because the test input is incorrect. The test input should be split by `,`.

Example:
```
[
    r#"serde(rename_all = "camelCase""#,
    r#" rename = "test")"#,
    r#"ts(export"#,
    r#" export_to = "path")"#
]
```

And in this case syn will panic because it can’t parse it into a valid TokenStream.

Issue #2561, opened 2025-04-14
Sqlx0.8.4 breaks sea-orm


satyamjay-iitd:
## Description
sqlx-0.8.4 breaks the compilation. We get this error:-
```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
  --> /home/satyam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sea-orm-1.1.9/src/database/transaction.rs:57:17
   |
57 |                 <sqlx::Postgres as sqlx::Database>::TransactionManager::begin(c)
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^--- argument #2 of type `std::option::Option<Cow<'static, str>>` is missing
   |
note: associated function defined here
  --> /home/satyam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sqlx-core-0.8.4/src/transaction.rs:25:8
   |
25 |     fn begin<'conn>(
   |        ^^^^^
help: provide the argument
   |
57 |                 <sqlx::Postgres as sqlx::Database>::TransactionManager::begin(c, /* std::option::Option<Cow<'static, str>> */)
   |                                  
```
They have changed their `TransactionManager::begin` interface in 0.8.4 release.

## Steps to Reproduce

1. `cargo init`
2. Add `sea-orm = { version = "1.1.9", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros" ] }` in the dependency
3. `cargo build`

### Expected Behavior
Successful Compilation

### Actual Behavior
Compilation Error

### Reproduces How Often
Always

### Workarounds
Currently the workaround is to pin the sqlx-core version to '0.8.3' in the user's project.

antoinecharbonneau:
Same issue with MySQL dependency

abonander:
I highly recommend auditing for use of other SemVer-exempt APIs. 

[The `sqlx-core` crate itself is SemVer-exempt](https://docs.rs/sqlx-core/latest/sqlx_core/#note-semver-exempt-api). This applies to all consistent crates except the `sqlx` crate itself.

antoinecharbonneau:
> I highly recommend auditing for use of other SemVer-exempt APIs.
> 
[...]

PR #2562, opened 2025-04-14
Hotfixed compatibility with sqlx 0.8.4 TransactionManager interface

antoinecharbonneau:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2561

## Bug Fixes

- [x] Fixes issue where library wouldn't compile due to sqlx 0.8.4 change of interface for TransactionManager interface

## Breaking Changes

Should be all backwards compatible as it doesn't add any code ran to the transaction as it used to. Works with all tests.

## Changes

Added None option to TransactionManager begin function, implying no query should be ran if a transaction is initiated.

### DIFF
--- a/src/database/transaction.rs
+++ b/src/database/transaction.rs
@@ -48,13 +48,13 @@ impl DatabaseTransaction {
                 // in MySQL SET TRANSACTION operations must be executed before transaction start
                 crate::driver::sqlx_mysql::set_transaction_config(c, isolation_level, access_mode)
                     .await?;
-                <sqlx::MySql as sqlx::Database>::TransactionManager::begin(c)
+                <sqlx::MySql as sqlx::Database>::TransactionManager::begin(c, None)
                     .await
                     .map_err(sqlx_error_to_query_err)
             }
             #[cfg(feature = "sqlx-postgres")]
             InnerConnection::Postgres(ref mut c) => {
-                <sqlx::Postgres as sqlx::Database>::TransactionManager::begin(c)
+                <sqlx::Postgres as sqlx::Database>::TransactionManager::begin(c, None)
                     .await
                     .map_err(sqlx_error_to_query_err)?;
                 // in PostgreSQL SET TRANSACTION operations must be executed inside transaction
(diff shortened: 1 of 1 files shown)


antoinecharbonneau:
This would require hotfixing latest release of sea-orm

tyt2y3:
thank you for submitting this quickly!

tyt2y3:
Let me release this quick

abonander:
`TransactionManager` is [a `#[doc(hidden)]` trait](https://docs.rs/sqlx-core/latest/src/sqlx_core/transaction.rs.html#14-15) in an [explicitly SemVer-exempt crate](https://docs.rs/sqlx-core/latest/sqlx_core/#note-semver-exempt-api). This is a temporary fix at best.

abonander:
I just noticed that it's actually exported from `sqlx`. This export will be deleted in 0.9.0 as it's not meant to be used directly.

abonander:
I see what's going on: you're using your own RAII wrapper that's not compatible with `sqlx::Transaction`. I can see the motivation for that.

The issue is that `TransactionManager` is an implementation detail that needs to remain flexible so we can _do_ things like add [`begin_with`](https://docs.rs/sqlx/latest/sqlx/trait.Connection.html#method.begin_with) backwards-compatibly. That's _why_ it's a `#[doc(hidden)]` trait.

We need to figure out a long-term solution here:

* You could use the trait from `sqlx-core` [with a pinned version](https://docs.rs/sqlx-core/latest/sqlx_core/#note-semver-exempt-api), but then you have to cut a release every time SQLx does.
* You could implement your own transaction handling and execute `BEGIN`, `COMMIT` and `ROLLBACK` statements directly. This offers SeaORM the most potential flexibility.
* We could design a non-RAII transaction begin/end API for SQLx. My main concern with this is that the RAII-based API _should_ still be the preferred method for 99% of use-cases, so the design should make that obvious.

However, perhaps in general we need a better strategy for distinguishing whether SemVer applies only to _usages_ of a trait, and not implementations. Then `TransactionManager` _could_ just be a public trait. It should probably just be a supertrait of `Connection`, honestly.

PR #2566, opened 2025-04-15
Support `migration_name/mod.rs` migration layout

Huliiiiii:
## PR Info
Currently, migration names are generated by the derive macro `DeriveMigrationName`, which obtains the file name of the current migration file. However, this approach does not support a directory-based layout like:
```
- migration_name
  - up.sql
  - down.sql
  - mod.rs
```
In this PR, I modified the `get_file_stem` function to return the parent directory name when the current file is mod.rs. This allows migration names to be correctly resolved even when using this layout.

## New Features

- [x] Support `migration_name/mod.rs` migration layout

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/sea-orm-migration/src/util.rs
+++ b/sea-orm-migration/src/util.rs
@@ -1,8 +1,15 @@
 pub fn get_file_stem(path: &str) -> &str {
-    std::path::Path::new(path)
-        .file_stem()
-        .map(|f| f.to_str().unwrap())
-        .unwrap()
+    let path = std::path::Path::new(path);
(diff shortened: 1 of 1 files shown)


Huliiiiii:
To be honest, I don't think this is a good solution, shouldn't we specify names when generating migrations that don't depend on the file location?

Stefan99353:
I am currently using this as a workaroung:

```rust
impl MigrationName for Migration {
    fn name(&self) -> &str {
        module_path!()
            .split("::")
            .last()
            .expect("Module path must have at least one element")
    }
}
```

PR #2580, opened 2025-04-23
cli: Implement `MigrationName` for Migration in migration generation

Huliiiiii:
## PR Info

A better solution of #2566.
Make migration names independent of file name to support different file structures.

I also cleaned up the migration template.

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [x] Migration names are no longer dependent on file names, and are now generated when generating migrations.

### DIFF
--- a/sea-orm-cli/src/commands/migrate.rs
+++ b/sea-orm-cli/src/commands/migrate.rs
@@ -179,12 +179,42 @@ fn create_new_migration(migration_name: &str, migration_dir: &str) -> Result<(),
     let migration_filepath =
         get_full_migration_dir(migration_dir).join(format!("{}.rs", &migration_name));
     println!("Creating migration file `{}`", migration_filepath.display());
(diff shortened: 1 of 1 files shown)


sea-orm-cli/src/commands/migrate.rs
```diff
@@ -179,14 +179,41 @@ fn create_new_migration(migration_name: &str, migration_dir: &str) -> Result<(),
...
 
+fn fmt_migration_tempelate(migration_name: &str) -> String {
+    format! {
+        r#"use sea_orm_migration::{{prelude::*, schema::*}};
+
+pub struct Migration;
+
+impl MigrationName for Migration {{
+    fn name(&self) -> &str {{
+        "{migration_name}"
```
Huliiiiii, line 197: Should we check for invalid input?

PR #2583, opened 2025-04-27
codegen: Generate sqlite integer as i64

### LINKED ISSUE #2582: Sqlite integer is mapped to i32
## Description
When using sea-orm-cli to generate entities from the Database, and the target is Sqlite, `INTEGER` columns, including `ROWID` aliases, are mapped as `i32`. As mentioned in #1067, Sqlite does not differentiate the `INTEGER` storage class, and uses up to a 8 byte signed integer for backing such fields.

`sea-orm-cli`'s decision to map these as `i32` renders most of the entire value space incorrect, including a huge amount where handling becomes so wrong it's likely to cause data corruption of the "correctly" handled values in the positive `i32` space (for example `4294967297` is stripped to the lower 4 bytes and is read as `1`)

## Steps to Reproduce

1. Create integer column (as `integer().primary_key()` if Sqlite is the only DB target, otherwise `pk_auto()`)
2. Run `sea-orm-cli generate entity` to generate entities from Database
3. Insert some data
4. Observe that all values above `2_147_483_647` in the database are read wrong
5. Observe all values above `4_294_967_296` in the database are irrecoverably wrong and likely to cause fundamental data corruption (particularly on `ROWID` aliases and foreign keys)
6. Manually inserting a row with a `ROWID` of `4294967297` as well as `1` will demonstrate that this fundamentally breaks ability to set foreign keys or to correctly reference rows

### Expected Behavior

The Integer type is correctly mapped for the data which it may contain: `i64`.

### Actual Behavior

Integer type is incorrectly mapped as `i32`

### Reproduces How Often

Always

### Workarounds

Only available workarounds to get correct data handling is to manually change all `i32` columns, especially `ROWID` aliases, to `i64` when Sqlite is in use/available as a DB backend.
On non-`ROWID` columns, where no `AUTOINCREMENT`-like behaviour is required, changing the type definition to `big_integer()` is also viable

## Versions

│   ├── sea-orm v1.1.10
│   │   ├── sea-orm-macros v1.1.10 (proc-macro)
│   │   │   ├── sea-bae v0.2.1 (proc-macro)
│   │   ├── sea-query v0.32.4
│   │   │   ├── sea-query-derive v0.4.3 (proc-macro)
│   │   ├── sea-query-binder v0.7.0
│   │   │   ├── sea-query v0.32.4 (*)
│   ├── sea-orm-migration v1.1.10
│   │   ├── sea-orm v1.1.10 (*)
│   │   ├── sea-orm-cli v1.1.10
│   │   │   ├── sea-schema v0.16.1
│   │   │   │   ├── sea-query v0.32.4 (*)
│   │   │   │   └── sea-schema-derive v0.3.0 (proc-macro)
│   │   ├── sea-schema v0.16.1 (*)
├── sea-orm v1.1.10 (*)

DB: Sqlite
OS: Linux

Huliiiiii:
## PR Info

- Closes #2582  

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [x] If database backend is sqlite, interger will generate as i64 instead of i32

### DIFF
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -1,7 +1,7 @@
 use core::time;
 use sea_orm_codegen::{
-    DateTimeCrate as CodegenDateTimeCrate, EntityTransformer, EntityWriterContext, OutputFile,
-    WithPrelude, WithSerde,
+    DatabaseBackend, DateTimeCrate as CodegenDateTimeCrate, EntityTransformer, EntityWriterContext,
+    OutputFile, WithPrelude, WithSerde,
 };
 use std::{error::Error, fs, io::Write, path::Path, process::Command, str::FromStr};
 use tracing_subscriber::{prelude::*, EnvFilter};
@@ -111,7 +111,7 @@ pub async fn run_generate_command(
                 Default::default()
             };
 
-            let (schema_name, table_stmts) = match url.scheme() {
+            let (schema_name, table_stmts, backend) = match url.scheme() {
                 "mysql" => {
                     use sea_schema::mysql::discovery::SchemaDiscovery;
                     use sqlx::MySql;
@@ -132,7 +132,7 @@ pub async fn run_generate_command(
                         .filter(|schema| filter_skip_tables(&schema.info.name))
                         .map(|schema| schema.write())
                         .collect();
-                    (None, table_stmts)
+                    (None, table_stmts, DatabaseBackend::MySql)
                 }
                 "sqlite" => {
                     use sea_schema::sqlite::discovery::SchemaDiscovery;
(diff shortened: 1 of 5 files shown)


tyt2y3:
thank you. I'd like to have an option to choose whether to map it to i32/i64.
another thing we can do (in sea-schema) is to recognize `bigint` as a special type, and return BigInt, similar to how SmallInt is handled now.

tyt2y3:
thank you. I'd like to have an option to choose whether to map it to i32/i64.
another thing we can do (in sea-schema) is to recognize `bigint` as a special type, and return BigInt, similar to how SmallInt is handled now.

tyt2y3:
I've added option to sea-orm-cli here dc8fe659d9a6ab0b066fc57ebffe09e4db21c98a

I actually like the pk fix in sea-schema better!

PR #2587, opened 2025-04-30
`impl IntoCondition for RelationDef`

Expurple:
## PR Info

- Closes: none
- Dependencies: none
- Dependents: none

## New Features

- `impl IntoCondition for RelationDef`.

    This allows using `RelationDef` directly, wherever `sea_query` expects an `IntoCondition` type:
    
    ```rust
    let query = Query::select()
        .from(fruit::Entity)
        .inner_join(cake::Entity, fruit::Relation::Cake.def())
        .to_owned();
    ```

    As well as generating the `JOIN` condition explicitly:

    ```rust
    fruit::Relation::Cake.def().into_condition()
    ```

I was very surprised when I couldn't do this. Generating `JOIN` conditions from `RelationDef` is obviously already implemented. The existing implementation just wasn't public and wan't using this trait.

## Changes


tyt2y3 (approved):
thank you

Issue #2528, opened 2025-03-14
on_empty_do_nothing() should be enabled by default


zevweiss:
## Description

The intuitive, obvious way to do a multi-record insert (`foo::Entity::insert_many(iter).exec(...)`) provides highly surprising behavior if the iterator passed to it is empty (inserting a single `DEFAULT` record instead of inserting nothing).

https://github.com/SeaQL/sea-orm/pull/1708 added `.on_empty_do_nothing()` to work around that problem and enable the more predictable behavior of an empty iterator resulting in nothing being inserted, but it seems quite regrettable (and highly counterintuitive IMO) for that to be opt-in instead of the default; without it most `insert_many()` calls likely end up as time bombs just waiting for an empty iterator to cause havoc.

I suppose it would technically be a backwards-incompatible change to make it the default and hence unfit for a minor or patch release, but it's too bad the 0.x->1.0 release didn't take the opportunity to make that change.  Could it please be added to the agenda for a 2.0 release?

## Versions

```
$ cargo tree | grep sea-
├── sea-orm v1.1.6
│   ├── sea-orm-macros v1.1.6 (proc-macro)
│   │   ├── sea-bae v0.2.1 (proc-macro)
│   ├── sea-query v0.32.2
│   │   ├── sea-query-derive v0.4.2 (proc-macro)
│   ├── sea-query-binder v0.7.0
│   │   ├── sea-query v0.32.2 (*)
├── sea-orm-migration v1.1.6
│   ├── sea-orm v1.1.6 (*)
│   ├── sea-orm-cli v1.1.6
│   │   ├── sea-schema v0.16.1
│   │   │   ├── sea-query v0.32.2 (*)
│   │   │   └── sea-schema-derive v0.3.0 (proc-macro)
│   ├── sea-schema v0.16.1 (*)
├── sea-query v0.32.2 (*)
```

DB: Postgres 16.4
OS: Linux

Expurple:
Agree, it's very annoying

Huliiiiii:
I think, for now, adding length checks to some methods that return a Vec would not cause breaking changes (e.g. methods on LoaderTrait, query_all), and would also reduce the amount of queries that don't return any data.

Currently, Inserter doesn't have the One and Many variants like the other methods, which means insert_many shares the same struct as insert(one).

If this is refactored into InsertOne and InsertMany, TryInsertResult::Empty will no longer be needed, as an empty Vec or None can be returned.

[...]

PR #2592, opened 2025-05-07
Remove insert default on empty behaviour

tyt2y3:
#2528


tyt2y3:
@zevweiss can you see if this is the half of what you want? I've removed insert_default from insert_many

tyt2y3:
close in favour of #2628

Issue #2601, opened 2025-05-20
Can't use `anyhow::Error` in trasactions


bragov4ik:
## Description

`anyhow` is a widely used error library, and it's `Error` type doesn't implement `std::error::Error`.
Also it's a good practice to set generic types' trait requirements to be as weak as possible.

In the `TransactionError` type, the underlying `E` type is currently restricted to be `std::error::Error`, but in fact it is enough to only implement `std::fmt::Display` and `std::fmt::Debug`. This does not seem to break anything and only allows to use more types within transaction context.

I've found https://github.com/SeaQL/sea-orm/issues/1091 but the suggestion to implement `std::error::Error` doesn't work here ([see issue in `anyhow`](https://github.com/dtolnay/anyhow/issues/25#issuecomment-544140480))

I will attach a PR soon.

## Steps to Reproduce

1. Create a function that returns `Result<_, anyhow::Error>`
2. Try to use it in transaction

### Expected Behavior

Works

### Actual Behavior

Doesn't :)

### Workarounds

Change error type to other (not convenient)

## Versions

Active in `master` branch (https://github.com/SeaQL/sea-orm/blob/4b4af1605addc5f5edbb0bd75b03644490004373/src/database/transaction.rs#L103)

PR #2602, opened 2025-05-20
Weaker requirements for underlying `TransactionError`'s error

bragov4ik:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2601

## Changes

- [ ] Weaken requirements for `TransactionError`'s generic parameter across all usage places I've found

## Breaking Changes
I believe it's not a breaking change, since `std::error::Error` already requires these traits.


bragov4ik:
I've run `cargo check` and the repo seems to compile; I hope I didn't miss any place to update the requirements at

tyt2y3 (approved):

Expurple (approved):
Thanks. Seems backward-compatible to me. [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) already implied `Debug + Display`, but now `E` is no longer required to implement `Error` in addition to these.

Issue #1870, opened 2023-09-25
Panic on `load_many` call with different column name on primary key


DMaxter:
## Description

I'm trying to use `load_many` to load the many side of multiple entities. However, if the many-side structure has the column name different from the struct name, and it is a primary key, it panics, as it tries to obtain the parameter through the column name and not through the struct parameter.

As an example, this is a many side of a relation:
```Rust
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "stock")]
pub struct Model {
    #[sea_orm(primary_key, auto_increment = false)]
    pub bakery_id: i32,
    #[sea_orm(primary_key, column_name = "cake_name")]
    pub cake_id: String,
    pub amount: i32,
}
```

I receive a panic message like
```
thread 'main' panicked at /home/dmaxter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sea-orm-0.12.3/src/query/loader.rs:360:37:
Failed at mapping string to column A:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

With deeper debugging, I found out that this is triggered by this line:
https://github.com/SeaQL/sea-orm/blob/0.12.3/src/query/loader.rs#L226

For this specific case, I found out that [here](https://github.com/SeaQL/sea-orm/blob/0.12.3/src/query/loader.rs#L358) the value is `cake_name` while the model given only has `cake_id`, not `cake_name` property

## Steps to Reproduce

1.  Create a postgres database named bakery
2. Fill it with the data on `db.sql`
3. Run the reproducer

### Expected Behavior

Obtaining the many side

### Actual Behavior

Panic

### Reproduces How Often

Always

### Workarounds

None, besides renaming the columns to match the model parameters (however this might not be possible, as is the case, because a migration from another platform is happening)

## Reproducible Example

[reproducer.zip](https://github.com/SeaQL/sea-orm/files/12720372/reproducer.zip)

## Versions

├── sea-orm v0.12.3
│   ├── sea-orm-macros v0.12.3 (proc-macro)
│   │   ├── sea-bae v0.2.0 (proc-macro)
│   ├── sea-query v0.30.1
│   ├── sea-query-binder v0.5.0
│   │   ├── sea-query v0.30.1 (*)

tyt2y3:
Thank you for the investigation. I am confident that this is fixable, although can you illustrate the problem by showing the examples of:

1. what works
2. what doesn't work
[...]

PR #2603, opened 2025-05-22
fix: include column_name in DeriveColumn FromStr impl

lazulit3:
## PR Info
I'm pretty sure this fixes the cause of:
- https://github.com/SeaQL/sea-orm/discussions/2134
- https://github.com/SeaQL/sea-orm/issues/1870

I encountered the same error when trying to use `load_many()`

```
thread 'test::all_artists' panicked at /REDACTED/sea-orm/src/query/loader.rs:361:21:
Failed at mapping string to column A:1: ID
```

My generated entities have column names in an (almost) UpperCamelCase style: `#[sea_orm(column_name = "UpperCamelCase")]` (with some exceptions like `ArtistID`)

When expanding my `DeriveEntityModel` and then `DeriveColumn`, I found that the `FromStr` implementation did not include my column names:

```rust
#[automatically_derived]
impl std::str::FromStr for Column {
    type Err = sea_orm::ColumnFromStrErr;
    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
        match s {
            "id" | "id" => Ok(Column::Id),
            "artist_id" | "artistId" => Ok(Column::ArtistId),
            _ => Err(sea_orm::ColumnFromStrErr(s.to_owned())),
        }
    }
}
```

This ensures that we include values from the `column_name` attributes too, like so:
```rust
            "id" | "id" | "ID" => Ok(Column::Id),
            "artist_id" | "artistId" | "ArtistID" => Ok(Column::ArtistId),
```

## Bug Fixes

- https://github.com/SeaQL/sea-orm/discussions/2134
- https://github.com/SeaQL/sea-orm/issues/1870

## Changes

- `Column::from_str()` now recognizes values from `#[sea_orm(column_name = "...")]` attributes set on:
  - The `Column` variants (when `Column` is an expanded definition with `DeriveColumn`)
  - The `Model` fields (when `DeriveEntityModel` is used) 

## Testing

I added a test case to an already existing test for `DeriveEntityModel` related to column names. This verifies the _specific_ change (and would identify a regression). The column used has a `column_name` with weird casing that is not covered by the snake_case or lowerCamelCase styles (that the macro already handled).

More importantly, I verified that this fixed my problems with `load_many()` in a personal project— which is the symptom that led to the linked issue & discussion.

### DIFF
--- a/sea-orm-macros/src/derives/column.rs
+++ b/sea-orm-macros/src/derives/column.rs
@@ -70,14 +70,36 @@ pub fn impl_col_from_str(ident: &Ident, data: &Data) -> syn::Result<TokenStream>
         }
     };
 
-    let columns = data_enum.variants.iter().map(|column| {
-        let column_iden = column.ident.clone();
-        let column_str_snake = column_iden.to_string().to_snake_case();
-        let column_str_mixed = column_iden.to_string().to_lower_camel_case();
-        quote!(
-            #column_str_snake | #column_str_mixed => Ok(#ident::#column_iden)
-        )
-    });
+    let columns = data_enum
+        .variants
+        .iter()
+        .map(|column| {
+            let column_iden = column.ident.clone();
+            let column_str_snake = column_iden.to_string().to_snake_case();
+            let column_str_mixed = column_iden.to_string().to_lower_camel_case();
+
+            let mut column_name = column_str_snake.clone();
+            for attr in column.attrs.iter() {
+                if !attr.path().is_ident("sea_orm") {
+                    continue;
+                }
+                attr.parse_nested_meta(|meta| {
+                    if meta.path.is_ident("column_name") {
+                        column_name = meta.value()?.parse::<LitStr>()?.value();
+                    } else {
+                        // Reads the value expression to advance the parse stream.
+                        // Some parameters, such as `primary_key`, do not have any value,
+                        // so ignoring an error occurred here.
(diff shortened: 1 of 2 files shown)


sea-orm-macros/src/derives/column.rs
```diff
@@ -70,14 +70,36 @@ pub fn impl_col_from_str(ident: &Ident, data: &Data) -> syn::Result<TokenStream>
...
+                        // Some parameters, such as `primary_key`, do not have any value,
+                        // so ignoring an error occurred here.
+                        let _: Option<Expr> = meta.value().and_then(|v| v.parse()).ok();
+                    }
+                    Ok(())
+                })?;
+            }
+            Ok::<TokenStream, syn::Error>(quote!(
+                #column_str_snake | #column_str_mixed | #column_name => Ok(#ident::#column_iden)
+            ))
```
lazulit3, line 100: I copied the attribute parsing from `impl_default_as_str` (above in `column.rs`).

In cases where `#[sea_orm(column_name = "...")]` is not found, this results in a redundant string match like:
```rust
"artist_id" | "artistId" | "artist_id" => Ok(Column::ArtistId),
```

I recognize that this is not very clean— it would be nicer to only include additional strings when they differ.
(I don't have the patience to think through that, so I invite help if we care about the redundancy. <3)

tyt2y3: I think this would be fine as long as it doesnt trigger clippy warning

tyt2y3 (approved):

PR #2606, opened 2025-05-24
Import missing `ExprTrait`

Huliiiiii:
## PR Info

- Closes 

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/890

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/tests/active_enum_tests.rs
+++ b/tests/active_enum_tests.rs
@@ -13,6 +13,7 @@ use sea_orm::{
     sea_query::{BinOper, Expr},
     ActiveEnum as ActiveEnumTrait, DatabaseConnection, FromQueryResult, QuerySelect,
 };
+use sea_query::ExprTrait;
 
 #[sea_orm_macros::test]
 async fn main() -> Result<(), DbErr> {
@@ -141,10 +142,10 @@ pub async fn insert_active_enum(db: &DatabaseConnection) -> Result<(), DbErr> {
     assert_eq!(
         model,
         Entity::find()
-            .filter(
-                Expr::col(Column::Tea)
-                    .binary(BinOper::In, Expr::tuple([Tea::EverydayTea.as_enum()]))
-            )
+            .filter(Expr::col(Column::Tea).binary(
+                BinOper::In,
+                Expr::tuple([ActiveEnumTrait::as_enum(&Tea::EverydayTea)])
+            ))
             .one(db)
             .await?
             .unwrap()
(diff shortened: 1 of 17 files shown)


Huliiiiii:
I found a issue where `ExprTrait::as_enum` would conflict with `ActiveEnum::as_enum`

tests/active_enum_tests.rs
```diff
@@ -141,10 +142,10 @@ pub async fn insert_active_enum(db: &DatabaseConnection) -> Result<(), DbErr> {
...
         model,
         Entity::find()
-            .filter(
-                Expr::col(Column::Tea)
-                    .binary(BinOper::In, Expr::tuple([Tea::EverydayTea.as_enum()]))
-            )
+            .filter(Expr::col(Column::Tea).binary(
+                BinOper::In,
+                Expr::tuple([ActiveEnumTrait::as_enum(&Tea::EverydayTea)])
+            ))
```
Expurple, line 148: > I found a issue where `ExprTrait::as_enum` would conflict with `ActiveEnum::as_enum`

I think, this is acceptable and won't result in much breakage. I have an application that heavily uses `ExprTrait`, but practically never uses `ActiveEnum`. So, I haven't encountered this issue there. I think, that's fairly typical for the user code.

But as I think about it deeper, we should probably have `ActiveEnum: Into<SimpleExpr>`, just like I want to have `ColumnTrait: Into<SimpleExpr>` (https://github.com/SeaQL/sea-orm/discussions/2548). I think, calling `ExprTrait` methods directly on enums would be very nice. And this way, `ActiveEnum` automatically "inherits" `ExprTrait::as_enum` and we can delete `ActiveEnum::as_enum`. Let's do this separately. I added a section with more details in https://github.com/SeaQL/sea-orm/discussions/2548.

Expurple (commented):
Thank you. Everything looks good, but `cargo test --workspace` fails. You need to add two more imports.

After that, we can trigger a CI run to check other test configurations

tests/active_enum_tests.rs
Huliiiiii: `ActiveEnum::as_enum` is different from `ExprTrait::as_enum`, where the latter requires a typename, while the former always uses its own typename.

Expurple (commented):
@tyt2y3, can you trigger a CI run? The PR seems ready in general

Expurple (approved):
Everything seems good in both PRs, thank you.

Now we can:

- merge https://github.com/SeaQL/sea-query/pull/890
- then rollback the `Cargo.toml` changes here (should've kept the Draft status until then, so that we don't forget)
- merge this PR

Expurple:
https://github.com/SeaQL/sea-query/pull/890 has been merged.

Now we can rollback the `Cargo.toml` changes and merge this

Expurple:
@tyt2y3, you can merge this now.

The new CI failures come from https://github.com/SeaQL/sea-query/pull/891 (https://github.com/SeaQL/sea-orm/pull/2607 not being merged yet)

PR #2607, opened 2025-05-25
Fix non_exhaustive enums missing arms

Huliiiiii:
## PR Info

- Closes 

- Dependencies:
  - https://github.com/SeaQL/sea-query/pull/891

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/src/query/combine.rs
+++ b/src/query/combine.rs
@@ -49,6 +49,7 @@ where
                             ColumnRef::Asterisk | ColumnRef::TableAsterisk(_) => {
                                 panic!("cannot apply alias for Column with asterisk")
                             }
+                            _ => unreachable!(),
                         },
                         SimpleExpr::AsEnum(_, simple_expr) => match simple_expr.as_ref() {
                             SimpleExpr::Column(col_ref) => match &col_ref {
@@ -58,6 +59,7 @@ where
                                 ColumnRef::Asterisk | ColumnRef::TableAsterisk(_) => {
                                     panic!("cannot apply alias for AsEnum with asterisk")
                                 }
+                                _ => unreachable!(),
                             },
                             _ => {
                                 panic!("cannot apply alias for AsEnum with expr other than Column")
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,5 +107,5 @@ tests-cfg = ["serde/derive"]
 seaography = ["sea-orm-macros/seaography"]
 
 # This allows us to develop using a local version of sea-query
-# [patch.crates-io]
-# sea-query = { path = "../sea-query" }
+[patch.crates-io]
+sea-query = { git = "https://github.com/Huliiiiii/sea-query.git", branch = "expr-3" }
--- a/src/query/helper.rs
+++ b/src/query/helper.rs
@@ -886,6 +886,7 @@ pub(crate) fn unpack_table_ref(table_ref: &TableRef) -> DynIden {
         | TableRef::SubQuery(_, tbl)
         | TableRef::ValuesList(_, tbl)
         | TableRef::FunctionCall(_, tbl) => SeaRc::clone(tbl),
+        _ => unreachable!(),
     }
 }
 
@@ -900,5 +901,6 @@ pub(crate) fn unpack_table_alias(table_ref: &TableRef) -> Option<DynIden> {
         | TableRef::SchemaTableAlias(_, _, alias)
         | TableRef::DatabaseSchemaTableAlias(_, _, _, alias)
         | TableRef::FunctionCall(_, alias) => Some(SeaRc::clone(alias)),
+        _ => unreachable!(),
     }
 }


sea-orm-codegen/src/entity/relation.rs
```diff
@@ -181,6 +181,7 @@ impl Relation {
             ForeignKeyAction::SetNull => "SetNull",
             ForeignKeyAction::NoAction => "NoAction",
             ForeignKeyAction::SetDefault => "SetDefault",
+            _ => todo!(),
```
Expurple, line 184: Instead of adding a wildcard, this method could be moved into `sea_query` as

```rust
impl ForeignKeyAction {
    pub fn variant_name(&self) -> &'static str {
        match self {
            ..
        }
    }
}
```

similarly to what I proposed for the other methods in an older comment.

But this one's a bit more controversial than the others, since `sea_query` doesn't really care about this functionality. The variant name string is an internal need of `sea_orm_codegen`. What do you think? Should we move it to `sea_query` anyway and forget about wildcard problems?

Also, the original method is public, so we'll need a changelog entry if we remove it from `sea_orm_codegen`. Alternatively, we can keep it for compatibility and just call the `sea_query` method inside. Personally, I'd remove it. It doesn't look like it belongs on `Relation` anyway.

Expurple (changes requested):
Thank you. `cargo test --workspace` passes. We've discovered all relevant `match` statements (unless the CI finds more under the other test configurations). Now we need to decide what to do with them. See the other comments

src/query/combine.rs
```diff
@@ -49,6 +49,7 @@ where
                             ColumnRef::Asterisk | ColumnRef::TableAsterisk(_) => {
                                 panic!("cannot apply alias for Column with asterisk")
                             }
+                            _ => unreachable!(),
```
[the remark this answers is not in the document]
tyt2y3: 100% agree

tyt2y3:
It now contains changes needed for SeaQL/sea-query#890 as well

Expurple:
@tyt2y3, we already had https://github.com/SeaQL/sea-orm/pull/2606 for that. Btw, it's ready to be merged.

Here, everything looks OK too, but we need to fix the remaining `ExprTrait`-related CI failures to know for sure

tyt2y3:
Everything is now okay. Those examples are expected to fail

Expurple:
> Those examples are expected to fail

Why? What's the deal here? I'm out of the loop

sea-orm-cli/Cargo.toml
```diff
@@ -67,3 +67,6 @@ runtime-tokio-native-tls = ["sqlx?/runtime-tokio-native-tls", "sea-schema?/runti
 runtime-actix-rustls = ["sqlx?/runtime-tokio-rustls", "sea-schema?/runtime-actix-rustls"]
 runtime-async-std-rustls = ["sqlx?/runtime-async-std-rustls", "sea-schema?/runtime-async-std-rustls"]
 runtime-tokio-rustls = ["sqlx?/runtime-tokio-rustls", "sea-schema?/runtime-tokio-rustls"]
+
+[patch.crates-io]
+sea-orm = { path = ".." }
```
Expurple: If you're preparing to merge, we should probably roll back these

tyt2y3:
> > Those examples are expected to fail
> 
> Why? What's the deal here? I'm out of the loop

those examples depends on downstream crates: loco and seaography. we have our own fork of loco for now, so may be we can use git dependency there, then everything should work

sea-orm-cli/Cargo.toml
tyt2y3: umm, I intend to keep this for now. otherwise CI is difficult to pass without releasing a new version

Expurple: Then, I think, we could keep it as a git dependency for now:

```suggestion

[patch.crates-io]
sea-orm = { git = "https://github.com/SeaQL/sea-orm.git", branch = "master" }
```

Does this solve the problem with version numbers?

Expurple:
Yeah, I think we should use git dependencies for now. And when CI fails, we fix the fork to make CI pass. This will also help to assess the breakage and prevent the merge if it's too bad. No different from fixing SeaORM before merging breaking SeaQuery PRs

sea-orm-cli/Cargo.toml
tyt2y3: what'd be the difference?

Expurple: Hmm, I thought that this would be somehow better and work in more cases. But I googled the topic now and can't find any benefits. The path override is actually better, because it can also be used while developing locally. Keep it

tyt2y3: oh, and I was wrong. we don't need it here. removed

tyt2y3:
Actually, I have a counter example to consider:

```rust
use crate::ExprTrait;

struct MyType(String);

impl PartialEq for MyType {
    fn eq(&self, other: &Self) -> bool {
        self.0.eq(&other.0)
    }
}
```

it collided with `std::cmp::PartialEq::eq`
```
self.0.eq(&other.0)
   |         ^^^^^^^^^^^^^^^^^^^ expected `bool`, found `SimpleExpr`
```

is this annoyance acceptable?

this is the only change required in seaography

Expurple:
`ExprTrait` has always had `PartialOrd`-named methods that could cause these issues. I even documented this right from the start. Search for this in SeaQuery:

```rust
// Sometimes, you'll have to qualify the call because of conflicting std traits.
```

`ExprTrait` adopted these methods from `Expr` and `SimpleExpr`, where these inherent methods shadowed the `PartialOrd` methods instead of resulting in a conflict error.

In my experience, these conflicts are very rare. In my queries, I usually compare a column and a value, an expression and a value, a column and an expression, etc. These pairs of different types don't have `PartialOrd` implementations. You won't compare a `String` to a `String` in a real query 😁 And when I mean to call `PartialEq::eq`, I always use `==`. It's more readable, and now it's nice to know that it also helps to avoid conflicts. My 50k LoC codebase has *one* qualified `ExprTrait::` call, but I's the `equals` method and done because of `rustfmt`, not because of conflicts. So, not a single conflict in my codebase so far. And I've already refactored it aggressively to use `ExprTrait` and avoid verbose `Expr` wrappers.

After https://github.com/SeaQL/sea-query/pull/890, the only difference is that now you need to `use ExprTrait` in files that didn't use it, but used inherent `Expr` and `SimpleExpr` methods. I've updated the "migration guide" in that PR to mention the potential conflicts.

If this conflict really bothers you, we could use this major release to rename these methods into something more verbose that doesn't conflict. But IMO, this breaking change and verbosity is not worth it. These conflicts are rare

tyt2y3:
> If this conflict really bothers you

I think it's acceptable. but then I can't put it in prelude.

I will merge this now and fix other examples, easier to have it on master

Expurple:
Ok, I'll mention the prelude in the PR description

Expurple (approved):

PR #2623, opened 2025-06-13
Add exists method to PaginatorTrait with tests

IgnisDa:
Introduce an `exists` method to the `PaginatorTrait` to efficiently check for record existence. Include comprehensive tests covering various scenarios, including filtering, joins, and performance comparisons with the count method.

Fixes #730.

### DIFF
--- a/src/executor/paginator.rs
+++ b/src/executor/paginator.rs
@@ -225,6 +225,33 @@ where
     {
         self.paginate(db, 1).num_items().await
     }
+
+    /// Check if any records exist
+    async fn exists(self, db: &'db C) -> Result<bool, DbErr>
+    where
+        Self: Send + Sized,
+    {
+        let paginator = self.paginate(db, 1);
+        let builder = db.get_database_backend();
+        let stmt = SelectStatement::new()
+            .expr(Expr::cust("1"))
+            .from_subquery(
+                paginator
+                    .query
+                    .clone()
+                    .reset_limit()
+                    .reset_offset()
+                    .clear_order_by()
+                    .limit(1)
+                    .to_owned(),
+                "sub_query",
+            )
+            .limit(1)
+            .to_owned();
+        let stmt = builder.build(&stmt);
+        let result = db.query_one(stmt).await?;
+        Ok(result.is_some())
+    }
 }
 
 impl<'db, C, S> PaginatorTrait<'db, C> for Selector<S>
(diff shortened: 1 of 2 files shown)


Expurple:
Thank you! I'll review this a bit later.

TODO for myself: also merge this in `master` (2.0.0).

If you're ok with using your own unreleased fork of 1.1.x until 2.0.0 comes out, you can change the base of the PR to `master` and we'll merge this only in `master`.

IgnisDa:
> Thank you! I'll review this a bit later.
> 
> TODO for myself: also merge this in `master` (2.0.0).
> 
> If you're ok with using your own unreleased fork of 1.1.x until 2.0.0 comes out, you can change the base of the PR to `master` and we'll merge this only in `master`.

Can this be released as a patch 1.1.x release? I think it should be fine since it does not have any breaking changes. That way I can continue to use the crates.io version.

Expurple:
Yeah, I assumed that you may want that. I'm not the one making (the decisions about) releases, but @tyt2y3 will probably do this

tests/common/bakery_chain/seed_data.rs
```diff
@@ -69,7 +70,7 @@ pub async fn init_1(ctx: &TestContext, link: bool) {
             total: Set(10.into()),
             bakery_id: Set(42),
             customer_id: Set(11),
-            placed_at: Set(DateTime::UNIX_EPOCH),
+            placed_at: Set(ChronoDateTime::UNIX_EPOCH.naive_utc()),
```
IgnisDa: did this because this was throwing a deprecation warning

Expurple (approved):
Sorry for the delay. Everything looks good to me!

@tyt2y3. we can merge this and then cherry-pick into `master` too

IgnisDa:
@tyt2y3 Can you take a look at this please? Thanks!

src/executor/paginator.rs
```diff
@@ -225,6 +225,33 @@ where
...
+
+    /// Check if any records exist
+    async fn exists(self, db: &'db C) -> Result<bool, DbErr>
+    where
+        Self: Send + Sized,
+    {
+        let paginator = self.paginate(db, 1);
+        let builder = db.get_database_backend();
+        let stmt = SelectStatement::new()
+            .expr(Expr::cust("1"))
```
Huliiiiii, line 232: I think using `Expr::exists` would be better.

IgnisDa: @Huliiiiii Thanks for the comment. Why would it be better?

Huliiiiii: Well, I forgot that different databases return different types of EXISTS.

PR #2626, opened 2025-06-14
Add error variant BackendNotSupported

tyt2y3:
As part of the effort to remove panic and wraps.

Another PR underway.


Expurple (approved):
Thank you. This looks like a step in the right direction!

Note that `DbErr` isn't `non_exhaustive` yet, so adding a new variant is a breaking change and we can't backport this to `1.1.x`

tyt2y3:
Yes, I am targeting this on master. will not backport this.

PR #2627, opened 2025-06-15
Remove panics from select / update / delete

tyt2y3:
This refactor removed two possible panics:

1. arity mismatch
2. primary key not set

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -83,12 +83,12 @@ pub enum DbErr {
         ctx: &'static str,
     },
     /// (Primary) Key arity mismatch
-    #[error("Key arity mismatch: expected {expected}, got {got}")]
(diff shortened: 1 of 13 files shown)


tyt2y3:
@Huliiiiii please take a look

src/error.rs
```diff
@@ -81,6 +82,20 @@ pub enum DbErr {
         /// Context
         ctx: &'static str,
     },
+    /// (Primary) Key arity mismatch
+    #[error("Key arity mismatch: expected {expected}, got {got}")]
+    KeyArityMismatch {
```
Huliiiiii, line 87: I think "received" is a more appropriate and commonly used term.

tyt2y3:
CI only fails because I have to release a new sea-query version

PR #2635, opened 2025-06-25
FromJsonQueryResult: panic on serialization failures

nodeav:
## Breaking Changes

As of 1.1.12, attempting to save/insert/update a model with a json-backed struct fails *silently* when the struct fails to serialize (which is quite a rare occurrence in my experience). 

It makes sense to panic here since this is almost certainly a programmer error, and the alternative (using `TryFrom` instead of `From`) would be a major change throughout the codebase as far as I can tell.

### DIFF
--- a/src/database/proxy.rs
+++ b/src/database/proxy.rs
@@ -68,7 +68,7 @@ impl From<ExecResult> for ProxyExecResult {
         match result.result {
             #[cfg(feature = "sqlx-mysql")]
             ExecResultHolder::SqlxMySql(result) => Self {
-                last_insert_id: result.last_insert_id() as u64,
+                last_insert_id: result.last_insert_id(),
                 rows_affected: result.rows_affected(),
             },
             #[cfg(feature = "sqlx-postgres")]
@@ -93,7 +93,7 @@ impl From<ExecResult> for ProxyExecResult {
 
 /// Defines the structure of a Row for the [ProxyDatabase]
 /// which is just a [BTreeMap]<[String], [Value]>
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct ProxyRow {
     /// The values of the single row
     pub values: BTreeMap<String, Value>,
@@ -103,14 +103,6 @@ impl ProxyRow {
     /// Create a new [ProxyRow] from a [BTreeMap]<[String], [Value]>
     pub fn new(values: BTreeMap<String, Value>) -> Self {
         Self { values }
-    }
-}
-
-impl Default for ProxyRow {
-    fn default() -> Self {
-        Self {
-            values: BTreeMap::new(),
-        }
     }
 }
 
@@ -141,9 +133,9 @@ impl From<ProxyRow> for QueryResult {
 }
 
 #[cfg(feature = "with-json")]
-impl Into<serde_json::Value> for ProxyRow {
-    fn into(self) -> serde_json::Value {
-        self.values
+impl From<ProxyRow> for serde_json::Value {
+    fn from(val: ProxyRow) -> serde_json::Value {
+        val.values
             .into_iter()
             .map(|(k, v)| (k, sea_query::sea_value_to_json_value(&v)))
             .collect()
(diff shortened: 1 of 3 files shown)


ytanay:
Wow - I was just about to make this change myself; we've had a security near-miss due to the miss error here. Thanks!!

mastershif:
This is a must

Expurple (approved):
I'm not familiar with this area, so I'm not sure about the impact of `TryFrom`. I agree that panicking is already better than the current behavior. I'd definitely merge this into `master`. Right now, it's a dev branch for the upcoming 2.0.0 release, so we can land breaking changes without hesitation.

@tyt2y3, what do you think about merging into `1.1.x`? I'd do that too. I doubt that anyone intentionally relies on the current behavior.

Afterwards, we can explore the `TryFrom` direction on `master`

nodeav:
@Expurple Thanks for the review!
I see that there are some unrelated cargo fmt and clips failures; Should I fix them in this PR as well?

(re-pushed to fix a test)

Expurple:
Meanwhile, you can add this in your project's `Cargo.toml`:

```toml
[patch.crates-io]
sea-orm = { git = "https://github.com/nodeav/sea-orm", branch = "panic-on-serialization-failure" }
```

(in case someone's affected and doesn't know about this Cargo feature)

Expurple:
@nodeav oops, something more serious broke this time 😁 As I think about it, let's roll back that `Cargo.toml` change. It can also cause conflicts when cherry-picking the commit into `master` later

nodeav:
> @nodeav oops, something more serious broke this time 😁 As I think about it, let's roll back that `Cargo.toml` change. It can also cause conflicts when cherry-picking the commit into `master` later

Oops, should be resolved now hopefully

tyt2y3:
sorry, as much as I'd like to fix a bug, this contradicts the effort to reduce panics in SeaORM.
ideally, we should somehow propagate and return this error.
do you have a full example of the scenario (with entity, and db operations)?
there might be something we can do without breaking the API

ytanay:
> sorry, as much as I'd like to fix a bug, this contradicts the effort to reduce panics in SeaORM. ideally, we should somehow propagate and return this error. do you have a full example of the scenario (with entity, and db operations)? there might be something we can do without breaking the API

The situation we experienced is an object that fails `serde` serialization, e.g. 
```rust
#[derive(Serialize])
#[serde(tag = "type")]
enum Foo {
    Opt1(String)
    // ...
}
```

When we try to update a column containing this type, the update operation is ignored without errors or warnings. Not only did this eat up considerable engineerings efforts to isolate (it's pretty unusual that ORMs silently ignore operations), but when these objects are sensitive (e.g. authentication/authorization states), they can lead to very dangerous situations.

I completely agree that panics are a last resort but they're always better than silently failing, especially when the error state threatens the integrity of the database.

tyt2y3:
@nodeav @ytanay if you can add a test case under `tests/` that'd be great, and I can see if it's possible to return an Error with some plumbing work

Expurple:
@tyt2y3, I agree with them on this:

> panics are a last resort but they're always better than silently failing

A panic is not ideal long-term, but it's already better than what we have now. We can quickly merge the panic, let everyone sleep peacefully, and *then* proceed with the `Error` experiments to improve the situation further.

nodeav:
@tyt2y3 
> @nodeav @ytanay if you can add a test case under tests/ that'd be great, and I can see if it's possible to return an Error with some plumbing work

Moved the test to the `tests/` directory.

I'm willing to try going through with the move to `TryFrom` - though that's arguably more of a breaking change than panicking since it can cause compilation failures in users that rely on the `From` impl (or we can panic in the `From` impl)

Anyway, up to you obviously

tyt2y3:
I see. thanks. can you revert the sea-orm-codegen diff? I will merge this afterwards

nodeav:
> I see. thanks. can you revert the sea-orm-codegen diff? I will merge this afterwards

@tyt2y3 Done!

PR #2641, opened 2025-07-01
Return iterators in loader

negezor:
## PR Info

I think the previous version with iterators was more concise. It is worth remembering sometimes that `.collect()` can also work with `Result` and iterators have `.try_fold()`. 
- This change partially reverts this commit https://github.com/SeaQL/sea-orm/commit/22ec68c3d874140457d244cdfa47e01849f6ad8c

### DIFF
--- a/src/query/loader.rs
+++ b/src/query/loader.rs
@@ -150,11 +150,10 @@ where
             return Ok(Vec::new());
         }
 
-        let mut keys: Vec<ValueTuple> = Default::default();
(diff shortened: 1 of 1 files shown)


tyt2y3:
> Open source is a job that doesn't pay

sigh... can't help but notice this haha.

anyway I won't fight against this, although I prefer imperative version more

tyt2y3 (approved):

PR #2655, opened 2025-07-06
with-uuid: Implement TryFromU64 for format variants

### LINKED ISSUE #2654: `TryFromU64` missing on Uuid Format Variants
## Description

The `Uuid` format variants like `Hyphenated`, `Simple`, `Braced` and `Urn` are missing the `TryFromU64` implementation making them unusable as a primary key.

This is extremely important because sqlite does not support Uuids, and the default sqlx implementation deserializes the column as a 16 byte field instead of the 32 byte hyphenated text. To overcome this, the `Hyphenated` type must be used instead of `Uuid`. This again though is not possible because it can't be used as a primary key.

## Steps to Reproduce

1. Make a new entity using type `uuid::fmt::Hyphenated` as the primary key.

### Expected Behavior

Compiles.

### Actual Behavior

Doesn't compile because it's missing an implementation for `TryFromU64`:

```
the trait bound `Hyphenated: TryFromU64` is not satisfied
the following other types implement trait `TryFromU64`:
  (T0, T1)
  (T0, T1, T2)
  (T0, T1, T2, T3)
  (T0, T1, T2, T3, T4)
  (T0, T1, T2, T3, T4, T5)
  (T0, T1, T2, T3, T4, T5, T6)
  (T0, T1, T2, T3, T4, T5, T6, T7)
  (T0, T1, T2, T3, T4, T5, T6, T7, T8)
and 29 others
required for `Hyphenated` to implement `TryFromU64`

Related information:

  * primary_key.rs#49,11: required by a bound in `sea_orm::PrimaryKeyTrait::ValueType`

```

### Workarounds

None that I know of. This is a breaking bug. `Hyphenated` can't be used as a primary key. I'm not sure if I'm doing something very wrong or how this has slipped past everyone for so long.

## Versions

1.1.13

Kek5chen:
## PR Info

- Closes #2654 

## Bug Fixes

- [x] Implements `TryFromU64` for `uuid` format types so they can be used as a primary key.

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -1522,6 +1522,18 @@ try_from_u64_err!(rust_decimal::Decimal);
 #[cfg(feature = "with-uuid")]
 try_from_u64_err!(uuid::Uuid);
 
+#[cfg(feature = "with-uuid")]
+try_from_u64_err!(uuid::fmt::Simple);
+
+#[cfg(feature = "with-uuid")]
+try_from_u64_err!(uuid::fmt::Hyphenated);
+
+#[cfg(feature = "with-uuid")]
+try_from_u64_err!(uuid::fmt::Urn);
+
+#[cfg(feature = "with-uuid")]
+try_from_u64_err!(uuid::fmt::Braced);
+
 #[cfg(feature = "with-ipnetwork")]
 try_from_u64_err!(ipnetwork::IpNetwork);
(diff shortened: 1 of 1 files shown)


Expurple:
@tyt2y3, I don't have permissions to cherry pick this commit into `master` without creating another PR

PR #2667, opened 2025-07-18
Update sea-query to rc.4

tyt2y3:
The biggest diff is the change of the Iden system.


tyt2y3:
Actually it's now rc.6

PR #2686, opened 2025-08-08
Allow DerivePartialModel to have nested aliases

tyt2y3:
The current solution attaches the alias definition to the nested model, which is not very flexible.
Defining it in the parent model is the better way.

We only need to fix / rewrite the test cases


tyt2y3 (commented):
Approve @Huliiiiii thanks a lot!

tests/partial_model/local/model/bakery.rs
```diff
@@ -0,0 +1,36 @@
...
+        to = "super::worker::Column::Id"
+    )]
+    Cashier,
+}
+
+impl Related<super::worker::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::Manager.def()
+    }
+}
```
tyt2y3:
```suggestion
```

Huliiiiii: This is just to make it compile 🤣

PR #2688, opened 2025-08-09
Refactor CI

Huliiiiii:
I noticed that the action cache is never hit.
Instead of fixing it, I think it’s better to replace it with sccache.

In addition:
- Enable clippy for sea-orm-cli as we have upgraded clap to v4
- Disable debuginfo
- Disable incremental compilation
- Split test build and execution


tyt2y3:
Thank you!

PR #2705, opened 2025-08-27
chore: replace async-std by tokio

cimandef:
## PR Info

Since `async-std` is discontinued this PR meant to change the template and use tokio instead: https://rustsec.org/advisories/RUSTSEC-2025-0052


Huliiiiii (approved):

tyt2y3 (approved):
I'll merge this into main instead of 1.1.x

PR #2710, opened 2025-08-28
feat: Allow returning syntax for MariaDB

### LINKED ISSUE #2709: Allow `exec_with_returning` with MariaDB
## Motivation

Currently, when using a MariaDB as backend, it is hidden behind the MySql backend, which is normal. However, there are things that are not available in MySQL but in MariaDB, like the `INSERT ... RETURNING` syntax. Nevertheless, the methods building queries with the "returning" syntax are configured to be only available with the Postgres and Sqlite backend (see https://docs.rs/sea-orm/latest/src/sea_orm/database/db_connection.rs.html#585).

## Proposed Solutions

Just add a match arm in the [`DbBackend::support_returning`](https://docs.rs/sea-orm/latest/sea_orm/type.DbBackend.html#method.support_returning) method, for the MySql backend, with an additional feature, similar to `"sqlite-use-returning-for-3_35"`, e.g. `"mariadb"`.

This way, developers using MariaDB will be able to use the "returning" syntax, by enabling this feature.

ahmadbky:
## PR Info

- Closes #2709.

## New Features

- [X] Adds the `"mariadb"` feature flag.
- [X] Check for it when matching the MySql database backend when checking if the DB backend supports the "returning" syntax (see method https://docs.rs/sea-orm/1.1.14/sea_orm/type.DbBackend.html#method.support_returning).


Expurple (approved):
Thank you.

I recognize that the clippy errors in CI come from the base branch and are not related to your change

Cargo.toml
```diff
@@ -93,6 +93,7 @@ sqlx-all = ["sqlx-mysql", "sqlx-postgres", "sqlx-sqlite"]
 sqlx-mysql = ["sqlx-dep", "sea-query-binder/sqlx-mysql", "sqlx/mysql"]
 sqlx-postgres = ["sqlx-dep", "sea-query-binder/sqlx-postgres", "sqlx/postgres", "postgres-array"]
 sqlx-sqlite = ["sqlx-dep", "sea-query-binder/sqlx-sqlite", "sqlx/sqlite"]
+mariadb = ["sqlx-mysql"]
```
Huliiiiii, line 96: Hi, thanks for the contribution!

Regarding feature flag names, I feel MariaDB is a bit vague. I think the following options are clearer:

- `(with-)mariadb-returning` if we want more fine-grained feature control.
- `(with-)mariadb-dialect` if we intend to group all MariaDB-specific features under this flag.

@Expurple @tyt2y3 What do you think?

Huliiiiii: I also think that implicitly enabling sqlx-mysql is unnecessary.

tyt2y3: I agree with you. let's go with `mariadb-use-returning` to be more consistent with `sqlite-use-returning-for-3_35`.
I;'ll make the change after merge.

PR #2720, opened 2025-09-04
Fix enum casting in DerivePartialModel with table alias

Huliiiiii:


sea-orm-macros/src/derives/partial_model.rs
```diff
@@ -274,12 +274,16 @@ impl DerivePartialModel {
                     <#entity as ::sea_orm::EntityTrait>::Column::#variant_name
                 };
 
-                let maybe_aliased_column = match alias {
+                // We cast enum as text in select_as if the backend is postgres
+
+                let non_nested = match alias {
                     Some(alias) => quote! {
-                        ::sea_orm::sea_query::Expr::col((#alias, #column))
+                        let col_expr = ::sea_orm::sea_query::Expr::col((#alias, #column));
```
tyt2y3: may be I missed this previously, we don't normally use `::sea_orm`, just `sea_orm`, because it allows the module path to be re-exported from a different crate.

PR #2731, opened 2025-09-22
Support map sqlx connect options

### LINKED ISSUE #2729: Support adding connection opts to sqlx::{Pg,MySql,Sqlite}ConnectOptions
## Motivation
From time to time, it is necessary to config the connection opts to a database, [for example](https://github.com/launchbadge/sqlx/issues/3205#issuecomment-2181827300),
```rs
#[async_std::main]
async fn main() -> Result<(), Box<dyn Error>> {
    // sqlx setting, source: https://github.com/launchbadge/sqlx/issues/3205#issuecomment-2181827300
    {
        let options = MySqlConnectOptions::parse("<database url>")?
            // Disable these options to not send `SET @@sql_mode=CONCAT(@@sql_mode, {})` which is not supported on StarRocks.
            // Defaults to `true` on the server side for MySQL 8 anyway
            .no_engine_substitution(false)
            // MySQL uses `||` as logical OR instead of string concatenation (standard SQL):
            // https://dev.mysql.com/doc/refman/8.4/en/logical-operators.html#operator_or
            .pipes_as_concat(false);
    }

    cli::run_cli_with_connection(migration::Migrator, |opt| async {
        // What do I do to `opt` to apply `no_engine_substitution` and `pipes_as_concat` ?
        Database::connect(opt).await
    })
    .await;

    Ok(())
}
```
At the moment, there is no way to configure that(at least I couldn't work it out).
## Proposed Solutions

From my limited understanding, this could/might theoretically be done in the [url parsing](https://docs.rs/sqlx-mysql/0.8.6/src/sqlx_mysql/options/parse.rs.html#48) but I am not sure why `sqlx` just hand picked a few keys only.

Without changing sqlx, I am not sure how it should be done as the `opt` needs to be modified somewhere between https://github.com/SeaQL/sea-orm/blob/1.1.x/src/driver/sqlx_mysql.rs#L63C17-L87

Huliiiiii:
- Closes #2729

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ ipnetwork = { version = "0.20", default-features = false, optional = true }
 ouroboros = { version = "0.18", default-features = false }
 url = { version = "2.2", default-features = false }
 thiserror = { version = "2", default-features = false }
+derive_more = { version = "2.0.1", features = ["debug"] }
 
 [dev-dependencies]
 smol = { version = "1.2" }
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -46,7 +46,7 @@ use crate::error::*;
 pub struct Database;
 
 /// Defines the configuration options of a database
-#[derive(Clone)]
+#[derive(derive_more::Debug, Clone)]
 pub struct ConnectOptions {
     /// The URI of the database
     pub(crate) url: String,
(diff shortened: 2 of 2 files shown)


Huliiiiii:
@chungwong What do you think of this API?

chungwong:
Originally I thought `impl core::fmt::Debug for ConnectOptions` is quite annoying and I went on the journey and see if the following is feasible.
```rs
diff --git a/src/driver/sqlx_mysql.rs b/src/driver/sqlx_mysql.rs
index 872bc1f5..505a0bed 100644
--- a/src/driver/sqlx_mysql.rs
+++ b/src/driver/sqlx_mysql.rs
@@ -60,6 +60,17 @@ impl SqlxMySqlConnector {
     /// Add configuration options for the MySQL database
     pub async fn connect(options: ConnectOptions) -> Result<DatabaseConnection, DbErr> {
+        Self::connect_and_then(options, |o| o).await
+    }
+
+    #[instrument(level = "trace")]
+    pub async fn connect_and_then<F>(
+        options: ConnectOptions,
+        and_then: F,
+    ) -> Result<DatabaseConnection, DbErr>
+    where
+        F: FnOnce(MySqlConnectOptions) -> MySqlConnectOptions + std::fmt::Debug,
+    {
         let mut opt = options
             .url
             .parse::<MySqlConnectOptions>()
```
And no, `FnOnce` and its friends don't implement `Debug` and it won't work. Now I might understand why you went with `*_opts_fn`

PR #2732, opened 2025-09-22
sea-orm-rocket: update for compatibility with rocket 0.5.1

### LINKED ISSUE #2730: sea-orm-rocket: update for 0.5.1 compatibility
## Motivation

Rocket has, since the last release of `sea-orm-rocket`, come out with a new version - `v0.5.1`. This version has completely changed how database pools and connections integrate, and the current release of `sea-orm-rocket` is entirely incompatible with the current version. While single database connections are still trivial to use as per the guide, the traditional model of having database pools managed by Rocket using sea-orm-rocket requires changes, including moving over to the new `rocket_db_pools` management interface which is now separate from Rocket itself.

## Proposed Solutions

I currently use the following patched repo, which works very well: https://github.com/amyipdev/sea-orm-rocket/ 

Feel free to utilize the changes made there once they pass integration testing. I've locally numbered the version as "0.6.0" to disambiguate it from other incompatible versions, but I'm by no means partial to what the crate version number should be.

Once similar changes are merged in, the `sea-orm-rocket` and `sea-orm-rocket-codegen` crates should have updates pushed ASAP. I can't find the old guide I used to get my v0.5.0-era `sea-orm-rocket` pools up, but if anyone has that on hand it should be updated as the code in it would no longer be compatible - I'm down to provide the code I'm using currently for it.

## Additional Information

Is there anyone who's keeping an eye out for new Rocket releases? I'd imagine Rocket will probably be aiming for a v0.5.2 or even a v0.6 soon. I know Sergio's been working on HTTP/3 support, and while it isn't certain that breaking database changes will be made, database support also isn't [an afterthought either](https://github.com/rwf2/Rocket/commit/1435f6574fbcbe578720b45f618c6c4b74294acd).

amyipdev:
## PR Info

- Closes #2730 

## New Features

- [x] Adds Rocket v0.5.1 compatibility via a new crate version (v0.6.0)

## Bug Fixes

- [x] The current versions of sea-orm-rocket and sea-orm-rocket-codegen are not compatible with Rocket v0.5.1, which is not declared anywhere. This provides a new version, v0.6.0, which is compatible.

## Breaking Changes

- [ ] I haven't tested to be sure, but I'm pretty sure this patch would be incompatible with Rocket v0.5.0, which is why the proposed version number is bumped to v0.6.0 such that it indicates incompatibility with the prior v0.5.5. Especially given that the README is currently empty, it might be a good idea to put an explainer in there (or at least somewhere) of which crate versions are usable for which versions of Rocket. If this would be helpful, lemme know and I'll send another commit through!

### DIFF
--- a/examples/rocket_example/api/src/lib.rs
+++ b/examples/rocket_example/api/src/lib.rs
@@ -41,11 +41,7 @@ async fn create(conn: Connection<Db>, post_form: Form<post::Model>) -> Flash<Red
 }
 
 #[post("/<id>", data = "<post_form>")]
-async fn update(
-    conn: Connection<Db>,
-    id: i32,
-    post_form: Form<post::Model>,
-) -> Flash<Redirect> {
+async fn update(conn: Connection<Db>, id: i32, post_form: Form<post::Model>) -> Flash<Redirect> {
     let db = conn.into_inner();
 
     let form = post_form.into_inner();
--- a/examples/rocket_okapi_example/api/src/okapi_example.rs
+++ b/examples/rocket_okapi_example/api/src/okapi_example.rs
@@ -29,10 +29,7 @@ pub type DataResult<'a, T> =
 /// # Add a new post
 #[openapi(tag = "POST")]
 #[post("/", data = "<post_data>")]
-async fn create(
-    conn: Connection<Db>,
-    post_data: DataResult<'_, post::Model>,
-) -> R<Option<String>> {
+async fn create(conn: Connection<Db>, post_data: DataResult<'_, post::Model>) -> R<Option<String>> {
     let db = conn.into_inner();
     let form = post_data?.into_inner();
     let cmd = Mutation::create_post(&db, form);


amyipdev:
@Huliiiiii

sea-orm-rocket/lib/src/config.rs
```diff
@@ -85,5 +85,6 @@ pub struct Config {
     pub idle_timeout: Option<u64>,
 
     /// Enable SQLx statement logging (default true)
+    #[serde(default)]
```
Huliiiiii, line 88: This change seems a bit unclear. What is it used for?

Huliiiiii:
It would be helpful to add a compatibility table to the README.

Huliiiiii:
Could you please update the examples as well?

sea-orm-rocket/lib/src/config.rs
amyipdev: I'm not sure why it doesn't set the default correctly on newer versions, but it doesn't - without manually allowing it to set a default, it'll crash unless I manually set it.

amyipdev:
> It would be helpful to add a compatibility table to the README.

Will do!

> Could you please update the examples as well?

Where are these examples located?

amyipdev:
Nevermind, I see what you mean on the failing checks. I'll get those updated.

Is the clippy check supposed to fail like that? It's failing on something completely unrelated to the PR.

Huliiiiii:
Just ignore the Clippy check; Those will only fail on the 1.1.x branch.

amyipdev:
@Huliiiiii requested patches added!

sea-orm-rocket/lib/src/config.rs
tyt2y3: this is inconsistent with the doc comment. I will change this to default true

amyipdev:
@Huliiiiii formatting fixed in most recent patch

sea-orm-rocket/lib/src/config.rs
Huliiiiii: I think maybe we need to manually implement the Default trait here. Or we should figure out what's causing this to crash.

tyt2y3 (approved):
thank you all!

PR #2734, opened 2025-09-27
Add cargo semver check to CI

Huliiiiii:


tyt2y3:
Thanks! I agree with this idea, but since we're still in RC stage which breaking changes can happen, will this basically fail on every commit?

Huliiiiii:
The logs show this will only be compared to version 1.1 (as 2.0 has not yet been released).

PR #2744, opened 2025-10-11
Handle NULL values in `from_sqlx_*_row_to_proxy_row` functions

### LINKED ISSUE #2703: Handle NULL values properly in from_sqlx_***_row_to_proxy_row function
eg:
Handle NULL values properly in `from_sqlx_mysql_row_to_proxy_row` function

## Problem Description

In the `from_sqlx_mysql_row_to_proxy_row` function located at `sea-orm-1.1.14/src/driver/sqlx_mysql.rs`, the code currently uses `.expect()` panics when trying to extract values from database rows. This approach assumes that all database columns will always contain non-NULL values, which is not always the case in real-world scenarios.

Image

eg:The current implementation:
```rust
"INT" => Value::Int(Some(
    row.try_get(c.ordinal()).expect("Failed to get integer"),
)),
```

This code will panic when encountering NULL values in integer columns, even though NULL values are perfectly valid in database schemas.

## Proposed Solution

Instead of panicking when encountering NULL values, the function should handle them gracefully by converting them to `Option::None`. For example:

```rust
"INT" => Value::Int(row.try_get(c.ordinal()).ok()),
```

This change would:
1. Properly handle NULL values by converting them to `Option::None`
2. Avoid unnecessary panics that disrupt application flow
3. Maintain type safety while respecting database schema constraints

## Impact

The current implementation causes applications to crash unexpectedly when querying databases that contain NULL values in certain columns. This is particularly problematic for:
1. Applications working with legacy databases that may have NULL values
2. Scenarios where NULL values are intentionally used in database design
3. Situations where database schemas evolve over time and may introduce NULL values

## Additional Considerations

This pattern should be applied consistently across all data type conversions in the function, not just for integer types. The same approach should be used for all database types to ensure consistent behavior.

Huliiiiii:
- Closes #2703

### DIFF
--- a/src/driver/sqlx_mysql.rs
+++ b/src/driver/sqlx_mysql.rs
@@ -356,123 +356,154 @@ pub(crate) fn from_sqlx_mysql_row_to_proxy_row(row: &sqlx::mysql::MySqlRow) -> c
                 (
                     c.name().to_string(),
                     match c.type_info().name() {
-                        "TINYINT(1)" | "BOOLEAN" => Value::Bool(Some(
-                            row.try_get(c.ordinal()).expect("Failed to get boolean"),
-                        )),
-                        "TINYINT UNSIGNED" => Value::TinyUnsigned(Some(
+                        "TINYINT(1)" | "BOOLEAN" => {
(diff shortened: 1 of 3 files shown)


tyt2y3 (approved):
thank you

PR #2745, opened 2025-10-11
feat: allow cross schema type/fk/enum generation

### LINKED ISSUE #2584: sea-orm-cli generate entity fails with cross-referenced types from other schemas
## Description
Entity Generation Fails with Cross-Schema Type References

## Steps to Reproduce

1. Create a database with the following schema:
``` sql
-- Create enum type in public schema
CREATE TYPE public.status_type AS ENUM ('active', 'inactive', 'pending');

-- Create custom schema
CREATE SCHEMA custom_schema;

-- Create table in custom schema that uses the enum from public schema
CREATE TABLE custom_schema.items (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) NOT NULL,
    status public.status_type NOT NULL
);
```

2. Run sea-orm-cli to generate entities for the public schema (works fine):
``` 
sea-orm-cli generate entity \
  -u "postgres://user:pass@localhost/dbname" \
  -o "./entities/public" \
  --database-schema public
```

3. Run sea-orm-cli to generate entities for the custom schema (fails):
```
sea-orm-cli generate entity \
  -u "postgres://user:pass@localhost/dbname" \
  -o "./entities/custom" \
  --database-schema custom_schema
```

### Error Message
`error returned from database: type "status_type" does not exist`

### Expected Behavior

The CLI should correctly recognize that status_type exists in the public schema and generate proper imports or references to it when generating entities for the custom schema.
Ideally we could pass a comma separated list into the `--database-schema` arg.

### Actual Behavior

The CLI fails to recognize types from other schemas, causing generation to fail with a "type does not exist" error.

### Reproduces How Often

Always

### Workarounds
Duplicate the enum across schemas

## Versions

sea-orm: 1.1.4
postgres: 16

kayandra:
## PR Info

  
  - Closes #2584

  
  - Dependencies:
    - N/A

  
  - Dependents:
    - N/A

  ## New Features

  - [x] Cross-schema dependency discovery: When generating entities for a Postgre schema, the CLI now automatically discovers and includes tables, enums, and other types from related schemas

  ## Bug Fixes

  - [x] **Fixed entity generation failure with cross-schema type references**: Previously, running `sea-orm-cli generate entity --database-schema custom_schema` would fail with "type does not exist" errors when tables in `custom_schema` referenced enums or had foreign keys to tables in other schemas (e.g., `public`).

    **Root cause**: sea-schema's discovery was limited to the target schema only, so it couldn't resolve cross-schema type references. This was fixed by first loading all schema types/enums etc before proceeding with generation. 

  ## Breaking Changes

  - [ ] No breaking changes - the fix is backward compatible and uses the existing CLI interface

  ## Changes

  - [x] **Updated `sea-orm-cli/src/commands/generate.rs`**: 
    - Added `search_path` configuration to include `"$user"` and `public` schemas
    - Preloaded the relevant types and schemas beofre generating

  ## Testing

  Manual testing performed with the reproduction case from #2584:

  ```sql
  -- Create enum type in public schema
  CREATE TYPE public.status_type AS ENUM ('active', 'inactive', 'pending');

  -- Create custom schema
  CREATE SCHEMA custom_schema;

  -- Create table in custom schema that uses the enum from public schema
  CREATE TABLE custom_schema.items (
      id SERIAL PRIMARY KEY,
      name VARCHAR(255) NOT NULL,
      status public.status_type NOT NULL
  );

  Before:
  sea-orm-cli generate entity --database-schema custom_schema
  error returned from database: type "status_type" does not exist

  After:
  sea-orm-cli generate entity --database-schema custom_schema
  Analyzing cross-schema dependencies ...
  Will discover schemas: {"custom_schema", "public"}
  Discovering tables in schema: custom_schema
  Discovering tables in schema: public
  Discovered 1 tables across 2 schemas
  ... Done.

  Generated entity correctly includes the cross-schema enum with proper schema annotation.
  ```

### DIFF
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -189,22 +189,108 @@ pub async fn run_generate_command(
                     #[cfg(feature = "sqlx-postgres")]
                     {
                         use sea_schema::postgres::discovery::SchemaDiscovery;
-                        use sqlx::Postgres;
+                        use sqlx::{Postgres, Row};
+                        use std::collections::{HashMap, HashSet};
 
                         println!("Connecting to Postgres ...");
                         let schema = database_schema.as_deref().unwrap_or("public");
-                        let connection = sqlx_connect::<Postgres>(
+                        let pool = sqlx_connect::<Postgres>(
                             max_connections,
                             acquire_timeout,
                             url.as_str(),
                             Some(schema),
                         )
                         .await?;
-                        println!("Discovering schema ...");
-                        let schema_discovery = SchemaDiscovery::new(connection, schema);
-                        let schema = schema_discovery.discover().await?;
-                        let table_stmts = schema
-                            .tables
+
+                        // Discover all schemas that need to be included based on cross-schema references
+                        println!("Analyzing cross-schema dependencies ...");
+
+                        let mut schemas_to_discover = HashSet::new();
+                        schemas_to_discover.insert(schema.to_string());
+
+                        // Query to find all schemas referenced by the target schema via foreign keys
+                        let fk_query = r#"
+                            SELECT DISTINCT
+                                fn.nspname AS foreign_schema
+                            FROM pg_constraint c
+                            JOIN pg_class t ON c.conrelid = t.oid
+                            JOIN pg_namespace n ON t.relnamespace = n.oid
+                            JOIN pg_class ft ON c.confrelid = ft.oid
+                            JOIN pg_namespace fn ON ft.relnamespace = fn.oid
+                            WHERE c.contype = 'f'
+                                AND n.nspname = $1
+                                AND fn.nspname != $1
+                        "#;
+
+                        let fk_rows = sqlx::query(fk_query).bind(schema).fetch_all(&pool).await?;
+
+                        for row in fk_rows {
+                            let foreign_schema: String = row.get("foreign_schema");
+                            schemas_to_discover.insert(foreign_schema);
+                        }
+
+                        // Query to find all schemas that have enums used by the target schema
+                        let enum_schema_query = r#"
+                            SELECT DISTINCT tn.nspname AS type_schema
+                            FROM pg_attribute a
+                            JOIN pg_class c ON a.attrelid = c.oid
+                            JOIN pg_namespace n ON c.relnamespace = n.oid
+                            JOIN pg_type t ON a.atttypid = t.oid
+                            JOIN pg_namespace tn ON t.typnamespace = tn.oid
+                            WHERE n.nspname = $1
+                                AND t.typtype = 'e'
+                                AND tn.nspname != $1
+                        "#;
+
+                        let enum_schema_rows = sqlx::query(enum_schema_query)
+                            .bind(schema)
+                            .fetch_all(&pool)
+                            .await?;
+
+                        for row in enum_schema_rows {
+                            let type_schema: String = row.get("type_schema");
+                            schemas_to_discover.insert(type_schema);
+                        }
+
+                        println!("Will discover schemas: {:?}", schemas_to_discover);
+
+                        // Discover all enums from all relevant schemas
+                        let enum_query = r#"
+                            SELECT n.nspname as schema, t.typname as typename, e.enumlabel
+                            FROM pg_type t
+                            JOIN pg_enum e ON t.oid = e.enumtypid
+                            JOIN pg_namespace n ON n.oid = t.typnamespace
+                            ORDER BY schema, typename, e.enumsortorder
+                        "#;
+
+                        let enum_rows = sqlx::query(enum_query).fetch_all(&pool).await?;
+                        let mut all_enums: HashMap<String, Vec<String>> = HashMap::new();
+                        for row in enum_rows {
+                            let typename: String = row.get("typename");
+                            let enumlabel: String = row.get("enumlabel");
+                            all_enums
+                                .entry(typename)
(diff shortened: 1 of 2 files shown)


sea-orm-cli/src/commands/generate.rs
```diff
@@ -189,22 +189,108 @@ pub async fn run_generate_command(
...
+                        let enum_query = r#"
+                            SELECT n.nspname as schema, t.typname as typename, e.enumlabel
+                            FROM pg_type t
+                            JOIN pg_enum e ON t.oid = e.enumtypid
+                            JOIN pg_namespace n ON n.oid = t.typnamespace
+                            ORDER BY schema, typename, e.enumsortorder
+                        "#;
+
+                        let enum_rows = sqlx::query(enum_query).fetch_all(&pool).await?;
+                        let mut all_enums: HashMap<String, Vec<String>> = HashMap::new();
```
Huliiiiii, line 267: What is this variable used for?

sea-orm-cli/src/commands/generate.rs
```diff
@@ -189,22 +189,108 @@ pub async fn run_generate_command(
...
+                            println!("Discovering tables in schema: {}", discover_schema);
+                            let discovery = SchemaDiscovery::new(pool.clone(), discover_schema);
+                            let discovered = discovery.discover().await?;
+                            all_tables.extend(discovered.tables);
+                        }
+
+                        println!(
+                            "Discovered {} tables across {} schemas",
+                            all_tables.len(),
+                            schemas_to_discover.len()
```
Huliiiiii, line 290: May I ask, are you doing vibe coding? This code is hard to understand and print too much logs.

sea-orm-cli/src/commands/generate.rs
```diff
@@ -278,10 +364,14 @@ where
...
-    // Set search_path for Postgres, E.g. Some("public") by default
+    // Set search_path for Postgres to allow cross-schema type resolution
     // MySQL & SQLite connection initialize with schema `None`
     if let Some(schema) = schema {
-        let sql = format!("SET search_path = '{schema}'");
+        // Always include "$user" and public in search_path to support cross-schema type references
+        // This allows types (like enums) defined in any schema to be discovered
+        // PostgreSQL will search in order: target schema, then "$user", then public
+        // See: https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH
+        let sql = format!("SET search_path = '{schema}', \"$user\", public");
```
Huliiiiii, line 374: I don't think this addresses the issue... If the referenced type comes from another schema, this would still fail. Also, there will be conflicts between tables/types with the same name from different schemas.

sea-orm-cli/src/commands/generate.rs
```diff
@@ -189,22 +189,108 @@ pub async fn run_generate_command(
...
+                            all_tables.extend(discovered.tables);
+                        }
+
+                        println!(
+                            "Discovered {} tables across {} schemas",
+                            all_tables.len(),
+                            schemas_to_discover.len()
+                        );
+
+                        let table_stmts = all_tables
```
Huliiiiii, line 293: If I understand correctly, this will generate tables of all schemas.

kayandra: It finds all the tables and only generates the necessary ones. For example if there's a foreign key on another schema it will bring the relation in, same enums and types in other schemas

sea-orm-cli/src/commands/generate.rs
kayandra: I used Claude to clean up when I was done

sea-orm-cli/src/commands/generate.rs
kayandra: Maybe my approach to testing locally is a bit naive, I tested with a custom schema and public schema and it generated the entities. I'll test again

kayandra: I moved the status_type to a new control_schema and it was generated under the public schema. I'll fix

sea-orm-cli/src/commands/generate.rs
Huliiiiii: This is your code, I didn't see you filtering out tables that are not part of the target schema.
```rust
 let mut all_tables = Vec::new();

for discover_schema in schemas_to_discover.iter() {
    println!("Discovering tables in schema: {}", discover_schema);
    let discovery = SchemaDiscovery::new(pool.clone(), discover_schema);
    let discovered = discovery.discover().await?;
    all_tables.extend(discovered.tables);
}

let table_stmts = all_tables
```

joshuaclayton: I recently checked out 2.0.0-rc.27, applied this change, and can confirm the buggy behavior flagged here that all tables from all of the included schemas end up generating files in the specified entities directory, rather than being limited to just the provided schema.

joshuaclayton: Related, this may also cause a follow-on bug, where if two tables share the same name in schemas pulled in (e.g. `public.notifications` and `otherschema.notifications`), the wrong table may be selected and written to the generated Rust file (confirmed during testing within a larger database where I attempted to generate an entity for a specific overlapped table)

yuriy-yarosh:
I'll need this, as well... 
Will review, maybe found couple points of improvement.

Huliiiiii:
I think we can close this. It seems like this was just vibe coding, and there haven’t been any follow up updates.

Issue #2347, opened 2024-09-02
No way to set `max_lifetime` and `idle_timeout` to None for database connections


kosayoda:
## Description

When `ConnectOptions` is converted to `sqlx::pool::PoolOptions`, the `max_lifetime` and `idle_timeout` fields are set only if the value provided is not `None`:
https://github.com/SeaQL/sea-orm/blob/b68e770f973adea8651420c0865f697bca2f5b9b/src/driver/sqlx_common.rs#L54-L56
https://github.com/SeaQL/sea-orm/blob/b68e770f973adea8651420c0865f697bca2f5b9b/src/driver/sqlx_common.rs#L60-L62

However, `None` is a valid field for `sqlx` to enable infinite connection lifetime and no idle timeout. By default, the values are set to [30 mins and 10 mins respectively](https://github.com/launchbadge/sqlx/blob/ad2936a9c479214f90456102e218189d563f103d/sqlx-core/src/pool/options.rs#L161-L162).

## Use Case

I'm using an SQLite in-memory database and the data will be wiped if connections to it are closed. Therefore I want to keep a connection to the database indefinitely.

jcreekmore:
@kosayoda while it doesn't directly address the issue you created, there is a workaround. I had to do something similar and used the `sqlx::sqlite::SqliteConnectionOptions` and `sqlx::sqlite::SqlitePoolOptions` directly to configure and create the pool and then used the `sea_orm::SqlxSqliteConnector` to build the `sea_orm::DatabaseConnection`.

As an example (I am winging this, so it might not compile, but it is based on what I did):
```rust
let mut conn_opts = sqlx::sqlite::SqliteConnectOptions::from_str(connstr)?;
conn_opts = conn_opts.journal_mode(sqlx::sqlite::SqliteJournalMode::Wal);

let pool_opts = sqlx::sqlite::SqlitePoolOptions::new()
    .idle_timeout(None)
    .max_lifetime(None);

let pool = pool_opts.connect_with(conn_opts).await?;
let connection = sea_orm::SqlxSqliteConnector::from_sqlx_sqlite_pool(pool);
```

I specifically needed this so that I could hook in an `after_connect` hook into the pool and couldn't come up with another way to do it.

Huliiiiii:
#2768 added after_connect
#2770 added methods to modify PoolOptions

PR #2748, opened 2025-10-12
Allowed None for `max_lifetime` and `idle_timeout` Parameters

MinuraPunchihewa:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2347

## Bug Fixes

- [X] This PR enables None values to be passed for the `max_connections` and `idle_timeout` parameters when creating database connections.

### DIFF
--- a/src/driver/sqlx_common.rs
+++ b/src/driver/sqlx_common.rs
@@ -52,13 +52,13 @@ impl ConnectOptions {
             opt = opt.acquire_timeout(connect_timeout);
         }
         if let Some(idle_timeout) = self.idle_timeout {
-            opt = opt.idle_timeout(Some(idle_timeout));
+            opt = opt.idle_timeout(idle_timeout);
         }
         if let Some(acquire_timeout) = self.acquire_timeout {
             opt = opt.acquire_timeout(acquire_timeout);
         }
         if let Some(max_lifetime) = self.max_lifetime {
-            opt = opt.max_lifetime(Some(max_lifetime));
+            opt = opt.max_lifetime(max_lifetime);
         }
         opt = opt.test_before_acquire(self.test_before_acquire);
         opt


src/driver/sqlx_common.rs
```diff
@@ -42,18 +42,14 @@ impl ConnectOptions {
         DB: sqlx::Database,
     {
         let mut opt = sqlx::pool::PoolOptions::new();
-        if let Some(max_connections) = self.max_connections {
-            opt = opt.max_connections(max_connections);
-        }
+        opt = opt.max_connections(self.max_connections);
```
Huliiiiii: You should modify max_lifetime instead of max_connections.

MinuraPunchihewa: Apologies. I used the wrong parameter. Fixed.

MinuraPunchihewa: Confirmed that all of the unit tests are running.

Expurple (commented):
From quickly looking at the diff, it's not obvious why the other surrounding fields are still `if let Some ...`. Seems like either a potential inconsistency or something that needs to be commented and maybe grouped together

Huliiiiii:
> From quickly looking at the diff, it's not obvious why the other surrounding fields are still `if let Some ...`. Seems like either a potential inconsistency or something that needs to be commented and maybe grouped together

https://github.com/launchbadge/sqlx/blob/ad2936a9c479214f90456102e218189d563f103d/sqlx-core/src/pool/options.rs#L161-L162

You are right, `max_lifetime` and i`dle_timeout` should be changed to `Option<Option<...>>`

MinuraPunchihewa:
> > From quickly looking at the diff, it's not obvious why the other surrounding fields are still `if let Some ...`. Seems like either a potential inconsistency or something that needs to be commented and maybe grouped together
> 
> https://github.com/launchbadge/sqlx/blob/ad2936a9c479214f90456102e218189d563f103d/sqlx-core/src/pool/options.rs#L161-L162
> 
> You are right, `max_lifetime` and i`dle_timeout` should be changed to `Option<Option<...>>`

I just pushed a commit; I think this is what we need?

src/database/mod.rs
```diff
@@ -247,12 +247,12 @@ impl ConnectOptions {
 
     /// Set the idle duration before closing a connection
     pub fn idle_timeout(&mut self, value: Duration) -> &mut Self {
```
Huliiiiii: The args should be `impl Into<Option<Duration>>` like sqlx.
The default value of the sqlx options is Some(value) (set by sqlx), but users can set it to None via Some(None).
If the user doesn't set this, the opts.idle_timeout will be None and we won't do anything.

Huliiiiii (changes requested):

src/database/mod.rs
MinuraPunchihewa: @Huliiiiii I have updated the setters now.

Huliiiiii (approved):

MinuraPunchihewa:
@Huliiiiii Does something need fixing here? I see that the unit tests are passing, but the examples are failing?

Huliiiiii:
Remove the wrapped Some.

https://github.com/SeaQL/sea-orm/blob/0887e7357c066ae54f6b37fd16e1611d8ae8d4a7/src/driver/sqlx_common.rs#L38-L66

MinuraPunchihewa:
> Remove the wrapped Some.
> 
> https://github.com/SeaQL/sea-orm/blob/0887e7357c066ae54f6b37fd16e1611d8ae8d4a7/src/driver/sqlx_common.rs#L38-L66

@Huliiiiii Done.

tyt2y3 (approved):

Issue #2690, opened 2025-08-11
Not using ` mod.rs`  as the entry point of the generated entities


Huliiiiii:
### Discussed in https://github.com/SeaQL/sea-orm/discussions/2559

<div type='discussions-op-text'>

<sup>Originally posted by **nyc4m** April 12, 2025</sup>
Hey there :wave: 

When generating entities, I noticed that the output code is using the "old" convention for modules, which is to output the source files as the following:

```bash
sea-orm-cli generate entity --database-url <DB_URL> --output-dir ./src/entities

# Resulting module structure
./src/entities
    /entities
        mod.rs
        foo.rs
        bar.rs
        prelude.rs
```

I'd rather use the "new" convention as described [here](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames). This would result in the following file structure:

```bash
sea-orm-cli generate entity --database-url <DB_URL> --output-dir ./src/entities

# Resulting module structure
./src/entities/
    entities.rs
    entities/
        foo.rs
        bar.rs
        prelude.rs
```

Is there anyway I can achieve this with the current CLI ?
   </div>

kev1N916:
So basically we have to create a separate entities folder in the specified output directory which contains all the entity files and the prelude.rs as well as have an entities.rs in this output directory.

<img width="271" height="222" alt="Image" src="https://github.com/user-attachments/assets/4563a7c6-39d8-41e2-9c2e-e6b6b589c641" />


So if my output directory was run2 this is the way it would look after the entities are generated.
I am just trying to verify if my understanding of the output structure is correct. @Huliiiiii

Huliiiiii:
Yes.

PR #2751, opened 2025-10-12
changed the directory structure of the generated entity files

kev1N916:
## PR Info

Closes https://github.com/SeaQL/sea-orm/issues/2690

## New Features
Changes the directory structure of the generated entitity files.

```
output-dir/
    entities.rs
    entities/
            foo.rs
            bar.rs
            prelude.rs
```

### DIFF
--- a/sea-orm-cli/src/cli.rs
+++ b/sea-orm-cli/src/cli.rs
@@ -298,6 +298,13 @@ pub enum GenerateSubcommands {
 
         #[arg(
             long,
+            default_value = "false",
+            help = "Use the new directory structure when generating the entity files."
+        )]
+        use_new_dir_structure: bool,
+
+        #[arg(
+            long,
             value_delimiter = ',',
             help = "Add extra derive macros to generated model struct (comma separated), e.g. `--model-extra-derives 'ts_rs::Ts','CustomDerive'`"
         )]
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -3,7 +3,14 @@ use sea_orm_codegen::{
     DateTimeCrate as CodegenDateTimeCrate, EntityTransformer, EntityWriterContext, OutputFile,
     WithPrelude, WithSerde,
 };
-use std::{error::Error, fs, io::Write, path::Path, process::Command, str::FromStr};
+use std::{
+    error::Error,
+    fs,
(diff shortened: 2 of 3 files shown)


Huliiiiii:
I think we should add a new flag and set the previous one as the default value.

kev1N916:
@Huliiiiii I've added a new flag.
If it is set the new structure will be followed else the old structure will be followed

sea-orm-cli/src/cli.rs
```diff
@@ -296,6 +296,13 @@ pub enum GenerateSubcommands {
         )]
         lib: bool,
 
+        #[arg(
+            long,
+            default_value = "false",
+            help = "Use the new directory structure when generating the entity files."
+        )]
+        use_new_dir_structure: bool,
```
Huliiiiii, line 304: Some alternatives:
- (use_)new_module_(style/structure/layout): bool
- module_(style/structure/layout): (old, 2018)

@tyt2y3 What do you think?

sea-orm-cli/src/commands/generate.rs
```diff
@@ -239,19 +241,28 @@ pub async fn run_generate_command(
...
 
-            let dir = Path::new(&output_dir);
-            fs::create_dir_all(dir)?;
+            let entities_dir_path = output_dir.clone() + "/entities";
 
+            let output_dir = Path::new(&(output_dir));
+            let entities_dir = Path::new(&(entities_dir_path));
+
+            if use_new_dir_structure {
+                fs::create_dir_all(entities_dir)?;
```
Huliiiiii: I think we can move `entities_dir_path` and `entities_dir` inside this if arm.

sea-orm-cli/src/commands/generate.rs
```diff
@@ -239,19 +241,28 @@ pub async fn run_generate_command(
             );
             let output = EntityTransformer::transform(table_stmts)?.generate(&writer_context);
 
-            let dir = Path::new(&output_dir);
-            fs::create_dir_all(dir)?;
+            let entities_dir_path = output_dir.clone() + "/entities";
```
Huliiiiii: It's better to use PathBuf rather than string concatenation.

tyt2y3:
I think this is a good idea, although we need to provide a flag for compatiblity and fix the conflicts.

PR #2753, opened 2025-10-13
Try to parse u32 in PG as i32 if Oid fails

nihohit:
## Bug Fixes

Currently, trying to parse u32 from an INT column in Postgres fails with `Err value: DBError(Query(SqlxError(ColumnDecode { index: "\"number\"", source: "mismatched types; Rust type core::option::Option<sqlx_postgres::types::oid::Oid> (as SQL type OID) is not compatible with SQL type INT4" })))`

This change adds a fallback - if parsing to Oid fails, try parsing as i32.


nihohit:
@tyt2y3 I rebased and no longer see the compilation failures I saw before.

tyt2y3:
thank you for rebasing. I was about to!

PR #2754, opened 2025-10-13
Allows support for postgres:db_name format

### LINKED ISSUE #2647: Invalid error "There is no database name as part of the url path" for a db url `postgres:dbname` when running `sea-orm-cli generate entity`
## Description

I use this string to reference the postgresql connection: `postgres:dbname`. This works fine everywhere, but not with `sea-orm-cli generate entity` where I have to use `postgres:///dbname` instead.

The error I get is:
```
There is no database name as part of the url path: postgres:dbname
```

## Steps to Reproduce

1. Run `sea-orm-cli generate entity -o directory -u postgres:dbname`

### Expected Behavior

`sea-orm-cli` can connect to the db just fine.

### Actual Behavior

I get the error described above.

### Reproduces How Often

Always reproducible.

### Workarounds

I have specified `-u postgres:///dbname` instead, which worked.

## Reproducible Example

(I don't think that's needed, given this is reproducible without even a proper db being present)

## Versions

│   └── sea-orm v1.1.13
│       ├── sea-orm-macros v1.1.13 (proc-macro)
│       │   ├── sea-bae v0.2.1 (proc-macro)
│       ├── sea-query v0.32.6
│       │   ├── sea-query-derive v0.4.3 (proc-macro)
│       ├── sea-query-binder v0.7.0
│       │   ├── sea-query v0.32.6 (*)
│   └── sea-orm-migration v1.1.13
│       ├── sea-orm v1.1.13 (*)
│       ├── sea-orm-cli v1.1.13
│       ├── sea-schema v0.16.2
│       │   ├── sea-query v0.32.6 (*)
│       │   ├── sea-query-binder v0.7.0 (*)
│       │   ├── sea-schema-derive v0.3.0 (proc-macro)
├── sea-orm v1.1.13 (*)

Using Linux and PostgreSQL.

Thanks!

kev1N916:
## PR Info

 Closes https://github.com/SeaQL/sea-orm/issues/2647

## Bug Fixes
Fixes the bug where we are not able to connect to the postgres database when the url is given in the postgres:db_name format.

### DIFF
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -89,16 +89,20 @@ pub async fn run_generate_command(
                 // Throwing an error if there is no database name since it might be
                 // accepted by the database without it, while we're looking to dump
                 // information from a particular database
-                let database_name = url
-                    .path_segments()
-                    .unwrap_or_else(|| {
-                        panic!(
-                            "There is no database name as part of the url path: {}",
-                            url.as_str()
-                        )
-                    })
-                    .next()
-                    .unwrap();
+                let database_name = match url.path_segments() {
+                    None => {
+                        let paths: Vec<&str> = url.path().split('/').collect();
+                        if paths.len() == 0 {
+                            panic!(
+                                "There is no database name as part of the url path: {}",
+                                url.as_str()
+                            )
+                        } else {
+                            paths[0]
+                        }
+                    }
+                    Some(mut path_segments) => path_segments.next().unwrap_or_else(|| return ""),
+                };
 
                 // An empty string as the database name is also an error
                 if database_name.is_empty() {


sea-orm-cli/src/commands/generate.rs
```diff
@@ -89,16 +89,20 @@ pub async fn run_generate_command(
...
-                    .path_segments()
-                    .unwrap_or_else(|| {
-                        panic!(
-                            "There is no database name as part of the url path: {}",
-                            url.as_str()
-                        )
-                    })
-                    .next()
-                    .unwrap();
+                let database_name = match url.path_segments() {
```
Huliiiiii, line 92: Could you move this logic into a separate function and write a unit test to ensure we get the correct database name?

Huliiiiii (changes requested):

sea-orm-cli/src/commands/generate.rs
Huliiiiii: If unit tests for this already exist, please let me know.

tyt2y3:
thank you. replaced by https://github.com/SeaQL/sea-orm/pull/3197

PR #2755, opened 2025-10-14
Allow user modifications to be preserved on entity generation

Huliiiiii:
Add a new cli flag: `preserve_user_modifications` to control this behaviour


src/lib.rs
tyt2y3, line 1: this might be a glitch of github?

tyt2y3:
image

not sure why got a bunch of new lines in `mod.rs`. but other than that everythings' fine!

PR #2759, opened 2025-10-15
Fix PostgreSQL enum primary key handling in loader

### LINKED ISSUE #2756: `load_many` doesn't work with enum keys in Postgres
## Description

`load_many` doesn't work in Postgres when the key of loaded table contains enum type.

## Steps to Reproduce

Repro as a test case:
https://github.com/bragov4ik/sea-orm/pull/1/files#diff-9ddf329772de6150ce5d40a6f97736a659966e0337a212c92be88b2327317599R907
1. Make two connected tables `A` and `B`
2. `B` has enum `E` as a (part of) its primary key
3. `A` has column with `E` type (that references `B`).
4. Having vector of values from `A`, try to `load_many` from `B` table

### Expected Behavior

`load_many` returns `Ok(_)`

### Actual Behavior

```
thread 'query::loader::tests::test_load_many_custom_enum_key_real_postgres' panicked at src/query/loader.rs:886:14:
Should return something: Query(SqlxError(Database(PgDatabaseError { severity: Error, code: "42883", message: "operator does not exist: tea = text", detail: None, hint: Some("No operator matches the given name and argument types. You might need to add explicit type casts."), position: Some(Original(143)), where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("parse_oper.c"), line: Some(635), routine: Some("op_error") })))
```

### Reproduces How Often

Always

### Workarounds

The only workaround I've found is to change column type to string. Otherwise, the values supplied with query have string type and the conversion always fails.

Another option is to not use `load_many` but it's not always possible/easy

## Reproducible Example

linked in steps

## Versions

`1.1.16` and `master` branch as of Oct 6.

`PostgreSQL 17.5 (Debian 17.5-1.pgdg120+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit`

Huliiiiii:
Closes #2756

### DIFF
--- a/src/query/loader.rs
+++ b/src/query/loader.rs
@@ -896,7 +896,7 @@ fn create_table_columns(table: &TableRef, cols: &Identity) -> Vec<Expr> {
         .cloned()
         .map(|col| table_column(table, &col))
         .map(Expr::col)
-        .collect_vec()
+        .collect()
 }
 
 /// Apply `save_as` to each value in the tuple, e.g. `(Cast(val1 as type1), Cast(val2 as type2), ...)`


Huliiiiii:
@bragov4ik 
Could you please test if this fixes your issue?

bragov4ik:
It seems to solve the issue, thank you! I've checked it on my quickly-made test as well, it passes

I have one question from checking out implementation:
Is it possible that usage of `Condition::any()` in `prepare_condition` might reduce performance for large number of keys? It seems that `IN` might be optimized for larger sets, whilst doing so for `OR` sounds more challenging.

I don't know about pg intrinsics and couldn't find it from quick search, though.

Cargo.toml
```diff
@@ -72,6 +73,7 @@ default = [
     "with-uuid",
     "with-time",
     "sqlite-use-returning-for-3_35",
+    "sqlite-3_15",
```
tyt2y3: why do we need to add a feature flag?

Cargo.toml
```diff
@@ -47,6 +47,7 @@ ouroboros = { version = "0.18", default-features = false }
 url = { version = "2.2", default-features = false }
 thiserror = { version = "2", default-features = false }
 derive_more = { version = "2", features = ["debug"] }
+itertools = "0.14.0"
```
tyt2y3, line 50: umm is it possible to implement the usage by hand and avoid a new dependency

src/query/loader.rs
```diff
@@ -759,41 +707,199 @@ fn dyn_model_to_key(dyn_model: dynamic::Model) -> Result<ValueTuple, DbErr> {
...
+    let keys = keys.iter().unique();
+
+    let table_columns = to
+        .iter()
+        .cloned()
+        .map(|col| table_column(table, &col))
+        .map(Expr::col)
+        .collect_vec();
+
+    if cfg!(not(feature = "sqlite-3_15")) && matches!(backend, DbBackend::Sqlite) {
```
tyt2y3: is this cfg! condition flipped?

Cargo.toml
tyt2y3: we can add a feature flag but we don't have to enable it by default. the default returning already assumes 3.35

src/query/loader.rs
```diff
@@ -1,15 +1,13 @@
...
 };
 use async_trait::async_trait;
-use sea_query::{
-    ColumnRef, DynIden, Expr, ExprTrait, IntoColumnRef, SimpleExpr, TableRef, ValueTuple,
-};
-use std::{
-    collections::{HashMap, HashSet},
-    str::FromStr,
-};
+use itertools::Itertools;
```
tyt2y3: I would move this to function scope where it is used

tyt2y3 (approved):

PR #2761, opened 2025-10-15
Implement incremental schema sync

tyt2y3:


sea-orm-cli/src/commands/generate.rs
```diff
@@ -257,7 +257,7 @@ pub async fn run_generate_command(
                 let file_path = dir.join(name);
                 println!("Writing {}", file_path.display());
 
-                if matches!(
+                if !matches!(
```
tyt2y3, line 260: an unrelated bug

PR #2770, opened 2025-10-23
Add sqlx pool options mapping for drivers

Huliiiiii:


tyt2y3:
oh sorry seems like there is a conflict with existing code already

PR #2780, opened 2025-10-30
Add the ability to use #[sea_orm(extra = "")]

### LINKED ISSUE #2779: Add support for #[sea_orm(extra = "")]
## Motivation

I am porting from migration scripts to the new entity format, however it is not possible to represent a case-insensitive strings.

I'm currently using sea_query's `ColumnDef.extra()` for the migrations (sea_orm 1.1)
```rust
pub fn sqlite_case_insensitive_string<T: IntoIden>(name: T) -> ColumnDef {
    string(name).extra("COLLATE NOCASE").take()
}
```

## Proposed Solutions

It would be nice to be able to write something like (sea_orm 2.0)
```rust
#[sea_orm::model]
#[derive(Debug, Clone, DeriveEntityModel)]
#[sea_orm(table_name = "nocase_test")]
pub struct Model {
    #[sea_orm(primary_key)]
    id: i64,
    #[sea_orm(extra = "COLLATE NOCASE")]
    nocase: String,
}
```

## Additional Information

~~I tried plumbing it in myself in `expand_derive_entity_model()` but quickly hit the issue that `sea_orm::entity::ColumnDef` is not the same as `sea_query::table::ColumnDef`.  I'm not sure how feasible it is to propagate `#[sea_orm(extra = "")]` all the way to table generation.~~  I found the missing link :)

All I need is "COLLATE NOCASE" for SQLite, but having `extra` is more generally useful.

DavidSkrundz:
Allows the following to define a case-insensitive string for the SQLite backend:
```rust
#[sea_orm::model]
#[derive(Debug, Clone, DeriveEntityModel)]
#[sea_orm(table_name = "nocase_test")]
pub struct Model {
	id: i64,
	#[sea_orm(extra = "COLLATE NOCASE")]
	nocase: String,
}
```

## PR Info
- Closes #2779

## Changes

- [x] Add support for `#[sea_orm(extra = "")]` in the new entity format

### DIFF
--- a/sea-orm-macros/src/derives/entity_model.rs
+++ b/sea-orm-macros/src/derives/entity_model.rs
@@ -127,6 +127,7 @@ pub fn expand_derive_entity_model(data: Data, attrs: Vec<Attribute>) -> syn::Res
                     let mut sql_type = None;
                     let mut enum_name = None;
                     let mut is_primary_key = false;
+                    let mut extra = None;
                     let mut seaography_ignore = false;
 
                     let mut column_name = if let Some(case_style) = rename_all {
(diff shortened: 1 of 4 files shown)


tyt2y3:
awesome, you indeed have figured out the intricate marco system :)
I will add a unit test later

PR #2785, opened 2025-10-30
fix(cli): Update active_enum to use Unknown for empty strings

### LINKED ISSUE #2783: sea-orm-cli generate entity blank enum
## Description

Using mysql 8.4 a database column encoded as an `enum('', 'a', 'b')` will fail to generate event though it
is a valid column configuration.

## Steps to Reproduce

1. create a table with an enum column where one of the variants is `''`

```sql
CREATE TABLE repro (
  id int unsigned NOT NULL AUTO_INCREMENT,
  state enum('', 'a', 'b')
  PRIMARY KEY (`id`)
);
```
2. run `sea-orm-cli generate entity -t repro`

### Expected Behavior

An identifier that indicates the Unknown or Empty variant should be used to avoid breaking
codegen.

### Actual Behavior

```
Warning: item '' in the enumeration 'state' cannot be converted into a valid Rust enum member name. It will be converted to its corresponding UTF-8 encoding. You can modify it later as needed.

thread 'main' panicked at $CARGO_HOME/registry/src/index.crates.io-1949cf8c6b5b557f/quote-1.0.41/src/runtime.rs:455:9:
Ident is not allowed to be empty; use Option<Ident>
```

### Reproduces How Often

Always

### Workarounds

Using a test database with an updated table column to use an valid identifier for the empty variant
will allow codegen to work. 

```sql
ALTER TABLE repro CHANGE `state` `state` enum('', 'a', 'b', 'unknown');
UPDATE repro SET state = 'unknown' where state = '';
ALTER TABLE repro CHANGE `state` `state` enum('a', 'b', 'unknown');
```

After that the generated enum can be updated to use 

```
#[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "state")]
enum State {
    #[sea_orm(string_value = "")]
    Unknown,
    #[sea_orm(string_value = "a")]
    A,
    #[sea_orm(string_value = "b")]
    B,
}
```

## Reproducible Example

See above

## Versions

```
$ sea-orm-cli --version                 
sea-orm-cli 2.0.0-rc.16
```

FreeMasen:
## PR Info

- Closes #2783 

## Bug Fixes

- [ ] 

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/sea-orm-codegen/src/entity/active_enum.rs
+++ b/sea-orm-codegen/src/entity/active_enum.rs
@@ -28,7 +28,12 @@ impl ActiveEnum {
             if v.chars().next().map(char::is_numeric).unwrap_or(false) {
                 format_ident!("_{}", v)
             } else {
-                let variant_name = v.to_upper_camel_case();
+                let variant_name = if v.is_empty() {
+                    println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
+                    "__EmptyString".to_string()
+                } else {
+                    v.to_upper_camel_case()
+                };
                 if variant_name.is_empty() {
                     println!("Warning: item '{v}' in the enumeration '{enum_name}' cannot be converted into a valid Rust enum member name. It will be converted to its corresponding UTF-8 encoding. You can modify it later as needed.");
                     let mut ss = String::new();


sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -28,7 +28,12 @@ impl ActiveEnum {
             if v.chars().next().map(char::is_numeric).unwrap_or(false) {
                 format_ident!("_{}", v)
             } else {
-                let variant_name = v.to_upper_camel_case();
+                let variant_name = if v.is_empty() {
+                    println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `Unknown`. You can modify it later as needed.");
+                    "Unknown".to_string()
```
Huliiiiii: I think "Unknown" is a common enough word that it may cause conflicts. It would be better to name it `EmptyString`.

FreeMasen: Thank you for the fast review, I've made the suggested update

Huliiiiii: It seems you forgot to amend your update.

FreeMasen: Woops, sorry about that.

FreeMasen:
Do you want me to try and address the clippy failures? They seem unrelated but I could try and find a solution

Huliiiiii:
This is because the new clippy lint added at 1.91. This can be fixed by replacing the handwritten impl with the derived macro.

Huliiiiii:
The fix has been merged, so you can fix ci by rebase.

tyt2y3:
thanks for reporting this edge case and the fix, never have thought about enum being empty string.

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -28,7 +28,12 @@ impl ActiveEnum {
             if v.chars().next().map(char::is_numeric).unwrap_or(false) {
                 format_ident!("_{}", v)
             } else {
-                let variant_name = v.to_upper_camel_case();
+                let variant_name = if v.is_empty() {
+                    println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
```
tyt2y3, line 32: very thoughtful, thank you!

Issue #2594, opened 2025-05-12
[`sea-orm-migration`] `MigratorTrait.reset` doesn't drop `seaql_migrations` table


lamualfa:
# Description

I was looking to migrator implementation and see that `seaql_migrations` doesn't have migration down implementation. Isn't the `MigratorTrait.reset` should drop all the migrated tables including the `seaql_migrations` table migrated by the library?

The `uninstall` method should also be implemented to this trait.

https://github.com/SeaQL/sea-orm/blob/4b4af1605addc5f5edbb0bd75b03644490004373/sea-orm-migration/src/migrator.rs#L161-L174

# Stack

```toml
sea-orm = { version = "1.1.10", features = [
  "sqlx-postgres",
  "runtime-tokio-rustls",
  "macros",
] }
sea-orm-migration = "1.1.10"
```

kev1N916:
I don't think there is any need to modify **MigratorTrait.reset** if we implement the uninstall method. If somebody wants to drop the migrations table as well as remove the migrations, **reset** followed by **uninstall** could be called.

Huliiiiii:
After discussing with @tyt2y3, we think that it is the right approach to drop the seaql_migrations table when resetting.

Huliiiiii:
It would also be nice to add the uninstall method.

MinuraPunchihewa:
I am happy to give this a go.

Huliiiiii:
Thanks! Assigned to you.

PR #2787, opened 2025-10-31
Fixed Reset and Uninstall Migration Operations

MinuraPunchihewa:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2594

## Bug Fixes

- [X] This PR updates the `reset()` method of the `MigratorTrait` to drop the `seaql_migrations` table in addition to rolling back already applied migrations. A non-destructive uninstall() method has also been added that will also drop the `seaql_migrations` table.

### DIFF
--- a/sea-orm-migration/src/migrator.rs
+++ b/sea-orm-migration/src/migrator.rs
@@ -236,6 +236,23 @@ pub trait MigratorTrait: Send {
         .await
     }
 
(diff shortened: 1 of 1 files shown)


MinuraPunchihewa:
@Huliiiiii I would appreciate a review here.

tyt2y3 (approved):
thank you! I think it makes sense indeed

PR #2788, opened 2025-11-01
Support #[deny(missing_docs)] with #[sea_orm(self_ref)]

DavidSkrundz:
I hit a new error when using `#[sea_orm::model]` + `#![deny(missing_docs)]` + `#[sea_orm(self_ref)]`.  There is an additional public field that gets generated by the macro when defining self-referencing relations.

Sorry for the small change, I've now completed my upgrade from sea-orm 1.1 to 2.0.0-rc.16 (with patches) so this is the final missing doc that I will encounter.

## PR Info

This PR addresses the issue by adding missing #[doc = " Generated by sea-orm-macros"] to derived fields in sea-orm-macros.

## Bug Fixes

- [x] Enables crates using #[sea_orm::model] to use #[deny(missing_docs)].


tyt2y3:
thank you. I will add this to an example so CI can catch this in the future

PR #2790, opened 2025-11-01
feat(macros): Allow defining a relation without a foreign key constraint

DavidSkrundz:
I ran into an issue where I have two models that should be able to refer to each other without enforcing foreign key constraints.  This is the best solution I could come up with but I'm open to alternative implementations from someone more familiar with the code base.

## PR Info

In my upgrade efforts from 1.1 to 2.0 I hit the issue of unexpected foreign key violations.  When writing the relations and migrations by hand I was able to skip foreign keys that I did not want, but with `#[sea_orm::model]` I need another way.

This PR adds `#[sea_orm(skip_fk)]` for relation definitions that allow a relation to be created without any foreign key constraints.

## New Features

- [x] Allow `#[sea_orm(skip_fk)]` to skip creating foreign keys

## Example
I am trying to track `movies` that I have available, and a complete list of `watched_movies` including any movies that are currently not available.  This means that there cannot be a foreign key constraint on `watched_movies.id`.
```rust
mod entity {
    pub mod movies {

        #[sea_orm::model]
        #[derive(Debug, Clone, DeriveEntityModel)]
        #[sea_orm(table_name = "movies")]
        pub struct Model {
        #[sea_orm(primary_key)]
        pub id: u32,

        #[sea_orm(has_many, relation_enum = "Watched", from = "id", to = "id")]
        pub watched: HasMany<super::watched_movies::Entity>,
    }

}

pub mod watched_movies {

    #[sea_orm::model]
    #[derive(Debug, Clone, DeriveEntityModel)]
    #[sea_orm(table_name = "watched_movies")]
    pub struct Model {
        pub id: u32,
        pub user_id: u32,

        #[sea_orm(belongs_to, relation_enum = "Content", from = "id", to = "id", skip_fk)]
        pub movie: HasOne<super::movies::Entity>,
    }

}
```


tyt2y3:
umm interesting use case! yes I think we need the ability to omit foreign keys, let me think of a name for `skip_fk`.

in Django it's called `db_constraint=false`. may be I'll do `foreign_key = false`. I do recognize that it's quite long, so may be also `skip_fk` is fine.

DavidSkrundz:
I wasn't sure on the name, I just threw something together that worked and hopefully didn't break other things.   I was also debating having just a flag like I implemented or having a bool attached.  Let me know which direction you want to go and I can make some changes :)

tyt2y3:
let's keep `skip_fk` for now! I'll let you know if I changed it / added an alias haha.

I've added a test case in 6f850e9f678919f9cb6cb3358be3bae40676a2eb

FYI for the last PR for `extra` I added in 07a9ed1859d49572b114f7f83f356dffa9d60195

tyt2y3:
FYI, I've bumped a new rc release including your recent PRs

DavidSkrundz:
Wonderful, thanks for taking a look at my PRs! Don’t worry about breaking things if a better name or implementation presents itself. I upgraded to 2.0 RC for the new features and accept the risks :)

PR #2794, opened 2025-11-05
Typed Column

tyt2y3:
This is still WIP but I want everyone to see this earlier.

So basically after some research I figured we can have typed columns without the complexity of "one struct for every column", and this solution solves three things:

1. Make column methods type safe
2. No longer have to use CamelCase for column identifier
3. Make the shorthand for `cake::Entity` more useful

We would generate the following structs, columns share the same type so it's fine if Entity has 100 Columns:

```rust
pub struct TypedColumn {
    pub id: NumericColumn<Entity>,
    pub name: StringColumn<Entity>,
    pub price: NumericColumn<Entity>,
}

impl Entity {
    pub const COLUMN: TypedColumn = TypedColumn {
        id: NumericColumn(Column::Id),
        name: StringColumn(Column::Name),
        price: NumericColumn(Column::Price),
    };
}
```

```rust
// before
let cakes = cake::Entity::find()
    .filter(cake::Column::Name.like("Ch%"))
    .all(db)
    .await?;

// after
let cakes = Cake::find()
    .filter(Cake::COLUMN.name.like("Ch%"))
    .all(db)
    .await?;
```

The following would not compile:

```rust
Cake::find()
    .filter(Cake::COLUMN.id.like("Ch")) // method cannot be called on `NumericColumn<Entity>`
    .filter(Cake::COLUMN.name.like(2))  // the trait `From<{integer}>` is not implemented for `String`
    .order_by_desc(Cake::COLUMN.name)
    .one(db)
    .await?
```

This also means that `contains` for String would not collide with `array_contains`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,4 +113,4 @@ entity-registry = ["inventory", "sea-orm-macros/entity-registry"]
 # This allows us to develop using a local version of sea-query
 [patch.crates-io]
 # sea-query = { path = "../sea-query" }
-sea-query = { git = "https://github.com/SeaQL/sea-query", branch = "master" }
+# sea-query = { git = "https://github.com/SeaQL/sea-query", branch = "master" }


tyt2y3:
Everything will pass after I bump sea-query

tests/collection_tests.rs
```diff
@@ -163,6 +163,48 @@ pub async fn insert_collection(db: &DatabaseConnection) -> Result<(), DbErr> {
...
+
+    let found = Entity::find()
+        .filter(
+            Entity::COLUMN
+                .teas
+                .overlap(vec![Tea::BreakfastTea, Tea::AfternoonTea]),
+        )
+        .order_by_asc(Entity::COLUMN.id)
+        .all(db)
+        .await?;
```
tyt2y3, line 193: we can finally have this

tyt2y3:
release early, release often

Issue #2803, opened 2025-11-10
Have `MigratorTrait` use `self`


ctron:
## Motivation
Defining migrations, it is sometimes handy to have additional information, coming through configuration or other means of passing runtime information.

However, as `MigratorTrait`'s `migrations` function doesn't have a `self`, that is rather tricky.

## Proposed Solutions

Add `&self` to `MigratorTrait` functions. So this can be used, if required.

tyt2y3:
right, I see what you mean. however this is a backwards-incompatible change.
I'll see if I can think of some way to remain backwards compatible.
for now, you can read environment variables from within the `migrations` function, or use `lazy_static`

ctron:
> for now, you can read environment variables from within the `migrations` function, or use `lazy_static`

Yes, and that's what I'm doing. But this makes it very hard for tests. Because basically all of this is global. So you can go with tokio's local stuff, but that's all far from optimal.

And yes, I do understand it's a breaking change. Maybe there is a way. But maybe it's a topic for SeaORM 3.0.

tyt2y3:
we don't have to wait until 3.0 to fix this

PR #2806, opened 2025-11-11
MigratorTrait with self

tyt2y3:
#2803

### DIFF
--- a/sea-orm-migration/src/cli.rs
+++ b/sea-orm-migration/src/cli.rs
@@ -8,13 +8,13 @@ use tracing_subscriber::{EnvFilter, prelude::*};
 use sea_orm::{ConnectOptions, Database, DbConn, DbErr};
 use sea_orm_cli::{MigrateSubcommands, run_migrate_generate, run_migrate_init};
 
-use super::MigratorTrait;
+use super::MigratorTraitSelf;
 
 const MIGRATION_DIR: &str = "./";
 
 pub async fn run_cli<M>(migrator: M)
 where
-    M: MigratorTrait,
+    M: MigratorTraitSelf,
 {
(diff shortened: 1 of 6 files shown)


tyt2y3:
@ctron can you try this and see if it works for you?

```rust
use crate::common::migration::*;
use sea_orm_migration::{MigratorTraitSelf, prelude::*};

pub struct Migrator {
    pub i: i32,
}

#[async_trait::async_trait]
impl MigratorTraitSelf for Migrator {
    fn migrations(&self) -> Vec<Box<dyn MigrationTrait>> {
        vec![
            Box::new(m20220118_000001_create_cake_table::Migration),
            Box::new(m20220118_000002_create_fruit_table::Migration),
        ]
    }
}

```

```toml
[patch.crates-io]
sea-orm-migration = { git = "https://github.com/SeaQL/sea-orm", branch = "migrator-self" }
```

ctron:
I won't be able to actually try it out this week. But taking a quick look, that looks very promising. As the `MigrationTrait` already uses `self`, I should be able to pass in arguments to the actual migration. And so that should work!

tyt2y3:
thanks, please give it a try and let me know. want to have some user feedback before merging this in.

tyt2y3:
@ctron got any fun?

ctron:
> @ctron got any fun?

I'm so sorry. Especially as you came up with a quick fix like that. But I wasn't able to test it yet. It's still on my todo list, as this problem still persists and we do need a fix. So it still is relevant.

tyt2y3:
no problem, I've rebased the branch

ctron:
Finally had some time to check this out. Yes, having this would be awesome!

tyt2y3:
I hope there's no conflict!

PR #2825, opened 2025-11-26
Fix `exec_with_returning` does not cast enums

joleeee:
## PR Info
This fixes an issue where `RETURNING *` was executed instead of listing the appropriate fields. For instance enums does not work without this patch due to type errors.

- Closes 

- Dependencies:
  - 

- Dependents:
  - 

## New Features

- [ ] 

## Bug Fixes

- [ ] 

Used to generate this
```sql
DELETE FROM "rides" WHERE "rides"."id" = 654 RETURNING *
```

Now generates in this style
```sql
RETURNING
    "rides"."id",
    Cast("rides"."vehicle" AS "TEXT"),
    -- ...
```

## Breaking Changes

- [ ] 

## Changes

- [ ]

### DIFF
--- a/src/executor/delete.rs
+++ b/src/executor/delete.rs
@@ -154,7 +154,10 @@ where
     let db_backend = db.get_database_backend();
     match db.support_returning() {
         true => {
-            query.returning_all();
+            let returning = Query::returning().exprs(
+                E::Column::iter().map(|c| c.select_enum_as(c.into_returning_expr(db_backend))),
+            );
+            query.returning(returning);
             ReturningSelector::<SelectModel<<E>::Model>, _>::from_query(query)
                 .one(db)
                 .await


src/executor/delete.rs
```diff
@@ -154,7 +154,10 @@ where
     let db_backend = db.get_database_backend();
     match db.support_returning() {
```
sqlerrorthing, line 155: why not just `if`?

joleeee: I mean, I did not introduce this... I only changed 1 line and added 3

sqlerrorthing: if you here, refactor it?

joleeee: out of scope for a bugfix

Issue #2830, opened 2025-11-29
DeriveValueType for stringy newtypes with unnamed field


angrynode:
This is related to https://github.com/SeaQL/sea-orm/issues/2735 because implementing support for `Utf8PathBuf` would fix my usecase, but i think more generally supporting this would be useful.

Since last RC we can create some stringy newtypes from structs with named fields. However newtype structs (single unnamed field) fail to compile because the inner type does not implement `Nullable`/`ValueType` and other traits.

That's because `DeriveValueTypeStruct` has different behavior than `DeriveValueTypeString`. I would say it would make sense to be able to produce a `DeriveValueTypeString` whatever the underlying form (enum, named field struct, unnamed field struct) if `value_type = "String"`.

This doesn't work:

```rust
use camino::Utf8PathBuf;
use sea_orm::*;

#[derive(Clone, Debug, PartialEq, DeriveValueType)]
#[sea_orm(value_type = "String")]
#[sea_orm(from_str = "NewType::from_str", to_str = "NewType::to_string")]
struct NewType(Utf8PathBuf);

impl std::str::FromStr for NewType {
    type Err = std::convert::Infallible;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(Self(Utf8PathBuf::from(s)))
    }
}

impl std::fmt::Display for NewType {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        self.0.fmt(f)
    }
}
```

Fails to compile with:

```
error[E0277]: the trait bound `sea_orm::Value: From<Utf8PathBuf>` is not satisfied
 --> src/main.rs:4:35
  |
4 | #[derive(Clone, Debug, PartialEq, DeriveValueType)]
  |                                   ^^^^^^^^^^^^^^^ the trait `From<Utf8PathBuf>` is not implemented for `sea_orm::Value`
  |
  = help: the following other types implement trait `From<T>`:
            `sea_orm::Value` implements `From<&[u8]>`
            `sea_orm::Value` implements `From<&std::string::String>`
            `sea_orm::Value` implements `From<&str>`
            `sea_orm::Value` implements `From<Braced>`
            `sea_orm::Value` implements `From<ChronoUnixTimestamp>`
            `sea_orm::Value` implements `From<ChronoUnixTimestampMillis>`
            `sea_orm::Value` implements `From<Cow<'_, str>>`
[...]

PR #2831, opened 2025-11-29
Allow more stringy newtypes

angrynode:
See https://github.com/SeaQL/sea-orm/issues/2830

The core thing is in commit fddfe0c but i thought the two other commits were good additions. Feel free to disregard them for any reason :)

### DIFF
--- a/sea-orm-macros/tests/derive_value_type_test.rs
+++ b/sea-orm-macros/tests/derive_value_type_test.rs
@@ -11,3 +11,69 @@ fn when_user_alias_result_macro_still_works_test() {
     #[derive(sea_orm::DeriveValueType)]
     struct MyString(String);
 }
+
+#[test]
+fn when_stringy_newtype_works_test() {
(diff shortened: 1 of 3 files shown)


angrynode:
My bad for not running the tests. I added a new commit assuming when no `#[sea_orm(...)]` derive attributes are set, we're going for a `DeriveValueTypeStruct`.

I'm also tempted to add more tests.

tyt2y3:
> I'm also tempted to add more tests.

please do, this'll be appreciated!

can you add some more test cases here https://github.com/SeaQL/sea-orm/blob/master/tests/common/features/value_type.rs
and https://github.com/SeaQL/sea-orm/blob/master/tests/value_type_tests.rs#L105 ?

angrynode:
I added some tests, and also added a better error message when trying to use the derive with a tuple struct with more than 1 field.

tyt2y3 (approved):
thank you!

PR #2835, opened 2025-12-02
Reduce Vec allocations

reneklacan:


src/dynamic/execute.rs
```diff
@@ -97,7 +97,7 @@ where
         C: ConnectionTrait,
     {
         let rows = db.query_all(&self.query).await?;
-        let mut models = Vec::new();
+        let mut models = Vec::with_capacity(rows.len());
```
Huliiiiii: Why not just:
```rust
rows
    .into_iter()
    .map(|row| self.selector.from_raw_query_result(row))
    .try_collect()
```

tyt2y3 (commented):
thank you. I think either way is fine! as long as they are consistent

src/dynamic/execute.rs
reneklacan: `try_collect` is not stable yet https://github.com/rust-lang/rust/issues/94047

Huliiiiii: > `try_collect` is not stable yet [rust-lang/rust#94047](https://github.com/rust-lang/rust/issues/94047)

I mean Itertools::try_collect.

reneklacan: ah ok, updated! thanks

tyt2y3 (approved):
thank you both!

Issue #2827, opened 2025-11-28
Add ``ConnectionOrTransaction`` as a Unified Executor Type


uuushiro:
First of all, thank you for developing such a fantastic ORM. It has been a great experience using SeaORM.

I propose promoting the `SchemaManagerConnection` pattern (currently found in sea-orm-migration) to sea-orm core as a generalized `ConnectionOrTransaction<'c>` type.

This type serves as a "Unified Executor" that wraps either a reference to a DatabaseConnection or a DatabaseTransaction. It solves fundamental limitations regarding runtime dispatch and object safety, enabling flexible transaction management patterns that are currently impossible with impl ConnectionTrait.

## Motivation

In real-world applications (Services, Repositories, Test Helpers), we frequently encounter scenarios where a function needs to accept either a connection or a transaction dynamically.

### Key Use Cases
- Runtime Branching: Deciding at runtime (e.g., based on a config flag) whether to run an operation atomically (in a transaction) or directly (for performance/batching).
- Test Isolation: Reusing production repository logic in tests where a transaction is started in the test setup and automatically rolled back, without changing the repository code.
- Safe Boundaries: Allowing the caller to manage the transaction lifecycle while passing only "execution rights" to downstream functions.

### The Problem: Why existing solutions fail
Currently, there is no way to represent "Either Connection or Transaction" as a single type due to Rust's type system constraints

1. &impl ConnectionTrait (Generics) is Static
Generics are monomorphized at compile time. We cannot switch types based on runtime conditions.

```rust
// ❌ Compile Error: `if` and `else` have incompatible types
let db = if use_transaction {
    &txn  // &DatabaseTransaction
} else {
    &conn // &DatabaseConnection
};
```

2. &dyn TransactionTrait is NOT Object Safe
We cannot use trait objects because TransactionTrait contains generic methods (transaction<F, T, E>) and associated types. It is impossible to create a &dyn (ConnectionTrait + TransactionTrait).

## Proposed Solutions
Introduce the enum pattern into sea-orm core. This pattern has been proven stable in sea-orm-migration for years.

```rust
// sea-orm/src/database/connection_or_transaction.rs

pub enum ConnectionOrTransaction<'c> {
    Connection(&'c DatabaseConnection),
    Transaction(&'c DatabaseTransaction),
}

// Ergonomics
impl<'c> From<&'c DatabaseConnection> for ConnectionOrTransaction<'c> { ... }
[...]

PR #2839, opened 2025-12-05
feat: Add ``DatabaseExecutor`` unified executor type

uuushiro:
- Close https://github.com/SeaQL/sea-orm/issues/2827

## Summary

- Introduces `DatabaseExecutor<'c>` enum that wraps either a `&DatabaseConnection` or `&DatabaseTransaction`
- Implements `ConnectionTrait` and `TransactionTrait` via delegation pattern
- Adds `IntoDatabaseExecutor` trait for ergonomic function signatures
- Promotes the pattern from `sea-orm-migration`'s `SchemaManagerConnection` to core

## Motivation

In real-world applications, we frequently need functions that can accept either a connection or transaction dynamically. This type solves:

- **Runtime Branching**: Deciding at runtime whether to run atomically (in a transaction) or directly
- **Test Isolation**: Reusing production repository logic in tests where transactions are rolled back
- **Safe Boundaries**: Allowing callers to manage transaction lifecycle while passing only "execution rights"

## Usage Example

```rust
use sea_orm::{DatabaseExecutor, ConnectionTrait, TransactionTrait};

async fn save_user(db: &DatabaseExecutor<'_>, name: String) -> Result<(), DbErr> {
    user::ActiveModel { name: Set(name), ..Default::default() }
        .insert(db)
        .await?;
    Ok(())
}

// Works with both connection and transaction
save_user(&DatabaseExecutor::from(&conn), "Alice".into()).await?;

save_user(&DatabaseExecutor::from(&txn), "Bob".into()).await?;
txn.commit().await?;
```

### DIFF
--- a/sea-orm-migration/src/connection.rs
+++ b/sea-orm-migration/src/connection.rs
@@ -1,1 +1,4 @@
-pub use sea_orm::{DatabaseExecutor as SchemaManagerConnection, IntoDatabaseExecutor as IntoSchemaManagerConnection};
+pub use sea_orm::{
+    DatabaseExecutor as SchemaManagerConnection,
+    IntoDatabaseExecutor as IntoSchemaManagerConnection,
+};


Huliiiiii:
There is already one in sea-orm-migration, so you don't need to write it from scratch. You can move/re-export it.

https://github.com/SeaQL/sea-orm/blob/b9b236d19a40c71eb57fe9f43758731f74161fd3/sea-orm-migration/src/connection.rs#L9-L12

uuushiro:
@Huliiiiii 
Thanks for the pointer! That makes sense.

My plan is

1. Move the SchemaManagerConnection logic from sea-orm-migration into sea-orm core.
2. Rename it to ConnectionOrTransaction (since it's now general-purpose).
3. Refactor sea-orm-migration to use a type alias (pub type SchemaManagerConnection<'c> = sea_orm::ConnectionOrTransaction<'c>;) to preserve backward compatibility.

I will update the PR shortly.

Huliiiiii:
ConnectionOrTransaction is a bit verbose. How about renaming it to DatabaseExecutor?

uuushiro:
@Huliiiiii 
That sounds great! I actually prefer DatabaseExecutor as well.

uuushiro:
@Huliiiiii 
I have updated the PR.

I moved the main logic to sea-orm as DatabaseExecutor, but I implemented SchemaManagerConnection in sea-orm-migration as a Newtype wrapper (struct) instead of a simple type alias.

Reason
1. Dependency Hygiene: If I used a type alias, I would have to implement sea_schema::Connection inside sea-orm core. This would force sea-orm core to depend on sea-schema, causing longer build times for users who don't use migrations.
2. Separation of Concerns: Schema management logic (like get_schema_builder) belongs to the migration domain (DDL), not the core query execution domain (DML). The Newtype pattern allows us to keep this specific logic strictly within sea-orm-migration.

This approach keeps sea-orm core clean and lightweight while satisfying the Orphan Rule. However, if you have a better idea or prefer a different approach, please let me know.

cnwzhu:
什么时候能合并？

uuushiro:
@tyt2y3 When you have a moment, could you review this PR?
No rush—just waiting on the final approval. Thanks!

sea-orm-migration/src/connection.rs
tyt2y3, line 1: I think we can remove this from sea-orm-migration completely, and only re-export it? like 
```rust
pub use sea_orm::{IntoDatabaseExecutor as IntoSchemaManagerConnection, DatabaseExecutor as SchemaManagerConnection};
```

sea-orm-migration/src/connection.rs
```diff
@@ -104,45 +80,38 @@ impl TransactionTrait for SchemaManagerConnection<'_> {
...
+        self.0
+            .transaction_with_config(callback, isolation_level, access_mode)
+            .await
     }
 }
 
 #[cfg(feature = "sqlx-dep")]
 mod sea_schema_shim {
-    use super::{DatabaseConnection, DatabaseTransaction, SchemaManagerConnection};
+    use super::SchemaManagerConnection;
```
tyt2y3: we can move this to sea-orm, inside `sea_schema_shim.rs`

tyt2y3 (changes requested):
thank you for the PR!

I prefer to have only one type instead of having a second wrapper type.

> Separation of Concerns: Schema management logic (like get_schema_builder) belongs to the migration domain (DDL), not the core query execution domain (DML). The Newtype pattern allows us to keep this specific logic strictly within sea-orm-migration.

I agree, this is already achived in sea-orm with feature flags

> Dependency Hygiene: If I used a type alias, I would have to implement sea_schema::Connection inside sea-orm core. 
This would force sea-orm core to depend on sea-schema, causing longer build times for users who don't use migrations.

sea-schema is already a dependency of sea-orm, and I am tempted to merge it into sea-orm, then we don't have to have this shim. (don't worry, it's not happening right now)
this is gated behind the feature flag `schema-sync`. `sea-orm-migration` turn this flag on here https://github.com/SeaQL/sea-orm/blob/6e68ccd72f97548d92dc36fbcf219cb7db85ced1/sea-orm-migration/Cargo.toml#L27

uuushiro:
@tyt2y3  Thank you for the review! I have updated the PR based on your feedback.

tyt2y3:
Released!

uuushiro:
Thank you!

PR #2843, opened 2025-12-07
`1.1.x`: Add `exists` method to `PaginatorTrait` with tests

Expurple:
Rebase #2623 from `master` into `1.1.x`.

This is a nice backwards-compatible feature, I'd like to have it in `1.1.x` and adopt in advance


Expurple:
Clippy complaints come from the `1.1.x` base, ignore those

PR #2845, opened 2025-12-07
fix: allow LEFT JOIN to produce None for nested models

### LINKED ISSUE #2617: FromQueryResult fails when having nested Option<Model>
## Description
The story began with me trying to join 5 tables. As currently `find_with_related` is implemented for only three tables, I decided to fallback to manual joins.
In order to process the results, I made the following "result" structure:
```rust
#[derive(FromQueryResult)]
pub struct Joined {
  #[sea_orm(nested)]
  table1: entities::table1::Model,
  #[sea_orm(nested)]
  table2: Option<entities::table2::Model>
}
```

The important thing to note here that `table2` is join by LEFT JOIN, so it might be optional.
And in case it's optional, query will fail with `DbErr::Type("{column_name} is null") `.

This is because, first of all, `FromQueryResult` is implemented for `Entity` in such way, so that when `try_get` would return `TryGetErr` it's converted into `DbErr` by `?` operator and secondly, macro implements `from_query_result_nullable` just by calling `from_query_result`.
Therefore the implementation of `FromQueryResult` on `Option<T>` wouldn't catch the error, as it's implemented like this:
```rust
fn from_query_result_nullable(res: &QueryResult, pre: &str) -> Result<Self, TryGetError> {
        match T::from_query_result_nullable(res, pre) {
            Ok(v) => Ok(Some(v)),
            Err(TryGetError::Null(_)) => Ok(None),
            Err(err @ TryGetError::DbErr(_)) => Err(err),
        }
    }
```

So in the end, the error gets propagated back up as "database/programming error", when in reality it's incorrectly implemented macros.

## Steps to Reproduce
1. Left join two entities
2. Create structure to hold results, where "right table" model is wrapped in `Option`

### Expected Behavior
The optional field should be `None.`

### Actual Behavior
 query will fail with `DbErr::Type("{column_name} is null") `

### Reproduces How Often
Always.

### Workarounds
Manually implement FromQueryResult, where you explicitly call `from_query_result_optional` for `Optional<T>` fields.

## Versions
sea-orm = 1.1.12
rust = 1.87.0
macOS 15.5

denyszhak:
## PR Info

- Closes #2617 

## Bug Fixes

When using LEFT JOIN with a nested `Option<Model>`, code tried to read the missing
related row as if it were always present. This caused a type error instead of simply
returning `None`. 

This change makes nested optional models behave normal by treating
all-null fields as a missing record.

## Testing

Added test that validates the fix.

### DIFF
--- a/sea-orm-macros/src/derives/model.rs
+++ b/sea-orm-macros/src/derives/model.rs
@@ -3,6 +3,7 @@ use super::{
     util::{escape_rust_keyword, field_not_ignored, trim_starting_raw_identifier},
 };
 use heck::ToUpperCamelCase;
+use itertools::izip;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
 use std::iter::FromIterator;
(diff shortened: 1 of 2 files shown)


tyt2y3:
the test case failed, any ideas? 

```
thread 'from_query_result_left_join_with_optional_model_does_not_exist' (9460) panicked at tests/from_query_result_tests.rs:108:10:
succeeds to get the result: Type("A null value was encountered while decoding \"id\"")
```

my current understanding: if the parent object is `Option<Model>`, then it effectively made all children fields `Option<T>`.
but if a struct has 2 required fields, we can only construct a model if all 2 required fields are `Some`.
is my understanding correct?

denyszhak:
> the test case failed, any ideas?
> 
> ```
> thread 'from_query_result_left_join_with_optional_model_does_not_exist' (9460) panicked at tests/from_query_result_tests.rs:108:10:
> succeeds to get the result: Type("A null value was encountered while decoding \"id\"")
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> my current understanding: if the parent object is `Option<Model>`, then it effectively made all children fields `Option<T>`. but if a struct has 2 required fields, we can only construct a model if all 2 required fields are `Some`. is my understanding correct?

Looking into this

denyszhak:
@tyt2y3  
The query had two columns named id (from cake and bakery), so the lookup found the wrong one. I prefixed one of them so could you please rerun the tests?

As to your understanding - correct, that's what it does.

tyt2y3:
Awesome!

sea-orm-macros/src/derives/model.rs
```diff
@@ -103,28 +103,74 @@ impl DeriveModel {
...
             .iter()
-            .zip(&self.ignore_attrs)
-            .map(|(column_ident, ignore)| {
-                if *ignore {
-                    quote! {
-                        Default::default()
-                    }
+            .zip(column_idents)
+            .zip(field_types)
+            .zip(ignore_attrs)
```
Huliiiiii: It might be clearer if use izip here.

sea-orm-macros/src/derives/model.rs
```diff
@@ -103,28 +103,74 @@ impl DeriveModel {
...
+                    };
+                    (reader, unwrapper)
                 }
             })
-            .collect();
+            .unzip();
+
+        let all_null_check = field_idents
+            .iter()
+            .zip(ignore_attrs.iter().cloned())
```
Huliiiiii: Use copied.

sea-orm-macros/src/derives/model.rs
```diff
@@ -103,28 +103,74 @@ impl DeriveModel {
...
+            .filter(|(_, ignore)| !ignore)
+            .fold(quote! { true }, |acc, (field_ident, _)| {
+                quote! { #acc && #field_ident.is_none() }
+            });
```
Huliiiiii: This might be a bit clearer.
```
let all_null_check = {
    let checks: Vec<_> = izip(&field_idents, &ignore_attrs)
        .filter_map(|(field_ident, &ignore)| {
            if ignore {
                None
            } else {
                Some(quote! { #field_ident.is_none() })
            }
        })
        .collect();

    quote! { true #(&& #checks)* }
}
```

sea-orm-macros/src/derives/model.rs
```diff
@@ -103,28 +103,74 @@ impl DeriveModel {
...
 
         quote!(
             #[automatically_derived]
             impl sea_orm::FromQueryResult for #ident {
                 fn from_query_result(row: &sea_orm::QueryResult, pre: &str) -> std::result::Result<Self, sea_orm::DbErr> {
+                    Self::from_query_result_nullable(row, pre).map_err(|e| e.into())
```
Huliiiiii: Use Into::into

sea-orm-macros/src/derives/model.rs
```diff
@@ -103,28 +103,74 @@ impl DeriveModel {
...
+                            )
+                        ))?
+                    };
+                    (reader, unwrapper)
                 }
             })
-            .collect();
+            .unzip();
+
+        let all_null_check = field_idents
```
Huliiiiii: We need to add comments here to help readers understand the context.

Huliiiiii (changes requested):

sea-orm-macros/src/derives/model.rs
denyszhak: Done

I originally avoided izip! since itertools wasn’t used elsewhere in the crate. Given that you suggested it, I’m happy to change this.

sea-orm-macros/src/derives/model.rs
denyszhak: Done

sea-orm-macros/src/derives/model.rs
denyszhak: Done

sea-orm-macros/src/derives/model.rs
denyszhak: Please check and let me know if this comment is clear enough?\

sea-orm-macros/src/derives/model.rs
denyszhak: Nicee, done

PR #2847, opened 2025-12-10
Rusqlite Support

tyt2y3:
Almost, since sea-schema is still missing. But the core ORM works


tyt2y3:
Add SeaSchema support, now schema-sync also works!

PR #2848, opened 2025-12-10
Feat centralized configuration for sea-orm-cli

CCXLV:
## New Features

Added support for a `sea-orm.toml` configuration file.

```toml
[database]
url = { env = "DATABASE_URL" }

[migrations]
directory = "./migration"
```

`database.url` - Can be a full database URL or a reference using the env to load the value from an environment variable.
`migrations.directory` - Relative path to the migration directory, resolved from the location of the config file.

This configuration file replaces the need to pass `migration_dir` and `database_url` as command-line arguments to `sea-orm-cli`.
It makes the CLI easier to use by centralizing configuration in one place and also opens the door for future features that rely on consistent config.

## Changes

The `sea-orm-cli migrate` command no longer accepts the `migration_dir` and `database_url` arguments.
These are now fully handled through `sea-orm.toml`.

## Usage

If the config file is missing, the CLI will show an error explaining that the project is not configured yet, and prompt the user to run:
```bash
sea-orm-cli config init
```
This command will automatically create a `sea-orm.toml` file in the current directory, prefilled with template values.
Users may then adjust the values as needed.


sea-orm-cli/src/cli.rs
```diff
@@ -53,31 +53,22 @@ pub struct Cli {
...
+        display_order = 20
     )]
     Generate {
         #[command(subcommand)]
         command: GenerateSubcommands,
     },
-    #[command(about = "Migration related commands", display_order = 20)]
+    #[command(about = "Migration related commands", display_order = 30)]
     Migrate {
-        #[arg(
```
Huliiiiii: For compatibility, CLI flags still need to be retained.

Huliiiiii (changes requested):

sea-orm-cli/src/cli.rs
CCXLV: Kept the flags. One question: should we also respect their argument values if they’re passed?
Right now they’re ignored and only the config file values are used.

sea-orm-cli/src/cli.rs
```diff
@@ -417,23 +428,17 @@ pub async fn main() {
...
     match cli.command {
+        Commands::Config { command } => run_config_command(command).unwrap_or_else(handle_error),
         Commands::Generate { command } => {
             run_generate_command(command, verbose)
                 .await
                 .unwrap_or_else(handle_error);
         }
         Commands::Migrate {
-            migration_dir,
+            migration_dir: _,
```
Huliiiiii: CLI flags should override the configuration file, not be silently ignored.

Huliiiiii (changes requested):

sea-orm-cli/src/commands/config.rs
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+        Ok(config.database.url)
+    }
+}
+
+pub fn get_migration_dir() -> Result<String, Box<dyn Error>> {
+    let config = get_config()?;
+
+    let migration_dir = config.migrations.directory;
+    if migration_dir.starts_with(".") {
```
Huliiiiii: Use [Path::is_relative](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_relative)

sea-orm-cli/src/commands/config.rs
```diff
@@ -0,0 +1,95 @@
...
+        std::io::ErrorKind::NotFound,
+        "SeaORM config file not found, use `sea-orm-cli config init` to create one",
+    )))
+}
+
+fn run_config_init() -> Result<(), Box<dyn Error>> {
+    let config_path = Path::new("sea-orm.toml");
+    let config_template = include_str!("../../template/sea-orm.toml");
+
+    fs::write(config_path, config_template.to_string())?;
```
Huliiiiii: This will overwrite existing files. It should check for this and raise an error, or provide a `--force` flag.

sea-orm-cli/src/commands/config.rs
```diff
@@ -0,0 +1,95 @@
...
+pub fn get_migration_dir() -> Result<String, Box<dyn Error>> {
+    let config = get_config()?;
+
+    let migration_dir = config.migrations.directory;
+    if migration_dir.starts_with(".") {
+        let config_dir = get_config_dir()?;
+        let migration_dir = config_dir.join(migration_dir);
+        return Ok(migration_dir.to_string_lossy().to_string());
+    } else {
+        return Ok(migration_dir);
```
Huliiiiii: No need to explicitly return.

sea-orm-cli/src/commands/config.rs
Huliiiiii: By the way, you don't need `to_string`, `write` accepts `AsRef<[u8]>`.

sea-orm-cli/Cargo.toml
```diff
@@ -63,6 +63,8 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features =
     "fmt",
 ] }
 url = { version = "2.2", default-features = false }
+toml = "0.9.8"
```
Huliiiiii: I think we can disable the default `display` feature of toml, since we don’t need to serialize TOML.

Huliiiiii: I think we can disable the default display feature of TOML, since we don’t need to serialize TOML.

Huliiiiii: At the same time, it’s better to put them behind the `cli` feature flag, because this crate can also be used as a library.

sea-orm-cli/template/sea-orm.toml
```diff
@@ -0,0 +1,5 @@
+[database]
+url = "env:DATABASE_URL"       # Support env: prefix or direct URL
```
Huliiiiii: Maybe using an enum instead of manually parsing string would be better?
`url = "string_url"` and `url = { env = "env_var" }`

Huliiiiii (changes requested):

sea-orm-cli/src/commands/mod.rs
```diff
@@ -1,10 +1,12 @@
 use std::fmt::Display;
 
 #[cfg(feature = "codegen")]
```
Huliiiiii: This feature flag is used by the generate module.

Huliiiiii (changes requested):

Huliiiiii (approved):
Wow, that was fast. Thanks for the quick update!

PR #2849, opened 2025-12-11
Value array

Huliiiiii:
https://github.com/SeaQL/sea-query/pull/1004

### DIFF
--- a/src/database/mock.rs
+++ b/src/database/mock.rs
@@ -792,7 +792,7 @@ mod tests {
                 #[sea_orm(primary_key)]
                 pub id: i32,
                 pub integers: Vec<i32>,
-                pub integers_opt: Option<Vec<Option<i32>>>,
+                pub integers_opt: Option<Vec<i32>>,
             }
 
             #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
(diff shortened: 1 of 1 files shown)


tests/common/features/active_enum_vec.rs
```diff
@@ -7,7 +7,7 @@ use sea_orm::entity::prelude::*;
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: i32,
-    pub categories: Option<Vec<Category>>,
+    pub categories: Option<Vec<Option<Category>>>,
```
tyt2y3: is it possible for this to remain `categories: Option<Vec<Category>>,`?

tests/common/features/json_vec_derive.rs
```diff
@@ -31,7 +31,7 @@ pub mod json_struct_vec {
         #[sea_orm(primary_key)]
         pub id: i32,
         #[sea_orm(column_type = "JsonBinary")]
-        pub struct_vec: Vec<JsonColumn>,
+        pub struct_vec: sea_orm::JsonField<Vec<JsonColumn>>,
```
tyt2y3: same here, I hope a wrapper type is not strictly needed?

tyt2y3:
Nice! seems like we're very close to having no (minimal) breaking changes to the API?

tyt2y3:
seems like we only have to remove all patch.crates-io before/after merge

and remove bigdecimal from default features

tyt2y3:
we can presumably remove some `cast_as` for enums right? @Huliiiiii

Huliiiiii:
> we can presumably remove some `cast_as` for enums right? @Huliiiiii

Yes, but I want to implement those in a separate PR.

tyt2y3:
pushed all sea-query crates

PR #2855, opened 2025-12-16
Fix incorrect serde(rename) field mapping in ActiveModel::from_json

### LINKED ISSUE #2854: Serde `rename_all` attribute causes `ActiveModelTrait::from_json` to no longer default to `ActiveValue::NotSet`
## Description

This is a bug specifically created by #2842 which was made and merged for an issue I was having, so I want to start off by thanking @Huliiiiii for working on it so quickly! 

This bug is simple to describe: if a Model definition has a `#[serde(rename_all = ...)]` container attribute, then the behavior of `ActiveModelTrait::from_json` to default to `ActiveValue::NotSet` with missing fields stops working, and a `DbErr::Json("missing field")` error is thrown. This notably does _not_ happen for field-level rename attributes.

## Steps to Reproduce

```rust

#[sea_orm::model]
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, DeriveEntityModel)]
#[sea_orm(table_name = "records")]
pub struct Model {
    pub station_id: i32,
    #[serde(rename = "AT_F")]
    pub temp: Option<Decimal>
}
raw_awn_records::ActiveModel::from_json(serde_json::json!({
     "AT_F": 10.1,
}));

//Err(
//   Json(
//      "missing field `STATION_ID`",
//  ),
//)

```
Which should really be
```rust
ActiveModel {
    station_id: ActiveValue::NotSet,
    temp: ActiveValue::Set(10.1),
}
```

### Workarounds

Like I said, I already tested it using individual `#[serde(rename = ...)]` field attributes which do not produce the same error; those still default to `NotSet` as expected. For now you can just rely on those!

## Versions

SeaORM 2.0.0-rc21

Huliiiiii:
Closes #2854

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -482,7 +482,7 @@ pub trait ActiveModelTrait: Clone + Debug {
 
     /// Create ActiveModel from a JSON value
     #[cfg(feature = "with-json")]
-    fn from_json(mut json: serde_json::Value) -> Result<Self, DbErr>
+    fn from_json(json: serde_json::Value) -> Result<Self, DbErr>
     where
         Self: crate::TryIntoModel<<Self::Entity as EntityTrait>::Model>,
         <<Self as ActiveModelTrait>::Entity as EntityTrait>::Model: IntoActiveModel<Self>,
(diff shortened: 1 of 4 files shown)


src/entity/active_model.rs
```diff
@@ -465,7 +465,7 @@ pub trait ActiveModelTrait: Clone + Debug {
 
     /// Create ActiveModel from a JSON value
     #[cfg(feature = "with-json")]
-    fn from_json(mut json: serde_json::Value) -> Result<Self, DbErr>
+    fn from_json(json: serde_json::Value) -> Result<Self, DbErr>
```
Huliiiiii, line 485: A better solution is to implement from_json in the macro, which make code cleaner.

PR #2877, opened 2025-12-25
Release cli binaries

### LINKED ISSUE #520: Distribute `sea-orm-cli` by Binstall
[cargo-binstall](https://github.com/ryankurte/cargo-binstall) is a project letting prebuilt binaries to be downloaded instead of compiling the code locally like `cargo install` does. This would save a lot of time.

Huliiiiii:
Closes #520

### DIFF
--- a/.github/workflows/sea-orm-cli-release.yml
+++ b/.github/workflows/sea-orm-cli-release.yml
@@ -8,17 +8,28 @@ on:
 permissions:
   contents: write
 
+env:
+  CARGO_TERM_COLOR: always
+  RUSTC_WRAPPER: sccache
+  SCCACHE_GHA_ENABLED: true
+
 jobs:
   create-release:
     runs-on: ubuntu-latest
     steps:
-      - uses: softprops/action-gh-release@v2
-        with:
-          tag_name: ${{ github.ref_name }}
-          name: ${{ github.ref_name }}
-          generate_release_notes: true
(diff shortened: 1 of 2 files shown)


tyt2y3 (approved):
Reviewed — this is in good shape and I'm approving. The binstall metadata lines up exactly with the artifacts the workflow produces (`pkg-url`/`bin-dir` resolve to the `sea-orm-cli-<target>-v<version>.tgz` → `…/sea-orm-cli` paths the build emits, Windows `.zip` override included), the draft→upload→publish flow is idempotent and re-runnable, and the size-optimized `[profile.release]` is honored because `sea-orm-cli` is its own workspace.

One follow-up I'll handle on `master` after merge rather than block on — the release-process docs need to catch up with the new tag convention:

- The workflow fires on `sea-orm-cli@*`, but `build-tools/RELEASE.md` step 8 only creates plain `2.0.0-rc.N` tags, so nothing triggers this until we add a step to push a `sea-orm-cli@<version>` tag.
- `--notes-from-tag` reads an *annotated* tag's message, but our documented release tags are lightweight — so the cli release tag should be created annotated (`git tag -a`), otherwise the notes come out empty. (Alternatively switch to `--generate-notes` or a `--notes-file` pointing at `changelog/*.md`.)

Merging now; will add the RELEASE.md step in a follow-up.

PR #2885, opened 2025-12-30
feat: Add tracing spans for database operations

### LINKED ISSUE #2867: Official support for span-based tracing
First of all, thank you for all your work maintaining SeaORM.

## Motivation 
[SeaORM’s official documentation](https://www.sea-ql.org/SeaORM/docs/install-and-config/debug-log/) currently focuses on debug logging (enabling SQL logs via RUST_LOG, etc.). This is helpful for local debugging, but it’s mostly logging-oriented, and it doesn’t provide strong support for span-based tracing (e.g., parent/child spans under an HTTP request span, standardized DB span attributes, timing, error status, etc.). 

In practice, many teams want distributed tracing style visibility: “this HTTP request → these DB operations → which query was slow / failed”. Logs alone make this much harder.Spans provide hierarchy and timing (“which DB calls happened under this request?”), which is difficult to achieve with logging-only approaches.

There is clear ecosystem interest in OpenTelemetry-compatible spans
- [sqlx-tracing](https://github.com/jdrouet/sqlx-tracing) instruments SQLx operations by wrapping pools/executors and emitting OpenTelemetry-style span fields. 
- [sea-orm-tracing](https://github.com/benbpyle/sea-orm-tracing) (community crate) implements a traced wrapper around DatabaseConnection / ConnectionTrait and emphasizes automatic nesting of DB spans under the current active span (HTTP middleware, etc.).  

These examples show demand, but the lack of an official hook forces users into “external wrapper” strategies or ad-hoc solutions.

## Proposed Solutions

Introduce a small, optional instrumentation hook in SeaORM core that can be used to create spans around DB operations.

### Concept
An ObservabilityHook (name bikesheddable) that is invoked around query execution paths
- called for execute, query_one, query_all, stream, execute_unprepared
- also used in transaction flows (begin/commit/rollback where applicable)

This hook would allow SeaORM to remain backend-agnostic while enabling
- span creation with correct parent/child relationships
- error recording
- duration measurement
- [standardized attributes following OpenTelemetry DB semantic conventions](https://opentelemetry.io/docs/specs/semconv/database/)

### Design goals
- Opt-in / zero overhead by default (no hook → no extra allocations/spans)
- Minimal API surface (don’t bake OpenTelemetry types into SeaORM; just enable tracing integration)
- Works for existing backends (Postgres/MySQL/SQLite)

### Related: Proxy / extension points

I saw recent work improving proxy-related behavior (e.g., [ProxyDatabase support in transaction impl](https://github.com/SeaQL/sea-orm/pull/2573)).

That’s great progress—separately, it would be helpful to discuss whether proxy/hook patterns could become a stable extension point for observability (not only testing/mocking).

## Questions for maintainers
Does the overall direction make sense to you — i.e., strengthening SeaORM’s span-based tracing instrumentation (OpenTelemetry-friendly) in addition to the existing debug logging? If you agree with the goal, what do you think is the best approach in SeaORM’s architecture?

A lightweight hook/middleware around execution paths (driver / connection / transaction), orSome other extension point you’d recommend?

uuushiro:
## Summary

This PR adds optional tracing instrumentation for database operations, enabling distributed tracing and observability.

- Introduces `tracing-spans` feature flag for opt-in tracing support
- Creates spans for all database operations (`execute`, `query_one`, `query_all`) and transaction flows (`begin`, `commit`, `rollback`)
- Records OpenTelemetry-compatible semantic attributes (`db.system`, `db.operation`, `db.statement`, `otel.status_code`)
- Uses `.instrument(span)` pattern for proper async span context propagation

## Motivation

As noted in #2867, logs alone make it difficult to understand database call hierarchies and timing within request contexts. Spans provide:

- **Hierarchy**: Which DB calls happened under which request
- **Timing**: Precise duration of each operation
- **Context**: Error status and query details for debugging

## Design Decision

The original issue proposed an `ObservabilityHook` pattern. This PR takes a simpler approach by using the `tracing` crate directly:

1. **Standard ecosystem**: `tracing` is the de facto standard for Rust instrumentation
2. **Simpler integration**: Users just enable the feature and configure their tracing subscriber
3. **Backend agnostic**: Works with any tracing-compatible backend (Jaeger, Datadog, OpenTelemetry, etc.)
4. **SQL safety**: Relies on tracing backends (e.g., Datadog Agent) for SQL obfuscation rather than implementing custom filtering

## Usage

```toml
[dependencies]
sea-orm = { version = "2.0", features = ["tracing-spans"] }
```

```rust
// Set up any tracing subscriber
tracing_subscriber::fmt()
    .with_max_level(tracing::Level::INFO)
    .init();

// All database operations automatically generate spans
let db = Database::connect("postgres://...").await?;
let users = User::find().all(&db).await?;  // Generates span with db.* attributes
```

## Span Attributes

| Attribute | Description | Example |
|-----------|-------------|---------|
| `db.system` | Database type | `postgresql`, `mysql`, `sqlite` |
| `db.operation` | SQL operation | `SELECT`, `INSERT`, `UPDATE`, `DELETE` |
| `db.statement` | SQL query | `SELECT * FROM users WHERE id = $1` |
| `otel.status_code` | Result status | `OK`, `ERROR` |
| `exception.message` | Error details | (on failure) |

Closes #2867

### DIFF
--- a/src/database/db_connection.rs
+++ b/src/database/db_connection.rs
@@ -4,8 +4,6 @@ use crate::{
     TransactionTrait, error::*,
 };
 use std::{fmt::Debug, future::Future, pin::Pin};
-#[cfg(feature = "tracing-spans")]
-use tracing::Instrument;
 use tracing::instrument;
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


uuushiro:
https://github.com/SeaQL/sea-orm/issues/2867#issuecomment-3699528880

src/database/db_connection.rs
```diff
@@ -144,95 +146,178 @@ impl ConnectionTrait for DatabaseConnection {
...
+                #[cfg(feature = "proxy")]
+                DatabaseConnectionType::ProxyDatabaseConnection(conn) => conn.query_all(stmt).await,
+                DatabaseConnectionType::Disconnected => Err(conn_err("Disconnected")),
+            }
+        };
+
+        #[cfg(feature = "tracing-spans")]
+        let result = fut.instrument(span.clone()).await;
+        #[cfg(not(feature = "tracing-spans"))]
+        let result = fut.await;
```
tyt2y3: thanks you for the PR. is it possible to reduce the number of feature guarded statements using macros?

uuushiro: @tyt2y3 Thanks for the suggestion — I refactored the tracing-spans instrumentation into a single with_db_span! macro.

PR #2891, opened 2026-01-03
Add method to create Models from only required fields

### LINKED ISSUE #2834: `AttrNotSet` error when using `.try_into_model()` even though the attribute listed is optional
## Description

When creating models as mock inputs for testing purposes, I am creating an ActiveModel with only the non-nullable fields filled out and using default to make the rest `NotSet`, then using `.try_into_model()` to convert that into a model (which is required by my function). When I run the test, I get an `AttrNotSet` error complaining that I haven't set the optional fields, even though these should default to `None`.

#### Model definition:
```rust
#[sea_orm::model]
#[sea_orm(table_name = "stations_creston")]
pub struct Model {
    pub station_id: i32,
    #[sea_orm(unique)]
    pub network_id: String,
    pub station_name: String,
    pub installation_date: Option<TimeDate>,
}
```
It should also be noted that installation_date is set to nullable in my MySQL database, and the default value is set to NULL
#### Test setup
```rust
let test_station = stations_creston::ActiveModel {
    station_id: ActiveValue::Set(628),
    network_id: ActiveValue::Set("CRESNEKN".to_owned()),
    station_name: ActiveValue::Set("example".to_owned()),
    ..Default::default()
}
.try_into_model()
.expect("couldn't make test station model");
```

### Expected Behavior

This shouldn't panic

### Actual Behavior
The following panic is produced:
```
thread 'apis::creston::test::get_data' panicked at src/apis/creston.rs:181:10:
couldn't make test station model: AttrNotSet("installation_date")
```

### Reproduces How Often

Happens every time

## Versions

SeaOrm 2.0.0-rc.18
cargo 1.93.0-nightly
Void Linux

PacificBird:
## PR Info

Closes #2834. As it stands right now this is very much a draft. I am unable to execute my unit test as running `cargo test --workspace` fails to compile, although `cargo check` finds no errors.

To-do:

- [ ] Figure out how to get tests to compile
- [ ] Clean up the macro implementation
- [ ] Workshop the method name
- [ ] Discuss if we want to split this up into an optional derive macro on Model instead of being derived by default in ModelEx.

## Breaking Changes

- [ ] A method `model_from_required` is added to `ModelEx` in the respective derive macro.

### DIFF
--- a/sea-orm-macros/src/derives/model_ex.rs
+++ b/sea-orm-macros/src/derives/model_ex.rs
@@ -50,13 +50,19 @@ pub fn expand_sea_orm_model(input: ItemStruct, compact: bool) -> syn::Result<Tok
     };
 
     let mut model_fields = Vec::new();
+    let mut required_model_fields = Vec::new();
 
     for field in all_fields.iter_mut() {
         let field_type = &field.ty;
         let field_type = quote! { #field_type }
             .to_string() // e.g.: "Option < String >"
             .replace(' ', ""); // Remove spaces
-
+        if !(field_type.starts_with("Option<")
+            || field_type.starts_with("HasMany<")
+            || field_type.starts_with("HasOne<"))
+        {
+            required_model_fields.push(field.clone());
+        }
         if is_compound_field(&field_type) {
             let entity_path = extract_compound_entity(&field_type);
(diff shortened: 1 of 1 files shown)


tyt2y3:
sorry for the trouble running tests, I've added some instructions in 51dd40b0d126d4745f5429b7be247e44941429de

Issue #2899, opened 2026-01-06
Documentation bug for SeaORM 2.0


tcanabrava:
The documentation, for the EntityFirst approach, states that the schema_registry should be done with a `::*` at the end, but this breaks registration:

```
// synchronizes database schema with entity definitions
db.get_schema_registry("my_crate::entity::*").sync(db).await?;
```

But the code that *does* the registration uses .starts_with:

```
    for entity in inventory::iter::<EntityRegistry>() {
        if entity.module_path.starts_with(module_path) {
```

it took me a while to figure out why nothing was working and I had to dig deep in code.
There are two fixes that might be worh.:

1 - fix the documentation (and blogposts) to remote the `::*`
2 - add a  print to stderr saying that the module path is unexpected, to help the development process.

tyt2y3:
`crate::mod::*` is actually supported, because the `*` will be removed from the pattern before matching.
not sure what's the issue with you specific case, may be the module path is different?

PR #2900, opened 2026-01-06
Add output for when an entity registration fails

tcanabrava:
Had some problems following the documentation for the schema first approach because the documentation doesn't follow what the code does. and because there is no output on failure, I had to dig into the sea-orm codebase, to understand the issue.

This is a simple patch, but it does help the developer that did a typo, or used a wrong module for some reason on registration.

improves https://github.com/SeaQL/sea-orm/issues/2899

### DIFF
--- a/src/entity/registry.rs
+++ b/src/entity/registry.rs
@@ -37,7 +37,11 @@ impl EntityRegistry {
             if entity.module_path.starts_with(prefix) {
                 schema.register_entity((entity.schema_info)(schema.helper()));
             } else {
-                tracing::warn!("Entity module {} does not start with {}",entity.module_path, prefix);
+                tracing::warn!(
+                    "Entity module {} does not start with {}",
+                    entity.module_path,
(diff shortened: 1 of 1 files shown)


src/entity/registry.rs
```diff
@@ -36,6 +36,8 @@ impl EntityRegistry {
         for entity in inventory::iter::<crate::EntityRegistry>() {
             if entity.module_path.starts_with(prefix) {
                 schema.register_entity((entity.schema_info)(schema.helper()));
+            } else {
+                println!("Entity module {} does not start with {}",entity.module_path, prefix);
```
Huliiiiii: Would it be better to use `tracing::warn` here?

tcanabrava: sure, will update shortly

Huliiiiii (approved):

tyt2y3:
Thank you. I've added some debug logs in another commit

Issue #2904, opened 2026-01-08
Enum variant is not a valid Rust identifier


a4n1:
## Description

Code gen panics when dealing with Rust identifiers that start with a numeric character and invalid Rust identifier characters, such as `0-foo`.

## Steps to Reproduce

`cargo run -- generate entity -u postgres://...`

### Expected Behavior

Code gen should succeed

### Actual Behavior

Panics with the following error message
```
thread 'main' (4589437) panicked at .../sea-orm-cli-1.1.17-vendor/quote-1.0.41/src/runtime.rs:455:9:
"_0-foo" is not a valid Ident
```

### Reproduces How Often

Always

### Workarounds

None

## Versions

```
sea-orm-cli v2.0.0-rc.27 (/Users/agni/development/open/sea-orm/sea-orm-cli)
├── sea-orm-codegen v2.0.0-rc.27 (/Users/agni/development/open/sea-orm/sea-orm-codegen)
│   ├── sea-query v1.0.0-rc.29
│   │   └── sea-query-derive v1.0.0-rc.11 (proc-macro)
├── sea-schema v0.17.0-rc.17
│   ├── sea-query v1.0.0-rc.29 (*)
│   ├── sea-query-sqlx v0.8.0-rc.11
│   │   ├── sea-query v1.0.0-rc.29 (*)
│   ├── sea-schema-derive v0.3.0 (proc-macro)
```

PR #2905, opened 2026-01-08
Fix renaming enum variants starting with digits

a4n1:
## PR Info

- Closes https://github.com/SeaQL/sea-orm/issues/2904

## Bug Fixes

- [ ] Fixes enum variants that start with a digit and then have characters which are not valid in Rust identifiers

### DIFF
--- a/sea-orm-codegen/src/entity/active_enum.rs
+++ b/sea-orm-codegen/src/entity/active_enum.rs
@@ -25,7 +25,7 @@ impl ActiveEnum {
         let values: Vec<String> = self.values.iter().map(|v| v.to_string()).collect();
 
         let variants = values.iter().map(|v| v.trim()).map(|v| {
-            if v.is_empty() && !v.to_upper_camel_case().is_empty() {
+            if v.is_empty() {
                 println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
                 return format_ident!("__EmptyString");
             }
@@ -104,6 +104,36 @@ mod tests {
     use crate::entity::writer::{bonus_attributes, bonus_derive};
     use pretty_assertions::assert_eq;
     use sea_query::{Alias, IntoIden};
+
+    #[test]
+    fn test_enum_variant_starts_with_empty_string() {
+        assert_eq!(
+            ActiveEnum {
+                enum_name: Alias::new("media_type").into_iden(),
+                values: vec![""]
+                    .into_iter()
+                    .map(|variant| Alias::new(variant).into_iden())
+                    .collect(),
+            }
+            .impl_active_enum(
+                &WithSerde::None,
+                true,
+                &TokenStream::new(),
+                &TokenStream::new(),
+                EntityFormat::Compact,
+            )
+            .to_string(),
+            quote!(
+                #[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Copy)]
+                #[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "media_type")]
+                pub enum MediaType {
+                    #[sea_orm(string_value = "")]
+                    __EmptyString,
+                }
+            )
+            .to_string()
+        )
+    }
 
     #[test]
     fn test_enum_variant_starts_with_number() {


sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -25,29 +25,32 @@ impl ActiveEnum {
...
-            if v.chars().next().map(char::is_numeric).unwrap_or(false) {
-                format_ident!("_{}", v)
-            } else {
-                let variant_name = if v.is_empty() {
-                    println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
-                    "__EmptyString".to_string()
+            let variant_name = if v.is_empty() {
+                println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
+                "__EmptyString".to_string()
+            } else if v.chars().next().map(char::is_numeric).unwrap_or(false) {
```
Huliiiiii: Use `Option::map_or_default`

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -25,29 +25,32 @@ impl ActiveEnum {
...
+                if v.chars().any(|c| !c.is_numeric() && !c.is_alphabetic() && c != '_' && c != '-' && c != ' ') {
+                    "".to_string()
```
Huliiiiii: I think this should be moved to __EmptyString and update the warning message.

a4n1: Thanks for the review! Yeah this is a confusing case. `to_upper_camel_case` removes invalid chars, so a field like "/" is remapped to "". In order to preserve existing behavior for variants starting with a digit, we have to strip chars manually if they are not valid. 

Empty strings are then converted to UTF-8 encoding in the next block. So, "0/" is remapped to "U0030U002F". I agree that the flow is pretty hard to follow now though.

We could replace this check with `format!("_{}", v.to_upper_camel_case())`. However, this would map `0A-B-C` to `_0aBC` instead of `_0ABC`.

Huliiiiii: Oh, that's really confusing. Could you refactor this to make the control flow clearer?

Huliiiiii: Or add a comment explaining that Unicode characters will be handled below.

a4n1: I've refactored the control flow, hopefully it's easier to follow now

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -24,31 +24,34 @@ impl ActiveEnum {
...
-                            write!(&mut ss, "U{:04X}", c as u32).unwrap();
-                        }
+            if v.is_empty() && !v.to_upper_camel_case().is_empty() {
+                println!("Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed.");
+                return format_ident!("__EmptyString");
+            }
+
+            if v
+                .chars()
+                .any(|c| !c.is_alphanumeric() && c != '_' && c != '-' && c != ' ') {
```
Huliiiiii: Use matches macro

Huliiiiii:
My idea is:
```rust
if v.is_empty() {
    println!(
        "Warning: item in the enumeration '{enum_name}' is an empty string, it will be converted to `__EmptyString`. You can modify it later as needed."
    );
    return format_ident!("__EmptyString");
}

let mut buf = String::new();

let mut print_warn = false;

let is_separator_only = v.chars().all(|c| matches!(c, '_' | '-' | ' '));

for char in v.chars() {
    if char.is_ascii_alphanumeric() {
        buf.push(char);
        continue;
    }

    print_warn = true;

    if !is_separator_only && matches!(char, '_' | '-' | ' ') {
        continue;
    } else {
        write!(&mut buf, "U{:04X}", char as u32).unwrap();
    }
}


if print_warn {
    println!(
        "Warning: item '{v}' in the enumeration '{enum_name}' cannot be converted into a valid Rust enum member name. It will be converted to its corresponding UTF-8 encoding. You can modify it later as needed."
    );
}

let variant_name = buf.to_upper_camel_case();

if v.chars().next().is_some_and(char::is_ascii_digit) {
    format_ident!("_{variant_name}")
} else {
    format_ident!("{variant_name}")
}
```

This should be clearer and prevent numbers from being formatted as Unicode.

a4n1:
This would cause breaking changes as we would be converting variant names to upper camel case if they start with a digit. The previous implementation preserved casing in this case. Also unicode values would be converted to upper camel case (`0U0030 → 0u0030`) which feels undesirable. I do agree that we should not convert numbers to unicode though.

I think we need separate behavior for the following cases:
1. Empty string 
→ return "__EmptyString"
2. Contains invalid characters (anything other than alphanumeric, -, _, space)
→ encode the invalid characters while preserving the original casing, separators (-, space) should be ignored
3. Otherwise
→ convert the value to upper camel case

Additionally, if the resulting value starts with a digit we should prepend with an underscore.

The latest change reflects this, although I'm happy to defer to your suggestion if you're okay with the caveats.

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -31,24 +30,35 @@ impl ActiveEnum {
                 return format_ident!("__EmptyString");
             }
 
-            if v
-                .chars()
-                .any(|c| !c.is_alphanumeric() && !matches!(c, '_' | '-' | ' ')) {
+            let is_leading_digit = v.chars().next().is_some_and(char::is_numeric);
```
Huliiiiii: Should use `is_ascii_digit`.
https://doc.rust-lang.org/std/primitive.char.html#method.is_numeric

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -31,24 +30,35 @@ impl ActiveEnum {
...
+            let is_valid_char = |c: char| c.is_ascii_alphanumeric() || c == '_';
+            let is_passthrough = |c: char| matches!(c, '-' | ' ');
+
+            if v.chars().any(|c| !is_valid_char(c) && !is_passthrough(c)) {
```
Huliiiiii: I think extracting this condition into a variable would make the intent clearer.

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -31,24 +30,35 @@ impl ActiveEnum {
...
+                    } else if is_valid_char(c) || c.len_utf8() > 1 {
+                        buf.push(c);
                     } else {
-                        write!(&mut ss, "U{:04X}", c as u32).unwrap();
+                        buf.push_str(&format!("U{:04X}", c as u32));
                     }
                 }
-                return format_ident!("{}", ss);
+
+                return format_ident!("{buf}");
```
Huliiiiii, line 57: This is just a personal preference, but I think using else if instead of early returns better indicates that these as sibling branches.

sea-orm-codegen/src/entity/active_enum.rs
a4n1: Good to know!

Huliiiiii:
Thanks!

sea-orm-codegen/src/entity/active_enum.rs
```diff
@@ -24,30 +23,45 @@ impl ActiveEnum {
...
-                if variant_name.is_empty() {
-                    println!("Warning: item '{v}' in the enumeration '{enum_name}' cannot be converted into a valid Rust enum member name. It will be converted to its corresponding UTF-8 encoding. You can modify it later as needed.");
-                    let mut ss = String::new();
-                    for c in v.chars() {
-                        if c.len_utf8() > 1 {
-                            write!(&mut ss, "{c}").unwrap();
-                        } else {
-                            write!(&mut ss, "U{:04X}", c as u32).unwrap();
-                        }
+            if v.is_empty() && !v.to_upper_camel_case().is_empty() {
```
Huliiiiii: I think it would be better to add a comment explaining why `!v.to_upper_camel_case().is_empty()` is needed.

a4n1: Ah this actually introduces a regression. I've added a test and removed the condition.

PR #2909, opened 2026-01-12
Optimize exists

### LINKED ISSUE #2908: Redundant query in PaginatorTrait::exists()
## Description

PaginatorTrait::exists() generates redundant SQL statements that may lead to performance degradation.

## Steps to Reproduce

Execute a query by the following function:
```rust
async fn exists_by_username(db: DatabaseConnection, username: &str) -> Result<bool, DbErr> {
    User::find()
        .filter(user::Column::Username.eq(username))
        .exists(&db)
        .await
}
```

### Expected Behavior

It should have generated the following SQL statement:
```sql
SELECT 1 
FROM "user" 
WHERE "username" = '...' 
LIMIT 1
```

### Actual Behavior

According to `tracing` logs:
```
DEBUG sea_orm::driver::sqlx_sqlite: SELECT 1 FROM (SELECT " user"."id", " user"."username", " user"."password" ...... FROM " user" WHERE " user"."username" = '...' LIMIT 1) AS "sub_query" LIMIT 1
INFO sqlx::query: summary="PRAGMA foreign_keys = ON; …" db.statement="\n\nPRAGMA foreign_keys = ON; \n" rows_affected=0 rows_returned=0 elapsed=54.3µs elapsed_secs=5.43e-5      
INFO sqlx::query: summary="SELECT 1 FROM (SELECT …" db.statement="\n\nSELECT 1 FROM (SELECT \" user\".\"id\", \" user\".\"username\", \"user\".\"password\"...... FROM \" user\" WHERE \" user\".\"username\" = ? LIMIT ?) AS \"sub_query\" LIMIT ?\n" rows_affected=0 rows_returned=1 elapsed=1.0797ms elapsed_secs=0.0010797
```

### Reproduces How Often

Always reproducible.

### Workarounds

None.

## Versions

2.0.0-rc.28

Huliiiiii:
- Closes #2908

### DIFF
--- a/src/executor/paginator.rs
+++ b/src/executor/paginator.rs
@@ -275,31 +275,6 @@ where
     {
         self.paginate(db, 1).num_items().await
     }
-
-    /// Check if any records exist
-    async fn exists(self, db: &'db C) -> Result<bool, DbErr>
-    where
-        Self: Send + Sized,
-    {
-        let paginator = self.paginate(db, 1);
-        let stmt = SelectStatement::new()
-            .expr(Expr::cust("1"))
-            .from_subquery(
-                paginator
-                    .query
-                    .clone()
-                    .reset_limit()
-                    .reset_offset()
-                    .clear_order_by()
-                    .limit(1)
-                    .to_owned(),
-                "sub_query",
-            )
-            .limit(1)
-            .to_owned();
-        let result = db.query_one(&stmt).await?;
-        Ok(result.is_some())
-    }
 }
 
 impl<'db, C, S> PaginatorTrait<'db, C> for Selector<S>
@@ -400,6 +375,7 @@ mod tests {
     #[cfg(feature = "sync")]
     use crate::util::StreamShim;
     use crate::{DatabaseConnection, DbBackend, MockDatabase, Transaction};
+    use crate::{QueryOrder, QuerySelect};
     use crate::{Statement, tests_cfg::*};
     use futures_util::{TryStreamExt, stream::TryNext};
     use pretty_assertions::assert_eq;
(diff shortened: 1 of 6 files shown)


tyt2y3 (approved):
do we need to add a test case or there is already coverage?

tyt2y3 (approved):

src/executor/select_ext.rs
```diff
@@ -0,0 +1,448 @@
...
+impl<S> SelectExt for Selector<S>
+where
+    S: SelectorTrait,
+{
+    fn exists_query(self) -> SelectStatement {
+        into_exists_query(self.query)
+    }
+}
+
+#[async_trait::async_trait]
```
tyt2y3, line 45:
```suggestion
```

src/executor/paginator.rs
```diff
@@ -275,31 +275,6 @@ where
     {
         self.paginate(db, 1).num_items().await
     }
-
-    /// Check if any records exist
-    async fn exists(self, db: &'db C) -> Result<bool, DbErr>
-    where
-        Self: Send + Sized,
```
tyt2y3, line 282: I am thinking whether we should keep this, at least leave behind a `deprecated` notice?

Huliiiiii: It was added in v2.0, so I think we can move it directly.

src/executor/select_ext.rs
Huliiiiii: async fn in trait can’t mark the future as Send and Sync, but we can use impl future instead. Which do you prefer?

src/executor/paginator.rs
tyt2y3: haha, didn't even remember!

PR #2913, opened 2026-01-13
Filter relations outside generation set

### LINKED ISSUE #2911: sea-orm-cli generates invalid belongs_to when `--tables` specified
## Description

 When generating entities with SeaORM CLI v2 (rc) using `--tables` to restrict output to a subset of tables, the generator still emits relations that reference tables outside the selected set. In modularized projects, this produces code that refers to `super::<other_table>::Entity` which is not present in the module, causing compilation failures.

## Steps to Reproduce

1. create tables with this SQL

```sql
create table parent ( id serial not null );
create table child ( id serial not null, parent_id int references parent(id) );
```

2. `sea-orm-cli generate entity` with specify `--tables parent`

### Expected Behavior

under `Relation` enum  `#[sea_orm(belongs_to ...)]` filtered with `--tables`.

### Actual Behavior

 `#[sea_orm(belongs_to ...)]` generated with FK defined, 

### Reproduces How Often

 Always

### Workarounds

 None

## Versions

```
$ sea-orm-cli --version
sea-orm-cli 2.0.0-rc.28
```

Huliiiiii:
Closes #2911

### DIFF
--- a/sea-orm-codegen/src/entity/transformer.rs
+++ b/sea-orm-codegen/src/entity/transformer.rs
@@ -3,7 +3,7 @@ use crate::{
     RelationType,
 };
 use sea_query::TableCreateStatement;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 #[derive(Clone, Debug)]
 pub struct EntityTransformer;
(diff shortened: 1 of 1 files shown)


tyt2y3 (approved):

tyt2y3:
thank you!

Issue #2717, opened 2025-09-02
Support storing Uuids as SQLite TEXT columns


raffomania:
## Motivation

Storing Uuids as `TEXT` columns in SQLite has two advantages:

- Easier to debug the DB using the `sqlite3` command directly
- Allows for queries like `WHERE id IN (<uuid>, <uuid>, ...)` which are hard (maybe impossible?) to write with `BLOB` values

## Proposed Solutions

I have multiple ideas for solving this:

- [sqlx will store `fmt` Uuid types such as `Hyphenated` as `TEXT`](https://github.com/launchbadge/sqlx/blob/69bb5952ab665f6edfb461b45e63cc3b6d99a4d0/sqlx-sqlite/src/types/uuid.rs#L47). SeaORM already supports using these types for column definitions, but still stores them as `BLOB` (tested with version 1.1.15). SeaORM could store them as text instead.
- Allow overriding the column type of any Uuid column:
    ```rs
    #[sea_orm(column_type = "Text")]
    pub id: Uuid
    ```
    On 1.1.15 this compiles, but still tries to store the Uuid as `BLOB` at runtime.

Are there thoughts on this? If anybody can give me a hint on which solution would be preferred, I could try to open a PR :)

Huliiiiii:
Currently, we erased the UUID format in sea-query.
Therefore, we need to:
1. Add UUID format support in sea-query.
2. Implement schema discovery in sea-schema.
3. Provide CLI configuration for UUID format generation.

Huliiiiii:
> ## Description
> Sea-ORM / SQLx is saving Uuids as binary blobs in sqlite. Based on the type mappings in https://www.sea-ql.org/SeaORM/docs/generate-entity/column-types/#type-mappings, the uuid should be saved as `uuid_text` for sqlite. This _seems_ like that would be the UUID's hyphenated text version.
> 
> ## Steps to Reproduce
> 1. Clone https://github.com/linux4life798/sea-orm-sqlite-uuid-issue
> 2. `cargo run`
> 3. `sqlite3 ./db.sqlite .dump`
> 4. Observe that the UUID shown is prefixed with `X'`, indicating that it is binary (displayed in hex).
> 
> ### Expected Behavior
> Based on the type mappings in https://www.sea-ql.org/SeaORM/docs/generate-entity/column-types/#type-mappings, the uuid should be saved as `uuid_text` for sqlite. This _seems_ like that would be the UUID's hyphenated text version.
> 
[...]

PR #2914, opened 2026-01-15
Add wrapper type for storing Uuids as TEXT

raffomania:
## PR Info

- Related to #2717, but does not close it
- I haven't tested this with databases other than SQLite
- I'd like to add a test asserting that the database schema actually uses the `TEXT` type but haven't found out how
- First time contributing, please let me know if I missed anything!

## New Features

- [x] Add `TextUuid` wrapper type for storing Uuids as a `TEXT` column

### DIFF
--- a/src/value/text_uuid.rs
+++ b/src/value/text_uuid.rs
@@ -1,0 +1,104 @@
+use std::ops::{Deref, DerefMut};
+
+use sea_query::{ValueType, ValueTypeErr};
+
+use crate::TryGetable;
+use crate::{self as sea_orm, TryFromU64};
+use crate::{DbErr, TryGetError};
+
+/// Newtype making sure that UUIDs will be stored as `TEXT` columns,
+/// instead of `BLOB` (which is the default).
+/// Advantages:
+/// - TEXT makes it easier to interact with the SQLite DB directly
+/// - Allows for queries like `WHERE id IN (<uuid>, <uuid>, ...)` which are
+///   impossible to write with `BLOB` values
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+pub struct TextUuid(pub uuid::Uuid);
+
+impl From<TextUuid> for sea_query::Value {
+    fn from(value: TextUuid) -> Self {
+        value.0.to_string().into()
+    }
+}
+
+impl TryGetable for TextUuid {
+    fn try_get_by<I: sea_orm::ColIdx>(
+        res: &sea_orm::QueryResult,
+        index: I,
+    ) -> Result<Self, sea_orm::TryGetError> {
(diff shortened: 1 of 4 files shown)


src/value/text_uuid.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::{DbErr, TryGetError};
+
+/// Newtype making sure that UUIDs will be stored as `TEXT` columns,
+/// instead of `BLOB` (which is the default).
+/// Advantages:
+/// - TEXT makes it easier to interact with the SQLite DB directly
+/// - Allows for queries like `WHERE id IN (<uuid>, <uuid>, ...)` which are
+///   impossible to write with `BLOB` values
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+pub struct TextUuid(pub uuid::Uuid);
```
Huliiiiii, line 16: UuidText or TextUuid? @tyt2y3 @Expurple

Expurple: I prefer `TextUuid` (AdjectiveNoun)

Expurple (commented):
This seems to only cover the [entity-first approach](https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-first/) where you define entities by hand and can set a custom column type.

I assume, you don't use the migration-first approach with [autogenerated entities](https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/#generating-entity-files)? This use case is currently unsupported by the PR.

This isn't necessarily a problem. We can postpone this feature (a new `sea-orm-cli` flag, a-la `--uuid-type=text`) until someone actually needs it.

If you want to contribute that too, you can do that in a separate PR, dependent on this one

Huliiiiii (approved):

tyt2y3:
I would probably have written it as UuidText haha, but doesn't matter.
we can merge this in first, I will try to add more support around it

raffomania:
> I assume, you don't use the migration-first approach with [autogenerated entities](https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/#generating-entity-files)? This use case is currently unsupported by the PR.

Yeah, we don't generate entities because we couldn't figure out how to use our own custom wrapper types with that workflow.

I'd be fine with contributing a feature for the entity generation as well, I assume I'd start by adding a field to the [codegen EntityWriterContext](https://github.com/seaql/sea-orm/blob/a8813bd2f5f82489e6bb174920eb5c54a40a0f90/sea-orm-codegen/src/entity/writer.rs#L79)?

Expurple:
Yeah, looks like a good place to start. I'm not familiar with the codegen/CLI internals, but you can probably just copy whatever `--big-integer-type` is doing

PR #2926, opened 2026-01-23
Fix DeriveIntoActiveModel

### LINKED ISSUE #2918: Fail to derive `IntoActiveModel` trait for a struct that maps `Option<V>` to `V`
## Description

IntoActiveValue fails to map `Option<V>` in the active model to `V` in the entity model.

## Steps to Reproduce

Consider the following entity model:
```rust
use serde::Serialize;

#[sea_orm::model]
#[sea_orm(table_name = "user")]
pub struct Model {
    pub id: i64,
    pub name: String,
    pub phone: String,
}

```

Derive `IntoActiveModel` trait for the following struct:
```rust
#[derive(Debug, DeriveIntoActiveModel)]
pub struct UserUpdateCommand {
    pub id: i64,
    pub name: Option<String>,
    pub phone: Option<String>,
}
```

### Expected Behavior

If `name` is None, its active value should be `NotSet`, and if it is `Some(v)`, its active value should be `Set(v)`. The same applies to `phone`.

### Actual Behavior

Compilation error:

> the trait bound `ActiveValue<std::string::String>: From<ActiveValue<std::option::Option<std::string::String>>>` is not satisfied
the trait `From<ActiveValue<V>>` is implemented for `ActiveValue<std::option::Option<V>>`
required for `ActiveValue<std::option::Option<std::string::String>>` to implement `Into<ActiveValue<std::string::String>>`

### Reproduces How Often

Always reproducible.

### Workarounds

In `src/entity/active_value.rs: 110`:
```rust
impl<V> IntoActiveValue<Option<V>> for Option<V>
where
    V: IntoActiveValue<V> + Into<Value> + Nullable,
{
    fn into_active_value(self) -> ActiveValue<Option<V>> {
        match self {
            Some(value) => Set(Some(value)),
            None => NotSet,
        }
    }
}
```
tries to map `Option<V>` to `ActiveValue<Option<V>>` which seems meaningless.
I guess the developer intended to map `Option<V>` to `ActiveValue<V>`:
```rust
impl<V> IntoActiveValue<V> for Option<V>
where
    V: IntoActiveValue<V> + Into<Value> + Nullable,
{
    fn into_active_value(self) -> ActiveValue<V> {
        match self {
            Some(value) => Set(value),
            None => NotSet,
        }
    }
}
```

See https://github.com/SeaQL/sea-orm/compare/master...GavenKam:sea-orm:master, which works for my use case.

## Versions

2.0.0-rc.28

GavenKam:
## PR Info

- Closes #2918 

## Changes

Changed blanket implementation in [src/entity/active_value.rs: 110](https://github.com/SeaQL/sea-orm/blob/master/src/entity/active_value.rs#L110), without which the added test cases cause compilation errors on `Derive(DeriveIntoActiveModel)`.

### DIFF
--- a/sea-orm-sync/src/database/tracing_spans.rs
+++ b/sea-orm-sync/src/database/tracing_spans.rs
@@ -24,79 +24,85 @@
 //! let cakes = Cake::find().all(&db)?;  // Generates a span
 //! ```
 
-use crate::DbBackend;
+#[cfg(feature = "tracing-spans")]
+mod inner {
+    use crate::DbBackend;
 
-/// Database operation type, following OpenTelemetry conventions.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum DbOperation {
-    /// SELECT query
-    Select,
-    /// INSERT statement
-    Insert,
-    /// UPDATE statement
-    Update,
-    /// DELETE statement
-    Delete,
-    /// Other/unknown SQL execution
-    Execute,
-}
+    /// Database operation type, following OpenTelemetry conventions.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub(crate) enum DbOperation {
+        /// SELECT query
+        Select,
+        /// INSERT statement
+        Insert,
+        /// UPDATE statement
+        Update,
+        /// DELETE statement
+        Delete,
+        /// Other/unknown SQL execution
+        Execute,
+    }
 
-impl DbOperation {
-    /// Parse the operation type from an SQL query string.
-    ///
-    /// This function is allocation-free and uses case-insensitive comparison.
-    pub fn from_sql(sql: &str) -> Self {
-        let first_word = sql.trim_start().split_whitespace().next().unwrap_or("");
-        if first_word.eq_ignore_ascii_case("SELECT") {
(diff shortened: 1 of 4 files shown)


GavenKam:
@tyt2y3 There was an issue #323  before refactoring `IntoActiveValue`. The focus of discussion was the behavior of `Option<V> -> Option<V>` - shall `Option<V>` in an active model represent `optional but not null` or `required but nullable`? There was not a final conclusion.
My change makes `Option<V>`s under all circumstances, no matter `Option<V> -> Option<V>` or `Option<V> -> V`, convert into `ActiveValue<V>`s. If it is the  `Option<V> -> Option<V>` case, `ActiveValue<V>` is later converted into `ActiveValue<Option<V>>` back again through [`ActiveValue<Option<V>>::From<ActiveValue<V>>`](https://github.com/SeaQL/sea-orm/blob/master/src/entity/active_value.rs#L433). However, both cases are affected by the same implementation now.
In current version, the actual behavior of `Option<V>` in the `Option<V> -> Option<V>` case is `optional but not null`. My change do not affect it so it is not a breaking change, but it may not meet your original intention.

tyt2y3:
@GavenKam thank you for the explanation!

PR #2944, opened 2026-02-05
fix: use i64 for COUNT(*) result on MySQL and SQLite

devgony:
## Summary

- Use `i64` consistently for `COUNT(*)` results across all database backends (MySQL, PostgreSQL, SQLite)
- Remove unnecessary `db_backend` variable that was only used for type selection

## Problem

MySQL's `COUNT(*)` returns `BIGINT` (64-bit), but SeaORM was reading it as `i32`, which could cause overflow for datasets with more than ~2 billion rows.

## Historical Context

The original code (before commit 9c865cbb) used `i32` for all databases:

```rust
let num_items = result.try_get::<i32>("", "num_items")? as usize;
```

In August 2021, commit [9c865cbb](https://github.com/SeaQL/sea-orm/commit/9c865cbb31d8b63bce7de296f44c46d4b7a570b0) ("Try to fix count()") added PostgreSQL-specific handling because PostgreSQL's `COUNT(*)` returns `BIGINT` which cannot be decoded as `i32`:

```rust
let num_items = match self.db {
    #[cfg(feature = "sqlx-postgres")]
    DatabaseConnection::SqlxPostgresPoolConnection(_) => 
        result.try_get::<i64>("", "num_items")? as usize,
    _ => result.try_get::<i32>("", "num_items")? as usize
};
```

However, **MySQL also returns `BIGINT` for `COUNT(*)`**. The fix only addressed PostgreSQL because:
1. PostgreSQL's sqlx driver strictly enforces type matching (i64 required)
2. MySQL's sqlx driver was more lenient and could decode small values as i32
3. Test datasets were small enough that overflow never occurred

## Solution

Use `i64` for all backends since:
- PostgreSQL `COUNT(*)` → `BIGINT` (i64)
- MySQL `COUNT(*)` → `BIGINT` (i64)
- SQLite handles dynamic integer sizing and works with i64

```rust
let num_items = result.try_get::<i64>("", "num_items")? as u64;
```

## Changes

- `src/executor/paginator.rs` (async version)
- `sea-orm-sync/src/executor/paginator.rs` (sync version)

### DIFF
--- a/tests/paginator_tests.rs
+++ b/tests/paginator_tests.rs
@@ -1,0 +1,192 @@
+#![allow(unused_imports, dead_code)]
+
+pub mod common;
+
+pub use common::{TestContext, features::*, setup::*};
+use pretty_assertions::assert_eq;
+use sea_orm::{entity::prelude::*, PaginatorTrait, QueryOrder, Set};
+
+#[sea_orm_macros::test]
+async fn paginator_tests() -> Result<(), DbErr> {
+    let ctx = TestContext::new("paginator_tests").await;
+    create_tables(&ctx.db).await?;
+    create_insert_default(&ctx.db).await?;
+    paginator_num_items(&ctx.db).await?;
+    paginator_num_pages(&ctx.db).await?;
+    paginator_num_items_and_pages(&ctx.db).await?;
+    paginator_fetch_page(&ctx.db).await?;
+    paginator_count(&ctx.db).await?;
+    ctx.delete().await;
+
+    Ok(())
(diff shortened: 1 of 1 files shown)


tyt2y3:
hi, thank you for your contribution. I realize we don't actually have test coverage for it in our test suite. can you add some test cases in a new file `paginator_tests`, with content like https://github.com/SeaQL/sea-orm/blob/master/tests/cursor_tests.rs ? this would give us more confidence, thanks!

PR #2950, opened 2026-02-11
Don't create redundant unique indexes on PG and SQLite

### LINKED ISSUE #2873: redundant unique index on entity-first approach
## Description
[Discord Ref](https://discord.com/channels/873880840487206962/1452817435102023753)
With the new entity-first approach, we can add unique as below
```rs
#[sea_orm(unique, column_type = "Text")]
pub url: String,
```
I found it will generate two indexes, at least on Postgres
```SQL
                                        Table "public.url"
   Column    |           Type           | Collation | Nullable |             Default
-------------+--------------------------+-----------+----------+----------------------------------
 id          | bigint                   |           | not null | generated by default as identity
 url         | text                     |           | not null |

Indexes:
    "url_pkey" PRIMARY KEY, btree (id)
    "idx-url-url" UNIQUE, btree (url)
    "url_url_key" UNIQUE CONSTRAINT, btree (url)
```
As you can see, it generated two indexes for  `UNIQUE` property. I seeked help from the PG discord channel and asked if it is necessary and the adivce I got back was hinting it could be a mistake to have both indexes and `UNIQUE CONSTRAINT` is more "CORRECT", althought the difference is minimal.

### Expected Behavior

It should use just `UNIQUE CONSTRAINT`
### Actual Behavior
```sql
 "idx-url-url" UNIQUE, btree (url)
```
is redundant

## Versions
sea-orm v2

rumpuslabs:
## PR Info

- Closes #2873

## Bug Fixes

- [ ] Prevent creation of redundant UNIQUE indexes on Postgres and SQLite

### DIFF
--- a/src/schema/entity.rs
+++ b/src/schema/entity.rs
@@ -141,7 +141,7 @@ where
 
 pub(crate) fn create_index_from_entity<E>(
     entity: E,
-    _backend: DbBackend,
+    backend: DbBackend,
 ) -> Vec<IndexCreateStatement>
 where
     E: EntityTrait,
@@ -152,7 +152,7 @@ where
     for column in E::Column::iter() {
         let column_def = column.def();
 
-        if column_def.indexed || column_def.unique {
+        if column_def.indexed || (column_def.unique && backend == DbBackend::MySql) {
             let mut stmt = Index::create()
                 .name(format!("idx-{}-{}", entity.to_string(), column.to_string()))
                 .table(entity)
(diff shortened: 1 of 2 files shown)


rumpuslabs:
Postgres and SQLite both automatically create indexes for `UNIQUE` constraints, so the additional explicit index creation is redundant.

[Postgres](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS):
> Adding a unique constraint will automatically create a unique B-tree index on the column or group of columns listed in the constraint.

[SQLite](https://www.sqlite.org/lang_createtable.html#unique_constraints):
> In most cases, UNIQUE and PRIMARY KEY constraints are implemented by creating a unique index in the database.

tyt2y3 (approved):

src/schema/entity.rs
```diff
@@ -152,7 +152,7 @@ where
     for column in E::Column::iter() {
         let column_def = column.def();
 
-        if column_def.indexed || column_def.unique {
+        if column_def.indexed || (column_def.unique && backend == DbBackend::MySql) {
```
Huliiiiii, line 155: It’d help to add a comment here to explain why we filter pg/sqlite out.

rumpuslabs:
In the process of writing the requested comment, I got to thinking and did some checking, and it seems like MySQL also [automatically creates an index](https://www.db-fiddle.com/f/eJcd1YpWJby8U8Wwh5haXk/0) from a UNIQUE constraint. I'm not sure where I got the idea that this was a MySQL-specific workaround.

Perhaps the correct fix is simply to remove `|| column_def.unique` from that line? It was added in https://github.com/SeaQL/sea-orm/commit/bc0f6b092fbec3e9b542080ea6f733cd116f1d91 by @tyt2y3

rumpuslabs:
Or even `if column_def.indexed && !column_def.unique` since the `indexed` attribute is redundant?

tyt2y3:
> Or even `if column_def.indexed && !column_def.unique` since the `indexed` attribute is redundant?

let's do this. I will make the change

PR #2954, opened 2026-02-16
Derive Clone for topologies.

nihohit:
Without this,
```
#[derive(Clone, Debug)]
pub struct SelectThree<E, F, G, TOP>
```
doesn't actually derive clone for these topologies, because `#[derive(Clone)]` only derives clone if all generic types implement clone.

### DIFF
--- a/src/query/select.rs
+++ b/src/query/select.rs
@@ -43,11 +43,11 @@ where
 pub trait Topology {}
 
 /// A star topology
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct TopologyStar;
 
 /// A chain topology
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct TopologyChain;
 
 impl Topology for TopologyStar {}


nihohit:
From local runs, the [unit test workflow](https://github.com/SeaQL/sea-orm/actions/runs/22051711347/job/63711066338?pr=2954) is broken in main due to [this commit](https://github.com/SeaQL/sea-orm/commit/449a78687be2210e64d431f1ab1f04f32f29cc9e). It wasn't broken by this change.

sea-orm-sync/src/query/select.rs
```diff
@@ -40,14 +40,14 @@ where
 }
 
 /// Topology of multi-joins
-pub trait Topology {}
+pub trait Topology: Clone {}
```
Huliiiiii: I think it's better to move this to the call site (e.g., Topology + clone) rather than adding this constraint on Topology.

nihohit: done

nihohit: another option is to completely avoid the constraint - only derive `Clone` for the topologies, and let the `#[derive(Clone)]` on the various selects be dependent on the chosen topology and not enforced.

Huliiiiii: > another option is to completely avoid the constraint - only derive `Clone` for the topologies, and let the `#[derive(Clone)]` on the various selects be dependent on the chosen topology and not enforced.

This looks cleaner.

nihohit: done

Huliiiiii (approved):

tyt2y3:
Thank you!

Issue #2947, opened 2026-02-09
Enum imports duplicated when merging grouped and individual use statements


lmurri:
## Description

When regenerating entities with `sea-orm-cli generate entity --expanded-format` and `preserve_user_modifications` enabled, enum imports can be duplicated if the existing file uses grouped imports and the newly generated output uses individual imports.

This results in semantically duplicate `use` statements being emitted.

## Steps to Reproduce

1. Generate entities with:
```
   sea-orm-cli generate entity --expanded-format
```
2. Modify the generated file to group enum imports, e.g.:
```
   use super::sea_orm_active_enums::{Status, Category};
```
3. Re-run entity generation with `preserve_user_modifications` enabled.
4. Observe the merged output.

### Expected Behavior

Only one import per enum path should exist in the merged output, e.g.:

```
use super::sea_orm_active_enums::Status;
use super::sea_orm_active_enums::Category;
```

(or the grouped form, but not both).

### Actual Behavior

The merge produces duplicate imports:
```
use super::sea_orm_active_enums::{Status, Category};
use super::sea_orm_active_enums::Status;
use super::sea_orm_active_enums::Category;
```

### Reproduces How Often

Always, given the grouped → individual import transition.

### Workarounds

delete all generated code before generation 

## Versions

- sea-orm-cli: 2.0.0-rc.30
- sea-orm-codegen: 2.0.0-rc.30
- OS: macOS (also reproducible on Linux)

lmurri:
Added a test to show the bug here: https://github.com/SeaQL/sea-orm/pull/2945/changes

PR #2964, opened 2026-02-20
Mark preserve-user-modifications as experimental and default it to false

Huliiiiii:
This is too hacky and should only be used as a temporary workaround. It will be replaced by the code-first workflow in the future.

I'm also setting the default value to false because there are currently several known issues and likely other unknown edge cases, such as #2947.


tyt2y3 (approved):
OK! thanks for the feature in the first place

PR #2971, opened 2026-02-26
Fix unique column in schema sync

### LINKED ISSUE #2970: Can't `sync` tables that has columns with `#[sea_orm(unique)]`
## Description

When using entity first workflow, I can no longer use `sync` if one of the column in the table is marked `#[sea_orm(unique)]` after upgrading to sea-orm-2.0.0-rc.34, it tries to drop the index for some reason

Looking at the recent changes, I believe these commits might be related

https://github.com/SeaQL/sea-orm/commit/0c7f8eaa6590c22ecf03d905a4065bc158a04610
https://github.com/SeaQL/sea-orm/commit/1bb4c12ba3f1a46b481d9591205df96fecf1d724

From debugging, it seems like we no longer push `unique` columns to `EntitySchemaInfo::indexes` so the diff code will now tries to remove the index

https://github.com/SeaQL/sea-orm/blob/e36eea425b641da14c484435e97cd462f17630bc/src/schema/builder.rs#L390-L412

I have only tested with Postgresql but I assume it would be the same for the other backends

## Steps to Reproduce

`test.rs`

```rust

#[sea_orm::model]
#[sea_orm(table_name = "test_table")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    #[sea_orm(unique)]
    pub name: String,
}

```

```rust
db.get_schema_builder()
    .register(test::Entity)
    .sync(&db)
    .await?;
```

### Expected Behavior

No error

### Actual Behavior

`Exec(SqlxError(Database(PgDatabaseError { severity: Error, code: "2BP01", message: "cannot drop index test_table_name_key because constraint test_table_name_key on table test_table requires it", detail: None, hint: Some("You can drop constraint test_table_name_key on table test_table instead."), position: None, where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("dependency.c"), line: Some(791), routine: Some("findDependentObjects") })))`

### Reproduces How Often

Every time after the first run (the run creating the table is fine)

### Workarounds

Downgrade to sea-orm-2.0.0-rc.32

## Versions

sea-orm-2.0.0-rc.34

tyt2y3:
Fixes #2970

### DIFF
--- a/tests/schema_sync_tests.rs
+++ b/tests/schema_sync_tests.rs
@@ -132,6 +132,7 @@ async fn test_sync_unique_column_no_drop() -> Result<(), DbErr> {
 /// COLUMN … UNIQUE), which creates a column-level unique index. A subsequent
 /// sync must not attempt to drop that index.
 #[sea_orm_macros::test]
+#[cfg(not(any(feature = "sqlx-sqlite", feature = "rusqlite")))]
 async fn test_sync_add_unique_column_no_drop() -> Result<(), DbErr> {
     let ctx = TestContext::new("test_sync_add_unique_column_no_drop").await;
     let db = &ctx.db;


tyt2y3:
@Legend-Master can you try this branch?

Legend-Master:
Tested and works in my project, thanks for the quick fix!

tyt2y3:
cool, thanks for confirming! yeah I don't like regression :(

PR #2980, opened 2026-03-03
Add per-migration transaction control for SeaORM migrations

tyt2y3:
Add `MigrationTrait::use_transaction()` to control whether individual migrations
run inside a transaction, replacing the previous all-or-nothing batch transaction
behavior on Postgres.

### Core (`sea-orm`)

- Add `OwnedTransaction(DatabaseTransaction)` variant to `DatabaseExecutor` enum,
  enabling `SchemaManager` to own a transaction
- Add `DatabaseExecutor::is_transaction()` method
- Add `IntoDatabaseExecutor` impl for owned `DatabaseTransaction`

### Migration (`sea-orm-migration`)

- Add `MigrationTrait::use_transaction() -> Option<bool>`:
  - `None` (default): follow backend convention (Postgres = txn, MySQL/SQLite = no txn)
  - `Some(true)`: force transaction on any backend
  - `Some(false)`: disable automatic transaction
- Add `SchemaManager::begin()` returning an owned transaction
- Refactor `exec_up_with` / `exec_down_with` to wrap each migration individually
  based on `use_transaction()` + backend, instead of batching all migrations

### Tests

- Add `run_transaction_test` covering all three `use_transaction` modes plus
  failure-with-txn (DDL rolled back) and failure-without-txn (DDL persists)
- Update existing Postgres rollback assertions to match per-migration behavior
- Switch test runtime from `async-std` to `tokio`

## Example Usage

```rust
// Opt out of automatic transaction (e.g. for CREATE INDEX CONCURRENTLY)
impl MigrationTrait for Migration {
    fn use_transaction(&self) -> Option<bool> { Some(false) }

        // DDL in one transaction
        let m = manager.begin().await?;
        m.create_table(...).await?;
        m.commit().await?;

        // Non-transactional DDL
        manager.get_connection()
            .execute_unprepared("CREATE INDEX CONCURRENTLY ...")
            .await?;
        Ok(())
    }
}
```

### DIFF
--- a/sea-orm-migration/tests/common/migration/m20250101_000002_manual_transaction.rs
+++ b/sea-orm-migration/tests/common/migration/m20250101_000002_manual_transaction.rs
@@ -1,0 +1,50 @@
+use sea_orm_migration::prelude::*;
+use sea_orm_migration::schema::*;
+
+pub struct Migration;
+
+impl MigrationName for Migration {
+    fn name(&self) -> &str {
+        "m20250101_000002_manual_transaction"
+    }
+}
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    fn use_transaction(&self) -> Option<bool> {
+        Some(false)
+    }
+
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        assert!(
+            !manager.get_connection().is_transaction(),
+            "outer manager should not be in a transaction"
+        );
+
+        let m = manager.begin().await?;
+        assert!(
+            m.get_connection().is_transaction(),
+            "inner manager should be in a transaction"
+        );
+        m.create_table(
+            Table::create()
+                .table("manual_txn_table")
+                .col(pk_auto("id"))
+                .col(string("name"))
+                .to_owned(),
+        )
+        .await?;
+        m.commit().await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let m = manager.begin().await?;
+        m.drop_table(Table::drop().table("manual_txn_table").to_owned())
+            .await?;
+        m.commit().await?;
+
+        Ok(())
+    }
+}
--- a/sea-orm-migration/tests/common/migration/mod.rs
+++ b/sea-orm-migration/tests/common/migration/mod.rs
@@ -5,3 +5,4 @@ pub mod m20220118_000004_create_tea_enum;
 pub mod m20220923_000001_seed_cake_table;
 pub mod m20230109_000001_seed_cake_table;
 pub mod m20250101_000001_create_test_table;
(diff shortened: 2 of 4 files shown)


sea-orm-migration/src/migrator/exec.rs
```diff
@@ -198,6 +198,48 @@ pub async fn drop_everything<C: ConnectionTrait>(db: &C) -> Result<(), DbErr> {
...
 
+fn should_use_transaction(migration: &dyn crate::MigrationTrait, backend: DbBackend) -> bool {
+    match migration.use_transaction() {
+        Some(v) => v,
+        None => backend == DbBackend::Postgres,
+    }
+}
+
+async fn insert_migration_record<C: ConnectionTrait>(
+    db: &C,
```
sinder38, line 209: this Is very minor but why not:
```rust
async fn insert_migration_record(db: &impl ConnectionTrait, ...)
```

tyt2y3: thanks for the suggestion, a long time ago the convention is to use `where T`. we can have another PR to migrate all of them in the codebase where it makes sense

yuriy-yarosh:
@tyt2y3 Yeah... but It does wrap on drop even when I've 
```rust
fn use_transaction(&self) -> Option<bool> {
        Some(false)
    }
```

```
Rolling back all applied migrations
Rolling back migration 'm0000006_base_enums'
Migration 'm0000006_base_enums' has been rolled back
Rolling back migration 'm0000005_base_types'
Migration 'm0000005_base_types' has been rolled back
Rolling back migration 'm0000004_extensions'
Migration 'm0000004_extensions' has been rolled back
Rolling back migration 'm0000003_schemas'
Migration 'm0000003_schemas' has been rolled back
Rolling back migration 'm0000002_databases'
Migration 'm0000002_databases' has been rolled back
Rolling back migration 'm0000001_tablespaces'
Execution Error: error returned from database: DROP TABLESPACE неможливо запустити всередині блоку транзакції
Fail to run migration
```

```rust

use sea_orm_migration::{prelude::*, sea_orm::Statement};

pub struct DatabaseMigration;

#[async_trait::async_trait]
impl MigrationTrait for DatabaseMigration {
    fn use_transaction(&self) -> Option<bool> {
        Some(false)
    }

        db.execute_unprepared("CREATE DATABASE app;").await?;

        Ok(())
    }

        // 1. Terminate all active connections to the target database
        let disconnect_query = format!(
            "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = '{}' AND pid <> pg_backend_pid();",
            "app"
        );
        db.execute_unprepared(&disconnect_query).await?;

        // db.execute_raw(Statement::from_string(db.get_database_backend(), "DROP DATABASE app;")).await?;
        db.execute_unprepared("DROP DATABASE IF EXISTS app;").await?;

        Ok(())
    }
}

```

PR #2989, opened 2026-03-06
Add `prefix` option to `nested` attribute in `FromQueryResult` and `PartialModel`

### LINKED ISSUE #2981: add `prefix` support to `FromQueryResult` `nested` attribute
## Motivation

Right now as far as I can see there's no way to add `prefix` to a nested model, not having this feautre stops us from using/nseting the same model multiple times inside same model.

## Proposed Solutions

We should add support for setting `prefix` key inside the `nested` attribute, so user can do `#[sea_orm(nested(prefix = "player_"))]` or maybe if just `#[sea_orm(nested(prefix))]` we can prefix the nested model with the variable name that it's set to...

YouKnow-sys:
## PR Info

- Closes #2981

## New Features

- add prefix option to nested attribute: `#[sea_orm(nested(prefix = "..."))]` for `FromQueryResult` and `PartialModel` derives.
- compile-time error when multiple nested fields share the same type without distinct prefixes.

## Bug Fixes

before this change having two nested model with same type will silently have unexpected behavior.

## Breaking Changes

nothing, unless user depends on two field of same nested model have exactly same value... which is wrong behavior to begin with.

## Changes

- add `get_list_args helper` to `GetMeta` trait for parsing list-style attributes.
- use `syn::Error` instead of custom Error enum in `FromQueryResult` derive for better error spans. (tbh we should do this for all other derive macros, but that's for another PR)

### DIFF
--- a/sea-orm-macros/src/lib.rs
+++ b/sea-orm-macros/src/lib.rs
@@ -708,7 +708,7 @@ pub fn derive_active_enum(input: TokenStream) -> TokenStream {
 /// ### Attributes
 ///
 /// - `skip`: will not try to pull this field from the query result. And set it to the default value of the type.
-/// - `nested`: allows nesting models. can be any type that implements `FromQueryResult`. upports `nested(prefix = "...")` to set an explicit column prefix.
+/// - `nested`: allows nesting models. can be any type that implements `FromQueryResult`. supports `nested(prefix = "...")` to set an explicit column prefix.
 /// - `alias` / `from_alias`: get the value from this column alias
 ///
 /// ### Usage
--- a/sea-orm-macros/src/lib.rs
+++ b/sea-orm-macros/src/lib.rs
@@ -708,7 +708,7 @@ pub fn derive_active_enum(input: TokenStream) -> TokenStream {
 /// ### Attributes
 ///
 /// - `skip`: will not try to pull this field from the query result. And set it to the default value of the type.
-/// - `nested`: allows nesting models. can be any type that implements `FromQueryResult`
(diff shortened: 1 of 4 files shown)


sea-orm-macros/src/lib.rs
```diff
@@ -747,6 +747,9 @@ pub fn derive_active_enum(input: TokenStream) -> TokenStream {
 ///     price: Decimal,
 ///     #[sea_orm(nested)]
 ///     baker: Option<cakes_bakers::Model>,
+///     // prefix is optional, useful when multiple fields share the same type
+///     #[sea_orm(nested(prefix = "reviewer_"))]
+///     reviewer: Option<cakes_bakers::Model>,
```
Expurple, line 752: Can we add a codegen test for this?

Expurple (commented):
Hi. Thank you for the PR

sea-orm-macros/src/lib.rs
YouKnow-sys: I added tests to `partial_model`, should I also add some test to the `tests/partial_model_nested/nested_alias.rs` and `tests/from_query_result_tests.rs` as well? seem unnecessary because the logic is clear, but if needed I can do it.

Expurple: Haha, I meant asserting the generated SQL query. Probably in a doctest. We have many examples of that in the codebase. Sorry for being unclear

YouKnow-sys: Hi, sorry for the late response...
I just pushed a new commit, please check it and tell me if any more changes is needed.
also feel free to change the code yourself, I have `Allow edits by maintainers` enabled.

PR #3002, opened 2026-03-12
Fix exec_with_connection ignoring per-migration use_transaction config

### LINKED ISSUE #3001: migration drop disregards `use_transaction -> Option<bool>`
## Description

I can't rollback migrations with statements that can not be wrapped in transactions, e.g. `DROP DATABASE` `DROP TABLESPACE`.

## Steps to Reproduce

As per [#2980](https://github.com/SeaQL/sea-orm/pull/2980#issuecomment-4048595849)

1. Create migration
```rust

use sea_orm_migration::{prelude::*, sea_orm::Statement};

pub struct DatabaseMigration;

#[async_trait::async_trait]
impl MigrationTrait for DatabaseMigration {
    fn use_transaction(&self) -> Option<bool> {
        Some(false)
    }

        db.execute_unprepared("CREATE DATABASE app;").await?;

        Ok(())
    }

        // 1. Terminate all active connections to the target database
        let disconnect_query = format!(
            "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = '{}' AND pid <> pg_backend_pid();",
            "app"
        );
        db.execute_unprepared(&disconnect_query).await?;


        // 2. Drop the database
        db.execute_unprepared("DROP DATABASE IF EXISTS app;").await?;

        // Raw does not do the trick either
        // db.execute_raw(Statement::from_string(db.get_database_backend(), "DROP DATABASE app;")).await?;

        Ok(())
    }
}
```

2. Drop it with  

```bash
sea-orm-cli migrate -d .
sea-orm-cli migrate -d . reset
```

3. See that it wraps DROP DATABASE into a transaction

Image

### Expected Behavior

Supposedly, it should've rollback the same way it's been applied.

### Actual Behavior

It does not rollback.

### Reproduces How Often

Always.

### Workarounds

Running raw `sqlx` queries myself.

## Versions

```bash
cargo tree | grep sea-
├── sea-orm-migration v2.0.0-rc.37
│   ├── sea-orm v2.0.0-rc.37
│   │   ├── sea-orm-macros v2.0.0-rc.37 (proc-macro)
│   │   │   ├── sea-bae v0.2.1 (proc-macro)
│   │   ├── sea-query v1.0.0-rc.31
│   │   │   ├── sea-query-derive v1.0.0-rc.12 (proc-macro)
│   │   ├── sea-query-sqlx v0.8.0-rc.14
│   │   │   ├── sea-query v1.0.0-rc.31 (*)
│   │   ├── sea-schema v0.17.0-rc.17
│   │   │   ├── sea-query v1.0.0-rc.31 (*)
│   │   │   ├── sea-query-sqlx v0.8.0-rc.14 (*)
│   │   │   ├── sea-schema-derive v0.3.0 (proc-macro)
│   ├── sea-orm-cli v2.0.0-rc.37
│   ├── sea-schema v0.17.0-rc.17 (*)
```

tyt2y3:
Fixes #3001

+ Remove `exec_with_connection!` macro that unconditionally wrapped all Postgres operations in a transaction
+ Inline the `into_database_executor()` / `SchemaManager::new()` pattern at each call site (fresh, refresh, reset, uninstall), matching the approach already used by `up()` and `down()`
+ Move the Postgres transaction for `drop_everything` into the function itself so atomicity is preserved

### DIFF
--- a/sea-orm-migration/src/migrator/exec.rs
+++ b/sea-orm-migration/src/migrator/exec.rs
@@ -10,7 +10,7 @@ use sea_orm::sea_query::{
 };
 use sea_orm::{
     ActiveValue, ConnectionTrait, DbBackend, DbErr, DynIden, EntityTrait, FromQueryResult,
-    Iterable, QueryFilter, Schema, Statement, TransactionTrait,
+    Iterable, QueryFilter, Schema, Statement, TransactionSession, TransactionTrait,
 };
 
 pub async fn get_migration_models<C>(
@@ -69,33 +69,6 @@ pub fn get_migration_with_status(
     }
 }
 
-macro_rules! exec_with_connection {
-    ($db:ident, $fn:expr) => {{
-        async {
-            let db = $db.into_database_executor();
-
-            match db.get_database_backend() {
-                DbBackend::Postgres => {
-                    let transaction = db.begin().await?;
-                    let manager = SchemaManager::new(&transaction);
-                    $fn(&manager).await?;
-                    transaction.commit().await
-                }
-                DbBackend::MySql | DbBackend::Sqlite => {
-                    let manager = SchemaManager::new(db);
-                    $fn(&manager).await
-                }
-                db => Err(DbErr::BackendNotSupported {
-                    db: db.as_str(),
-                    ctx: "exec_with_connection",
-                }),
-            }
-        }
-    }};
-}
-
-pub(crate) use exec_with_connection;
-
 pub async fn install<C>(db: &C, migration_table_name: DynIden) -> Result<(), DbErr>
 where
     C: ConnectionTrait,
@@ -120,7 +93,17 @@ pub async fn uninstall(
     Ok(())
 }
 
-pub async fn drop_everything<C: ConnectionTrait>(db: &C) -> Result<(), DbErr> {
+pub async fn drop_everything<C: ConnectionTrait + TransactionTrait>(db: &C) -> Result<(), DbErr> {
+    if db.get_database_backend() == DbBackend::Postgres {
+        let transaction = db.begin().await?;
+        drop_everything_impl(&transaction).await?;
+        transaction.commit().await
+    } else {
+        drop_everything_impl(db).await
+    }
+}
+
+async fn drop_everything_impl<C: ConnectionTrait>(db: &C) -> Result<(), DbErr> {
     let db_backend = db.get_database_backend();
 
     // Temporarily disable the foreign key check
--- a/sea-orm-migration/src/migrator.rs
+++ b/sea-orm-migration/src/migrator.rs
@@ -11,7 +11,7 @@ use tracing::info;
 
 use super::{IntoSchemaManagerConnection, MigrationTrait, SchemaManager, seaql_migrations};
 use sea_orm::sea_query::IntoIden;
-use sea_orm::{ConnectionTrait, DbBackend, DbErr, DynIden, TransactionTrait};
+use sea_orm::{ConnectionTrait, DbErr, DynIden};
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 /// Status of migration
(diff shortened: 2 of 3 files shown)


sea-orm-migration/src/migrator/exec.rs
```diff
@@ -120,7 +93,17 @@ pub async fn uninstall(
     Ok(())
 }
 
-pub async fn drop_everything<C: ConnectionTrait>(db: &C) -> Result<(), DbErr> {
+pub async fn drop_everything<C: ConnectionTrait + TransactionTrait>(db: &C) -> Result<(), DbErr> {
+    if db.get_database_backend() == DbBackend::Postgres {
```
yuriy-yarosh: Not sure about this, it'll also need to tolerate `should_use_transaction()`, not just wrap if postgres is used.

Surely, it's much simpler than the macro.

yuriy-yarosh:
Not sure about this.
Only `database` `role`/`user` `extension` and `tablespace` ops can not be put into a transaction.

I'm expecting it to tolerate `use_transaction()` for both apply and rollback ops, disregarding what exact database being used PostgreSQL or MySQL. So, that check is a bit pointless... We may get partial support for transactional DDL in MySQL one day, as well. So it worth respecting, when transaction wrap is explicitly enabled, or explicitly disabled.

tyt2y3:
> Not sure about this. Only `database` `role`/`user` `extension` and `tablespace` ops can not be put into a transaction.
> 
> I'm expecting it to tolerate `use_transaction()` for both apply and rollback ops, disregarding what exact database being used PostgreSQL or MySQL. So, that check is a bit pointless... We may get partial support for transactional DDL in MySQL one day, as well. So it worth respecting, when transaction wrap is explicitly enabled, or explicitly disabled.

sorry, I don't quite get the verdict. is the new behaviour in this PR good enough? note that before this change, everything is wrapped in transaction on Postgres which is a bit strange

tyt2y3:
I think this is a more sensible mechanism, will merge for now

yuriy-yarosh:
@tyt2y3 yes, it's feasible.
I'd skip DB type check and applied implicit transaction wrap, only when migration itself has not declared explicit `use_transaction()` block. But, I presume, this would require parsing and custom proc-macro use, which might be an overkill.

Right now I'm splitting CLI's into multiple admin accounts and run them separately, to fix these types of issues.

```rust

const MIGRATION_PHASE_ENV: &str = "APP_MIGRATION_PHASE";

pub fn set_database_url(url: &str) {
    unsafe {
        env::set_var("DATABASE_URL", url);
    }
}

async fn run_bootstrap() {
    let url = migration::conn::bootstrap_connection_url();
    set_database_url(&url);

    cli::run_cli_with_connection(BootstrapMigrator, |_| async move { Database::connect(url).await }).await;
}

async fn run_app() {
    let url = migration::conn::admin_connection_url();
    set_database_url(&url);

    cli::run_cli_with_connection(AppMigrator, |_| async move { Database::connect(url).await }).await;
}

fn run_phase_subprocess(phase: &str) -> Result<(), Box<dyn std::error::Error + 'static>> {
    let exe = env::current_exe()?;
    let args: Vec<_> = env::args_os().skip(1).collect();

    let status = Command::new(exe).args(args).env(MIGRATION_PHASE_ENV, phase).status()?;

    if status.success() {
        Ok(())
    } else {
        Err(format!("migration phase '{phase}' failed with status {status}").into())
    }
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + 'static>> {
    match env::var(MIGRATION_PHASE_ENV).ok().as_deref() {
        Some("bootstrap") => {
            run_bootstrap().await;
            return Ok(());
        }
        Some("app") => {
            run_app().await;
            return Ok(());
        }
        _ => {}
    }

    if conn::bootstrap_required().await {
        run_phase_subprocess("bootstrap")?;
        if !conn::bootstrap_required().await {
            run_phase_subprocess("app")?;
        }
    }

    if !conn::bootstrap_required().await {
        run_phase_subprocess("app")?;
        if conn::applied_migrations_count().await == 0 {
            run_phase_subprocess("bootstrap")?;
        }
    }
    
    Ok(())
}
```

A bit hacky way, but I didn't figure out anything better atm.

PR #3007, opened 2026-03-14
naive arrow 58 update

yuriy-yarosh:
## PR Info

Naively updates arrow to arrow 58.
Local tests are passing, so I though it should be fine.

## New Features

No more duplicate issues with up to date Cargo crates.

Arrow 58 has explicit [feature](https://github.com/apache/arrow-rs/blob/main/arrow/Cargo.toml#L65) for `async` for `arrow-arrays`, which [enables](https://github.com/apache/arrow-rs/blob/main/arrow-array/Cargo.toml#L58) `futures`.
So, there might be a slight change in feature flags for proper async parquet processing.

## Bug Fixes

Arrow 58 deprecates older API's, so I presume that there should be no regression.

## Breaking Changes

None found.

## Changes

None meaningful.

### DIFF
--- a/examples/parquet_example/Cargo.toml
+++ b/examples/parquet_example/Cargo.toml
@@ -12,7 +12,7 @@ version      = "0.1.0"
 env_logger = { version = "0.11" }
 fastrand   = "2"
 log        = { version = "0.4" }
-parquet    = { version = "57", default-features = false, features = ["arrow"] }
+parquet    = { version = "58", default-features = false, features = ["arrow"] }
 tokio      = { version = "1", features = ["full"] }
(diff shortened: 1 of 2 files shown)


yuriy-yarosh:
зображення
зображення

Expurple:
Hi. Thank you.

There are errors in CI. Apparently, we just need to bump to 58 in `parquet_example` too

yuriy-yarosh:
Yeah... need's fix. Will update this.
Will open additional issue if something pops up.

yuriy-yarosh:
@Expurple all good.

Issue #2952, opened 2026-02-12
entity sync does not work with postgres when tables have schemas defined


fnimick:
## Description

Entity sync fails at runtime with a SQL error when a table is defined to have a schema other than the default.

If the schema does not exist, a missing schema error occurs. If the schema is created, the next sync works. Then, subsequent syncs attempt to re-create the table and a table already exists error occurs:

```
called `Result::unwrap()` on an `Err` value: Exec(SqlxError(Database(PgDatabaseError { severity: Error, code: "42P07", message: "relation \"test_entity\" already exists", detail: None, hint: None, position: None, where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("heap.c"), line: Some(1162), routine: Some("heap_create_with_catalog") })))
```

The full SQL debug error log is:
```
026-02-12T22:58:36.169216Z DEBUG sea_orm::entity::registry: Registering entities with prefix `sea_orm_sync_test::entity::`
2026-02-12T22:58:36.169460Z DEBUG sea_orm::entity::registry: Registered sea_orm_sync_test::entity::test_entity
2026-02-12T22:58:36.169691Z DEBUG sea_orm::driver::sqlx_postgres: SELECT CURRENT_SCHEMA()
2026-02-12T22:58:36.170714Z  INFO sqlx::query: summary="SELECT CURRENT_SCHEMA()" db.statement="" rows_affected=1 rows_returned=1 elapsed=652.25µs elapsed_secs=0.00065225
2026-02-12T22:58:36.170800Z DEBUG sea_orm::driver::sqlx_postgres: SELECT "pg_type"."typname", "pg_enum"."enumlabel" FROM "pg_catalog"."pg_type" INNER JOIN "pg_catalog"."pg_enum" ON "pg_enum"."enumtypid" = "pg_type"."oid" INNER JOIN "pg_catalog"."pg_namespace" ON "pg_namespace"."oid" = "pg_type"."typnamespace" WHERE "pg_namespace"."nspname" = 'public' ORDER BY "pg_type"."typname" ASC, "pg_enum"."enumsortorder" ASC, "pg_enum"."enumlabel" ASC
[...]

PR #3016, opened 2026-03-22
Fix schema sync not discovering tables in non-default schemas

ouywm:
## PR Info

  - Closes https://github.com/SeaQL/sea-orm/issues/2952

  - Dependencies:
    - N/A

  - Dependents:
    - N/A

  ## New Features

  - N/A

  ## Bug Fixes

  - [x] `sync()` only ran `SchemaDiscovery` against `CURRENT_SCHEMA()` (typically `public`). Entities with `#[sea_orm(schema_name = "sys")]` were never discovered, so every sync
   attempted `CREATE TABLE` again, failing with `relation "xxx" already exists`. Root cause has two layers: (1) discovery was hardcoded to a single schema, (2) the table lookup
  compared full `TableName` (with schema qualifier) against discovered tables (without schema qualifier), so even if the table existed it could never match.

  ## Breaking Changes

  - N/A. Fully backward compatible — entities without `schema_name` produce the same single-schema discovery as before.

  ## Changes

  - [x] Add `schema_name: Option<String>` field to `EntitySchemaInfo`, captured from `EntityName::schema_name()`
  - [x] Replace `DiscoveredSchema.tables: Vec<TableCreateStatement>` with `tables_by_schema: HashMap<String, Vec<TableCreateStatement>>` keyed by schema name
  - [x] Collect all unique target schemas from registered entities before discovery; run `SchemaDiscovery::new_no_exec(schema)` for each schema (Postgres & MySQL)
  - [x] Add `DiscoveredSchema::find_table()` for schema-aware table lookup — resolves entity's effective schema (explicit or `current_schema` fallback), then compares bare table
   names within that bucket
  - [x] SQLite unchanged (uses empty string as schema key)
  - [x] Regenerated `sea-orm-sync` via `make-sync.sh`

### DIFF
--- a/sea-orm-sync/tests/schema_sync_tests.rs
+++ b/sea-orm-sync/tests/schema_sync_tests.rs
@@ -4,10 +4,10 @@ pub mod common;
 
 use crate::common::TestContext;
 use sea_orm::{
+    DatabaseBackend, DatabaseConnection, DbErr, Statement,
     entity::*,
     query::*,
-    sea_query::{Condition, Expr, Query},
-    DatabaseBackend, DatabaseConnection, DbErr, Statement,
+    sea_query::{Condition, Expr, Query, SelectStatement},
 };
 
 // Scenario 1: table is first synced with a `#[sea_orm(unique)]` column already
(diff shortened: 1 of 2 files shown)


ouywm:
Done. The unrelated changes were from make-sync.sh syncing other pending commits into sea-orm-sync/. Already cleaned up — now only the two builder.rs files remain.

Huliiiiii:
Could you add a test for this?

sea-orm-sync/src/schema/builder.rs
```diff
@@ -113,15 +132,33 @@ impl SchemaBuilder {
...
-                    .map_err(|err| DbErr::Query(crate::RuntimeErr::SqlxError(err.into())))?;
+                // Collect all unique schemas that registered entities belong to
+                let mut target_schemas = std::collections::BTreeSet::new();
+                for entity in &self.entities {
+                    let schema = entity.schema_name.as_deref().unwrap_or(&current_schema);
+                    target_schemas.insert(schema.to_string());
+                }
+
+                let mut tables_by_schema = std::collections::HashMap::new();
+                let mut all_enums = Vec::new();
```
Huliiiiii: Enums should be grouped by schema as well.

Huliiiiii:
I noticed that the changes in sea-orm-sync and sea-orm are not aligned. I've updated sea-orm-sync. Could you rebase onto master and make sure the changes are consistent between sea-orm and sea-orm-sync?

ouywm:
Good call, thanks — I'll add a regression test for this and update the PR soon.

PR #3023, opened 2026-03-27
Add serde feature and derive Serialize and Deserialize for TextUuid

### LINKED ISSUE #3022: serde support for `TextUuid`
## Motivation

Without `TextUuid` implementing `Serialize` and `Deserialize`, models cannot implement them without using custom serializers/deserializers for each field that uses a `TextUuid`.

## Proposed Solutions

1. Add a new feature which enables the uuid crate's `serde` feature, or
2. Always pull in uuid with `serde` enabled

Then defer serialization/deserialization to the uuid crate's implementations.

Huliiiiii:
- Closes #3022

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -482,6 +482,7 @@ let user = user::ActiveModel::builder()
 ### Enhancements
 
 * Added `serde` feature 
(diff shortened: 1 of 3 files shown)


Expurple (commented):
Not sure why the test is failing. The diff looks good to me.

PR #3031, opened 2026-04-01
fix: use split_whitespace() instead of replace(' ') in proc macros to…

owjs3901:
… handle newlines in long type paths

## PR Info

- Fix `DeriveActiveModelEx` (and other derive macros) producing `compile_error!("unexpected token")` when entity module names are long (~50+ characters)
- Replace `.replace(' ', "")` with `.split_whitespace().collect()` across all 11 occurrences in `sea-orm-macros`

## Bug Fixes

- [ x ] `proc_macro2::TokenStream::to_string()` inserts `\n` (newlines) for line-wrapping when the output exceeds ~80 characters. The existing code used `.replace(' ', "")` to normalize the stringified type, but this **only strips spaces — not newlines**.

When entity paths like `HasMany<super::estimate_user_checker_setting::Entity>` are long enough to trigger line-wrapping, the extracted path retains an embedded `\n`:

image
As shown in the image, an error occurs when a module with a long name is used inside HasMany.

```rust
// Before (breaks on long paths)
.replace(' ', "")

// After (strips all whitespace including \n, \r, \t)
.split_whitespace().collect()
```

### DIFF
--- a/sea-orm-macros/src/derives/active_model_ex.rs
+++ b/sea-orm-macros/src/derives/active_model_ex.rs
@@ -50,9 +50,10 @@ pub fn expand_derive_active_model_ex(
             for field in fields.named.iter() {
                 if field.ident.is_some() && field_not_ignored_compound(field) {
                     let field_type = &field.ty;
-                    let field_type = quote! { #field_type }
+                    let field_type: String = quote! { #field_type }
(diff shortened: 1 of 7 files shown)


tyt2y3:
Note to self: this might not be the best approach to detect `Option<T>`, worth investigating later.

PR #3037, opened 2026-04-03
Generate correct imports in `frontend` format

### LINKED ISSUE #2934: Frontend format uses TimeDateTimeWithTimeZone, which is sea-orm specific.
## Description

`sea-orm-cli generate entity --frontend-format --date-time-crate time`

generates structs with 

`pub created_at: TimeDateTimeWithTimeZone`

which is not imported, as it requires sea_orm. 

## Steps to Reproduce

1. 

```#[sea_orm::model]
#[sea_orm(table_name = "person")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
}
```

2. `sea-orm-cli generate entity --frontend-format --date-time-crate time`
3.
 
```
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct Model {
    pub id: i32,
}
```

### Expected Behavior

--date-time-crate time should use and import time::PrimateDateTime, while chrono uses it's equvilant. 

### Actual Behavior

A non-existant type is included

### Reproduces How Often

100%

## Versions
sea-orm-cli 2.0.0-rc.30

DexterHill0:
## PR Info

- Closes #2934

## Bug Fixes

- [x] When using the `--frontend-format` argument, if any model column types use external types from libraries like `Uuid` or any datetime types, for example, the generated code does not include the correct imports; it assumes there's a `prelude` import, but that is excluded in the `frontend` format, so the types just error.

## Breaking Changes

Should not be a breaking change.

## Notes

I added tests for this, following the existing format as best as I can.


DexterHill0:
Forgot to run the whole test suite, so I missed those failing tests. Hopefully they're fixed now

DexterHill0:
Any update on this?

tyt2y3:
sorry for the late reply, I'll take a look

sea-orm-codegen/src/entity/writer/frontend.rs
```diff
@@ -77,4 +81,159 @@ impl EntityWriter {
             }
         }
     }
+
+    pub fn gen_import_frontend(entity: &Entity, opt: &ColumnOption) -> TokenStream {
+        // seperate enum so `ColumnType` doesnt need to derive `Hash` or `Eq`
+        #[derive(Hash, PartialEq, Eq)]
+        enum ExternalTypes {
+            JsonOrJsonBinary,
+            Date,
```
tyt2y3: can you move this out of the function? and the impl block

sea-orm-codegen/src/entity/writer.rs
```diff
@@ -1879,6 +1988,32 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_gen_frontend_imports() -> io::Result<()> {
+        let imports_entity = setup().get(13).unwrap().clone();
```
tyt2y3: very minor nit:


```suggestion
        let imports_entity = setup().into_iter()
            .find(|e| e.get_table_name_snake_case() == "imports")
            .unwrap();

```

sea-orm-codegen/src/entity/writer/frontend.rs
DexterHill0: Done

Issue #2941, opened 2026-02-04
[tracing-spans] Add option to remove params from traces


pat-trunk-io:
## Motivation

Loving the new tracing feature, however traces include potentially PII info.

I assume these are generated because this value is set to true:
https://github.com/SeaQL/sea-orm/blob/master/src/database/db_connection.rs#L231

## Proposed Solutions

Ideally something similar to here:
[sea-orm-tracing configuration](https://docs.rs/sea-orm-tracing/latest/sea_orm_tracing/#configuration)

```
let config = TracingConfig::default()
    .with_statement_logging(true)  // Include SQL in spans (default: false for security)
    .with_parameter_logging(false) // Include query parameters (default: false)
```
## Additional Information

N/A

PR #3045, opened 2026-04-14
Add option to remove params from traces

WaterWhisperer:
## PR Info

- Part of #2941 

## New Features

- [x] Add `ConnectOptions::tracing_statement_logging(bool)` to let users disable `db.statement` recording when `tracing-spans` is enabled.

## Changes

- [x] Propagate the statement-logging setting through `DatabaseConnection` and `DatabaseTransaction`.
- [x] Keep the current default behavior unchanged, `db.statement` is still recorded unless explicitly disabled.

### DIFF
--- a/sea-orm-sync/src/database/mod.rs
+++ b/sea-orm-sync/src/database/mod.rs
@@ -91,7 +91,7 @@ pub struct ConnectOptions {
     /// Enable SQLx statement logging
     pub(crate) sqlx_logging: bool,
     /// Record SQL statements in tracing spans
-    pub(crate) tracing_statement_logging: bool,
+    pub(crate) record_stmt_in_spans: bool,
     /// SQLx statement logging level (ignored if `sqlx_logging` is false)
     pub(crate) sqlx_logging_level: log::LevelFilter,
     /// SQLx slow statements logging level (ignored if `sqlx_logging` is false)
@@ -233,7 +233,7 @@ impl ConnectOptions {
             acquire_timeout: None,
             max_lifetime: None,
             sqlx_logging: true,
-            tracing_statement_logging: true,
+            record_stmt_in_spans: true,
             sqlx_logging_level: log::LevelFilter::Info,
             sqlx_slow_statements_logging_level: log::LevelFilter::Off,
             sqlx_slow_statements_logging_threshold: Duration::from_secs(1),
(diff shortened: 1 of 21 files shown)


WaterWhisperer:
I used `expect` instead of `allow` because of https://github.com/servo/servo/issues/40383#issue-3584256994. However, to maintain consistency in the codebase, I could use `allow`, but personally, I might prefer `expect`.

tyt2y3:
thank you! just to confirm, you used the shell script to regenerate sea-orm-sync right?

WaterWhisperer:
> thank you! just to confirm, you used the shell script to regenerate sea-orm-sync right?

Oh, sorry, I updated the sync side manually. I didn't know there was a script to use before. I’ll rerun the script and update the PR

Huliiiiii (approved):

sea-orm-sync/src/database/mod.rs
```diff
@@ -344,6 +347,17 @@ impl ConnectOptions {
         self.sqlx_logging
     }
 
+    /// Enable recording `db.statement` in tracing spans (default true).
+    pub fn tracing_statement_logging(&mut self, value: bool) -> &mut Self {
```
Huliiiiii: I think renaming it to `record_stmt_in_spans` would be more consistent?

Or would `record_sql_in_span` be more intuitive?

Huliiiiii:
#2941 wanted was to only remove the parameter. This change just disables statement recording in spans, so it doesn’t seem to fully resolve that issue.

sea-orm-sync/src/database/mod.rs
WaterWhisperer: Thanks! I like `record_stmt_in_spans`.

WaterWhisperer:
> #2941 wanted was to only remove the parameter. This change just disables statement recording in spans, so it doesn’t seem to fully resolve that issue.

Yeah the current change only adds a switch for whether `db.statement` is recorded in spans.

After rechecking the current implementation and testing it locally, it looks like the tracing span path records `stmt.sql` (with placeholders), not the rendered SQL with bound parameters. The parameter injection seems to happen in `Statement`’s `Display` impl, which is used by debug logging, not by `tracing-spans`.

So I may be targeting a different concern than #2941 originally described. Does #2941 refer to another current path that records rendered SQL with parameters, or to an older behavior?

Huliiiiii:
Perhaps it's because of tracing::instrument.

WaterWhisperer:
> Perhaps it's because of tracing::instrument.

Thanks! After looking into it again, I think the real concern in #2941 may be `#[instrument]` recording `stmt: Statement`, rather than `db.statement` itself.

If that’s the case, then this PR may be addressing a separate concern: whether raw SQL is recorded into `db.statement`.

Would you prefer me to keep this PR scoped as that separate option, or would it be better to rework/replace it with a focused fix for the `#[instrument]` related issue? If a new PR would be cleaner for review, I’m happy to close this one and open another.

Huliiiiii:
Keep this PR and open a separate one for the tracing::instrument issue.

tyt2y3 (approved):

PR #3047, opened 2026-04-17
Avoid recording SQL arguments via `tracing::instrument`

### LINKED ISSUE #2941: [tracing-spans] Add option to remove params from traces

WaterWhisperer:
## PR Info

- Closes #2941 

- Dependencies:
  - #3045 

## New Features

- [ ] None.

## Changes

- [x] Skip raw `stmt` / `sql` arguments in `#[instrument]` generated spans across database execution paths.
- [x] Keep `record_stmt_in_spans` responsible only for manual `db.statement` recording.

### DIFF
--- a/sea-orm-sync/src/database/db_connection.rs
+++ b/sea-orm-sync/src/database/db_connection.rs
@@ -140,7 +140,7 @@ impl ConnectionTrait for DatabaseConnection {
         self.get_database_backend()
     }
 
-    #[instrument(level = "trace")]
+    #[instrument(level = "trace", skip(stmt))]
     #[allow(unused_variables)]
     fn execute_raw(&self, stmt: Statement) -> Result<ExecResult, DbErr> {
         super::tracing_spans::with_db_span!(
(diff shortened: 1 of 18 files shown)


WaterWhisperer:
`record_stmt_in_spans` in #3045 already separately controls manual `db.statement` recording. The remaining leakage here comes from `#[instrument]` automatically recording function arguments like `stmt` / `sql` into spans.

I did not turn that into another runtime option, because this part is driven by the `#[instrument]` macro itself. Making it configurable would likely require replacing these macro-generated spans with a different manual tracing path.

PR #3053, opened 2026-04-22
(Please Review Before 2.0 releases!) Remove need to use Box::pin for transactions.

### LINKED ISSUE #2749: Feature Request: Support for async closures
## Motivation

Since [`AsyncFn`](https://doc.rust-lang.org/std/ops/trait.AsyncFn.html) and therefore async closures have been stabilised in rust version 1.85.0 a few months back, are there any plans to support this for example in transactions?

## Proposed Solutions

For example the arguments to the transaction functions can be changed to instead receive `impl AsyncFnOnce() -> Result<T, E>` or if this is to late for version 2.0.0, another pair of transaction functions can be added which receive the async closure.

landonpoch:
## PR Info

With rust >= 1.85 we shouldn't need to Box::pin async closures anymore.  This fixes that.  I'm hoping you'll accept this because it is a breaking change and 2.0 is in RC right now.  This would be timely to get it in prior to a major upgrade.

- Closes #2749 

## New Features

- [x] Just uses the new async closures built in to the language now.

## Bug Fixes

- [x] Resolves #2749 

## Breaking Changes

- [x] You'll need to remove those nasty Box::pin from your transaction closures. <- this is a good thing.

Note!!! If you don't want this as a breaking change consider modifying this PR to just create new methods for this and leave the old stuff in tact.  However, I think it's much better just to rip the bandaid off in this release.

### DIFF
--- a/src/database/connection.rs
+++ b/src/database/connection.rs
@@ -165,6 +165,7 @@ pub struct TransactionOptions {
 }
 
 /// Spawn database transaction
+#[allow(async_fn_in_trait)]
 pub trait TransactionTrait {
     /// The concrete type for the transaction
     type Transaction: ConnectionTrait + TransactionTrait + TransactionSession;
--- a/src/database/connection.rs
+++ b/src/database/connection.rs
@@ -165,7 +165,6 @@ pub struct TransactionOptions {
 }
 
 /// Spawn database transaction
-#[async_trait::async_trait]
 pub trait TransactionTrait {
     /// The concrete type for the transaction
     type Transaction: ConnectionTrait + TransactionTrait + TransactionSession;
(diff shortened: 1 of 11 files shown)


landonpoch:
@tyt2y3 - I know you guys are pretty far along in the RC process at this point.  I just want to make sure you get some eyes on this to see if it could make the cut.  This would be a really nice thing to include as it makes using the api a lot easier.

ProbstDJakob:
The problem with this PR is that, as @Huliiiiii wrote, you can not specify the `Sync` requirement for the returned `Future`. Since those are only `AsyncFnOnce` traits, one could add `<F as AsyncFnOnce>::CallOnceFuture: Send` but this, as well as the [RTN](https://github.com/rust-lang/rust/issues/109417), is nightly only at the moment. Thus there is currently no way to add this without nightly.

landonpoch:
@ProbstDJakob - Got it.  Waiting for CallOnceFuture is probably the move then.  Once that is in, I can update this to add the Send contraint and then it should be ready to go.  Oh well.

tyt2y3:
would be nice if this works on stable!

landonpoch:
Realized I needed to update the tests to remove the Box::pin syntax to get them to pass.  This makes sense because this is the breaking change called out earlier.

Huliiiiii:
https://github.com/SeaQL/sea-orm/issues/2749#issuecomment-4449693440
> I think we could follow diesel's approach here. I don't really have time to work on this recently. Anyone interested in experimenting with it?
> 
> https://blog.weiznich.de/blog/diesel-async-0-9/

tyt2y3:
I've tried the approach Hulii suggested, but it couldn't work because of the associated type `TransactionTrait::Transaction`.

I've added an experimental API in https://github.com/SeaQL/sea-orm/blob/59af61eee31a244a53a5161e68befbf31ef58995/src/database/executor.rs#L205-L209

PR #3055, opened 2026-04-28
feat: support schema_name on DeriveActiveEnum

### LINKED ISSUE #2581: type "schema.x_enum" does not exist - after updating to SQLX 0.8.5
## Description

Before, to have an enum in a namespace that is not public, I would do the following:
```
#[derive(
    Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Serialize, Deserialize, JsonSchema,
)]
#[sea_orm(
    rs_type = "String",
    db_type = "Enum",
    enum_name = "SCHEMA.X_ENUM"
)]
pub enum XEnum {
    #[sea_orm(string_value = "A")]
    A,
    #[sea_orm(string_value = "B")]
    B,
}

impl IntoActiveValue<Self> for XEnum {
    fn into_active_value(self) -> sea_orm::ActiveValue<Self> {
        sea_orm::ActiveValue::Set(self.clone())
    }
}
```
This was hacky but it would work fine. It seems that now, after updating to sqlx 0.8.5. This breaks and returns the error type "schema.x_enum" does not exist. 

## Steps to Reproduce

1. Create an enum in a separate schema
2. Try to query it

### Expected Behavior

It's queriable

### Actual Behavior

Type doesn't exist

### Reproduces How Often

Yes

### Workarounds

I was able to work around it for now but setting the search_path to the schema, however, I am concerned about duplicate types if it ever shows up.

nils-degroot:
Hiya. I'm running a database with many would like to specify where which enum lives. Currently a workaround would be to do something like `enum_name = "some_schema\".\"some_enum"`. This PR adds a `schema_name` field to the `DeriveActiveEnum` to properly support this. By default it will be empty mirroring the current behavior.

## PR Info

- Closes #2581

## New Features

- [ ] Add the `schema_name` field to `DeriveActiveEnum`


tests/active_enum_tests.rs
```diff
@@ -1187,4 +1187,130 @@ mod tests {
...
+            [
+                r#"SELECT "schema_enum"."id","#,
+                r#"CAST("schema_enum"."mood" AS "text"),"#,
+                r#"CAST("schema_enum"."priority" AS "text")"#,
+                r#"FROM "my_schema"."schema_enum""#,
+                r#"WHERE "schema_enum"."mood" = 'Happy'::"mood""#,
+            ]
+            .join(" ")
+        );
+    }
```
Huliiiiii: For type casts, the current implementation is missing the schema.

tests/active_enum_tests.rs
```diff
@@ -1187,4 +1187,130 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            db_postgres.build(&schema.create_enum_from_active_enum::<Priority>().unwrap()),
+            Statement::from_string(
+                db_postgres,
+                r#"CREATE TYPE "my_schema"."priority" AS ENUM ('Low', 'High')"#.to_owned()
+            )
+        );
+    }
```
Huliiiiii: These two assertions seem to cover the same case.

tests/active_enum_tests.rs
```diff
@@ -1187,4 +1187,130 @@ mod tests {
             r#"SELECT CAST("aaa"."tea" AS "text") AS "tea", CAST("foo"."tea" AS "text") AS "nested_tea" FROM "public"."active_enum""#,
         );
     }
+
```
Huliiiiii: I think it would be better to split this into a separate file, since active_enum_tests.rs is already 1.2k loc.

Huliiiiii (changes requested):

sea-orm-macros/src/derives/active_enum.rs
```diff
@@ -795,6 +817,13 @@ impl ActiveEnum {
             #not_u8_impl
         )
     }
+
+    fn enum_name(&self) -> String {
```
Huliiiiii: `schema_qualified_name` is more accurate.

sea-orm-macros/src/derives/active_enum.rs
```diff
@@ -593,16 +611,20 @@ impl ActiveEnum {
         }
 
         let ident = &self.ident;
-        let enum_name = &self.enum_name;
         let ident_s = ident.to_string();
         let variant_idents = &self.variant_idents;
         let variant_values = &self.variant_values;
 
+        let pg_type_name = match &self.schema_name {
```
Huliiiiii: This can be replaced with the method you added below.

sea-orm-macros/src/derives/active_enum.rs
```diff
@@ -8,6 +8,7 @@ use syn::{Expr, Lit, LitInt, LitStr, UnOp, parse};
 struct ActiveEnum {
     ident: syn::Ident,
     enum_name: String,
+    schema_name: Option<String>,
```
Huliiiiii: I think changing this to `(Option<String>, String)` would make it harder to forget `schema_name` when using it.

Huliiiiii (changes requested):

tests/common/features/sea_orm_active_enums.rs
```diff
@@ -66,3 +66,31 @@ pub enum DisplayTea {
...
+    db_type = "Enum",
+    enum_name = "priority",
+    schema_name = "my_schema"
+)]
+pub enum Priority {
+    #[sea_orm(string_value = "Low")]
+    Low,
+    #[sea_orm(string_value = "High")]
+    High,
+}
```
Huliiiiii: Could you explain why two enums are needed here?

nils-degroot: They don't need to be public. I've added them to the test file itself with the entity.

PR #3058, opened 2026-05-06
feat: implement `CursorTrait` for `SelectFour`

zefr0x:
## PR Info

- Closes 

## New Features

- [x] Implement `cursor_by()` for `SelectFour`


zefr0x:
We still need tests for this, so it will stay as a draft PR.

I don't know why it wasn't implemented or if my implementation is valid. I just copied `SelectThree`'s impl and tuned it to handle four tables. Used it in my project and it seem to work.

PR #3059, opened 2026-05-07
fix: delete_by_* for unique keys with 1 column

sefidel:
## PR Info

Fixes the issue where [the documentation](https://www.sea-ql.org/SeaORM/docs/basic-crud/delete/#delete-by-unique-key) specifies usage of `delete_by_*` with one-column unique keys, but `sea-orm-macros` only generated them for unique keys where `columns > 1`.

## Bug Fixes

- [x] Generate `delete_by_*` functions for unique keys with one column

---
p.s. Should we have a test case for `delete_by_*` functions in `tests/active_model_ex_tests.rs`?


Huliiiiii:
Yes, we should add a test.

@tyt2y3 Why did we skip the case where column len = 1 at first?

PR #3071, opened 2026-05-25
Use `web-time::Instant` instead of `std::time::Instant` for wasm32 target

Huliiiiii:
- Closes #3068

## Bug Fixes

- [x] Use wasm-compatible `Instant` for wasm32 target


Expurple (approved):
Thanks

PR #3076, opened 2026-06-06
Newtype primary keys

### LINKED ISSUE #3066: Type-safe primary keys
## Motivation

SeaORM today represents an entity's primary key as the raw scalar type; `i32`,
`i64`, `Uuid`, `String`. That makes every PK structurally identical to every
other PK of the same scalar, and mixing them up won't be caught by the type system.

There are several areas this failure mode can occur:

### 1. Untyped API usage

Concretely, today's API looks like this:

```rust
// Two entities, same scalar PK type.
mod user { pub struct Model { pub id: i32, pub name: String, /* … */ } }
mod post { pub struct Model { pub id: i32, pub title: String, /* … */ } }

let post: post::Model = /* … */;
let owner = user::Entity::find_by_id(post.id).one(db).await?; // logic error
```

This compiles, runs, and silently looks up the user whose id happens to equal a post id.

### 2. Untyped domain code

The same shape shows up in domain code:

```rust
fn ban_user(db: &DbConn, user_id: i32) { /* … */ }

let post: post::Model = /* … */;
ban_user(db, post.id);    // compiles, bans the wrong account
```

### 3. Multiple FK to the same column

The other recurring failure mode is junction tables that reference the same parent twice. For example, a "follower" join with both `user_id` and `follower_id`. All four columns are `i32` (or `i64`). Position is the only thing distinguishing them:

```rust
user_follower::ActiveModel {
    user_id: Set(follower.id),   // swapped
    follower_id: Set(user.id),   // swapped
    ..Default::default()
}.insert(db).await?;             // compiles, records the inverse relationship
```

Foreign-key columns are equally exposed. A `comment::Model` carries `post_id: i32` and `user_id: i32`. Any code that builds, edits, or filters comments can swap the two with no compiler complaint.

## Proposed Solution

Use a core type `Id<E, T>` that gets generated for entities by the CLI, and integrate it with FK columns, and find_by_id and similar methods.

### 1. Codegen emits one-line aliases per table

```rust
// Generated by `sea-orm-cli generate entity --with-pk-newtypes`:
pub type CakeId = sea_orm::Id<Entity, i32>;

#[derive(DeriveEntityModel, …)]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: CakeId,        // the alias, not the raw scalar
    pub name: String,
    /* … */
}
```

Foreign-key columns spell the parent's alias:

```rust
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: CommentId,
    pub post_id: super::post::PostId,   // not raw i32
    pub user_id: super::user::UserId,
    /* … */
}
```

### 2. Role wrappers for self-referential junctions

When a table has more than one FK column pointing at the same parent, codegen emits per-column wrapper structs around the parent's alias so positional swaps fail to compile:

```rust
#[derive(DeriveValueType, …)]
pub struct UserFollowerPkUserId(pub super::user::UserId);
#[derive(DeriveValueType, …)]
pub struct UserFollowerPkFollowerId(pub super::user::UserId);

pub struct Model {
    #[sea_orm(primary_key)]
    pub user_id: UserFollowerPkUserId,
    #[sea_orm(primary_key)]
    pub follower_id: UserFollowerPkFollowerId,
    /* … */
}
```

Now the swapped insert from the motivation section is a compile error:

```rust
user_follower::ActiveModel {
    user_id: Set(UserFollowerPkFollowerId(follower.id)),  // type mismatch
    follower_id: Set(UserFollowerPkUserId(user.id)),      // type mismatch
    ..Default::default()
};
```

### 3. `find_by_id` / `filter_by_id` are a single permissive function

These functions can handle both typed and untyped entities:

```rust
fn find_by_id<T>(values: T) -> Select<Self>
where T: Into<<Self::PrimaryKey as PrimaryKeyTrait>::ValueType>
```

### 4. Opt-in via CLI flag

The whole feature is gated behind `sea-orm-cli generate entity --with-pk-newtypes`. Existing projects don't get newtypes unless they regenerate with the flag. Hand-written entities should be able to adopt incrementally by converting one table at a time by replacing `pub id: i32` with a one-line alias.

---

I've been working on a prototype for this that I'll work to make a draft PR soon.

AngelOnFira:
## PR Info

- Closes #3066 

### 1. Opt-in type-safe primary keys via newtypes

Adds a code-generation mode (`sea-orm-cli generate entity --with-pk-newtypes`) plus a library surface (`sea_orm::Id<E, T>`, `PkAutoIncrementHint`, `FindByIdArg`) that together give compile-time protection against mixing up primary-key values across entities. With the flag, `user::Entity::find_by_id(post.id)` no longer compiles because `post.id` is `PostId` (`Id<post::Entity, i32>`), which has no `Into<Id<user::Entity, i32>>` impl by design.

### 2. `sea_orm::Id<E, T>` library type

Phantom-typed PK wrapper, unique for each table's primary key. The only construction path is `Id::new(value)`, there is deliberately no `impl<E, T> From<T> for Id<E, T>` blanket, which is what makes the compiler reject cross-entity confusion at every use site.

### 3. `FindByIdArg<E>` trait

The shared bound on `find_by_id` / `filter_by_id` / `delete_by_id`. A thin sea-orm-owned wrapper around `Into<<E::PrimaryKey as PrimaryKeyTrait>::ValueType>` so we can attach a curated `#[diagnostic::on_unimplemented]` message.

### 4. `PkAutoIncrementHint` trait

Trait-resolved default for whether a PK column is `AUTO_INCREMENT` (integers -> `true`, `String`/`Vec<u8>`/`Uuid` -> `false`). `DeriveValueType` emits an impl so the new PK wrappers will resolve through their inner type. This will replace the previous text-search method.

## Breaking Changes

- None

Notes: Backwards-compatible: nothing changes for entities generated without `--with-pk-newtypes`. `find_by_id(7u8)` against an `i32` PK still works via the `Into`-forwarding blanket impl, and the `auto_increment()` body now resolves through `PkAutoIncrementHint` while producing the same defaults as before. `DeleteMany::filter_by_ids` (plural) keeps its pre-PR `IntoIterator<Item = ValueType>` bound to avoid unrelated behavior changes.

## Changes

- Codegen resolution chain. Per column, the emitted Rust type is resolved in a fixed precedence: (1) role wrapper for PK columns of junction tables with multiple FKs to the same parent (`UserFollowerPkUserId`), (2) single-parent FK column -> parent's PK alias, (3) own-PK alias (`CakePk`), (4) raw scalar fallback. `Column::refs: Vec<ColumnRef>` is populated from FK constraints (multi-parent supported).
- `auto_increment()` macro body: composite PK -> `false`; explicit annotation -> literal bool; otherwise `<FieldType as PkAutoIncrementHint>::IS_AUTO`. `DeriveValueType` emits `DelegatesPkAutoIncrementHint`.
- `--with-pk-newtypes` CLI flag (`sea-orm-cli/src/cli.rs`)
- `find_by_id` / `delete_by_id` / `filter_by_id` now take `T: FindByIdArg<Self>`

### Out of scope

- Multi-parent FK columns aren't handled, and instead fall back to raw scalars


Huliiiiii:
Could you split this into smaller PRs or commits? A 4k-loc change is quite difficult to review.

AngelOnFira:
Thanks for the suggestion @Huliiiiii! I have a few things I want to improve before setting this draft PR ready to review, so I'm happy to split the PR in places that make sense. I'm open for suggestions, here's the current breakdown;

- ~650 LoC new example in `examples/basic_typed_pk/`
- ~860 LoC `sea-orm-codegen/` 
- ~500 LoC new core code
- ~670 LoC new tests
- ~1.1k Loc duplicate code in `sea-orm-sync/` (duplicates tests + core code)

I'm not sure I can easily split the core functionality into multiple PRs, but it would be easy to split this up into separate commits. I can also move `examples/basic_typed_pk/` to another PR since that's just a demo of the feature.

If it seems like just making more commits won't help the problem, I can drill deeper into separating functionality into multiple PRs. Let me know what you think!

Huliiiiii:
I think you can keep going with the current implementations for now.

If it ends up being too hard to review once it’s ready, I’ll ask you to split it up then.

PR #3083, opened 2026-06-12
set_unset & friends for `ActiveValue`

### LINKED ISSUE #3077: "Set if unset" methods for ActiveValues
## Motivation

`ActiveValues` lack a "null coalescing" operation that is common for ORMs in other languages (as well as `Options` in Rust with `unwrap_or`). Sometimes you need to set defaults in a model, but only when the field is unset. Currently, that requires making an if block, which is very ugly and verbose.

## Proposed Solutions

A pair of new methods on `ActiveValue`, `set_unset`, `set_unset_with`, and `set_unset_default`, which will work sort of like `unwrap_or`, `unwrap_or_else`, and `unwrap_or_default` respectively.

For an `ActiveValue<T>`:
```rust
fn set_unset(&mut self, val: T) {
    match self {
        ActiveValue::Unset => self.set(val),
        _ => ();
    };
}

fn set_unset_with(&mut self, f: fn() -> T) {
    match self {
        ActiveValue::Unset => self.set(f()),
        _ => ();
    };
}


fn set_unset_default(&mut self)
where 
    T: Default
{
    match self {
        ActiveValue::Unset => self.set(T::default()),
        _ => ();
    };
}
```

PacificBird:
## PR Info

- Closes #3077 

## New Features

- `ActiveValue::set_unset(&mut self, value: V)`, which mutates `self` to `ActiveValue::Set(value)` if `self.is_not_set()`
- `ActiveValue::set_unset(&mut self, value: impl FnOnce() -> V)`, which mutates `self` to `ActiveValue::Set(f())` if `self.is_not_set()`
- `ActiveValue::set_unset(&mut self) where V: Default`, which mutates `self` to `ActiveValue::Set(V::default())` if `self.is_not_set()`

Includes documentation and examples/doc tests for all three. I cannot overstate how much grief these changes will save me in my production code base.


Huliiiiii:
I'd prefer the names set_if_unset / not_set, even though they’re a bit more verbose.
What do you think? @tyt2y3

PacificBird:
Can't run cargo clippy because I'm getting 
```
conflicting implementations of trait `From<format_description::parse::format_item::HourBase>` for type `<format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type
```
all over the place

Issue #2995, opened 2026-03-10
Column `extra` causes `sync` to add column again


nitroxis:
## Description

Using `#[sea_orm(extra = ...)]` for generated columns in sqlite causes the column to be added again on next `sync`.

## Steps to Reproduce

Defining an entity like this:

```rs
use sea_orm::entity::prelude::*;

#[sea_orm::model]
#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
#[sea_orm(table_name = "photo")]
pub struct Model {
	#[sea_orm(primary_key)]
	pub id: i64,
	pub width: u32,
	pub height: u32,
	#[sea_orm(extra = "GENERATED ALWAYS AS (width * height) VIRTUAL")]
	pub pixels: u32,
}

impl ActiveModelBehavior for ActiveModel {}
```

And then using `db.get_schema_registry("entity::*").sync(db).await?;` seemingly creates the table correctly, but will try to add the `pixels` column again when running `sync` again later.

### Expected Behavior

It shouldn't try to add the column again.

### Actual Behavior

In the trace I can see that it attempts to execute
```sql
ALTER TABLE "photo" ADD COLUMN "pixels" integer NOT NULL GENERATED ALWAYS AS (width * height) VIRTUAL
```
even though it's unnecessary, causing `Execution Error: error returned from database: (code: 1) duplicate column name: pixels`

### Reproduces How Often

Every time `sync` is called.

### Workarounds

Dropping the column beforehand will recreate it again without error.

A proper way to define generated/computed columns would be great as well but it seems to me this should work just fine.

## Versions

v2.0.0-rc.36

sinder38:
I think I can fix this one.

PR #3095, opened 2026-06-23
Skip SQLite generated columns in `generate entity`

### LINKED ISSUE #3094: sea-orm-cli generate entity emits SQL generated columns (VIRTUAL/STORED) as writable fields
## Description

`sea-orm-cli generate entity` now emits **SQL generated columns** (`GENERATED ALWAYS AS (...) VIRTUAL` / `STORED`) as ordinary, writable entity fields. Earlier 2.0 release candidates omitted them entirely.

Because the emitted field carries no marker distinguishing it from a normal column, it becomes part of the derived `ActiveModel`. Any INSERT/UPDATE that sets the field — most commonly via `model.into_active_model()` (which sets *every* field) — sends the generated column to the database, and SQLite rejects it:

```
cannot INSERT into generated column "translation"
cannot UPDATE generated column "translation"
```

A generated column is read-only at the storage layer, so the generated entity should not expose it as a writable column. The previous behavior (skipping it) was correct for our usage.

## Steps to Reproduce

1. Create a SQLite database with a virtual generated column:

   ```sql
   CREATE TABLE node (
       id          INTEGER PRIMARY KEY,
       x           REAL,
       y           REAL,
       z           REAL,
       translation TEXT GENERATED ALWAYS AS (json_array(x, y, z)) VIRTUAL
   );
   ```

2. Generate entities:

   ```bash
   sea-orm-cli generate entity -u 'sqlite://./test.sqlite' -o ./out
   ```

3. Inspect `out/node.rs`.

### Expected Behavior

The generated column should not be exposed as a writable column. Either omit it (as earlier RCs did), or generate it in a form that is excluded from INSERT/UPDATE — e.g. apply `#[sea_orm(ignore)]`, which is the maintainer-recommended workaround for generated columns (discussion #2420). Ideally there would be a documented, supported way to keep such a column readable while never writing it.

### Actual Behavior

`translation` is generated as a plain, writable column:

```rust
#[sea_orm(column_type = "Text", nullable)]
pub translation: Option<String>,
```

It is added to the `Column` enum and the derived `ActiveModel`, so writing the model fails at the database as shown above.

### Reproduces How Often

Always, for any `VIRTUAL`/`STORED` generated column on SQLite.

### Workarounds

- Manually add `#[sea_orm(ignore)]` to each generated-column field after codegen (per discussion #2420). This stops the write, but the field is then excluded from SELECT as well (filled via `Default`), and the edit has to be re-applied on every regeneration since the CLI re-emits the plain column.
- Pin `sea-orm-cli` to `2.0.0-rc.34`, which did not emit these columns.

There is no column-level exclusion flag for `generate entity` (only `--ignore-tables`), so neither workaround can be expressed through the CLI.

### Suggestion

It would help to either restore skipping of generated columns by default, or add a flag / emit `#[sea_orm(ignore)]` (or an equivalent read-only marker) automatically for columns reported as generated by the schema discovery (`PRAGMA table_xinfo` exposes the `hidden = 2/3` flag for VIRTUAL/STORED on SQLite).

## Versions

- `sea-orm` / `sea-orm-cli` / `sea-orm-migration`: `2.0.0-rc.41` (regression vs `2.0.0-rc.34`, which omitted these columns)
- `sea-schema`: `0.18.0`
- `sea-query`: `1.0.1`
- Database: SQLite
- OS: Linux

---
<sub>Filed by an AI agent (Claude Code, Anthropic) on behalf of, and reviewed by, a maintainer of a downstream project that hit this on a routine `sea-orm-cli` bump.</sub>

JMLX42:
## PR Info

- Closes: #3094
- Related: SeaQL/sea-schema#161, #2995

## Bug Fixes

- [x] `sea-orm-cli generate entity` no longer reverse-generates SQLite generated columns (`GENERATED ALWAYS AS (...) VIRTUAL`/`STORED`) as ordinary, writable entity fields.

## Description

SQLite generated columns became discoverable when discovery switched from `PRAGMA table_info` to `PRAGMA table_xinfo` in SeaQL/sea-schema#161 — a correct fix for schema sync (#2995), where generated columns were previously invisible and `sync` kept re-issuing `ALTER TABLE ... ADD COLUMN`.

As a side effect, `generate entity` consumes the same discovery and now emits these columns as plain fields:

```rust
#[sea_orm(column_type = "Text", nullable)]
pub pixels: Option<i32>,
```

Because there is no marker distinguishing them, they land in the derived `ActiveModel`. Any `Model::into_active_model()` round-trip then includes them in `INSERT`/`UPDATE`, and SQLite rejects it:

```
cannot INSERT into generated column "pixels"
cannot UPDATE generated column "pixels"
```

Generated columns are read-only at the storage layer, so they should not be reverse-generated as writable columns. In the Entity-First workflow they are declared with `#[sea_orm(extra = "GENERATED ALWAYS AS (...) VIRTUAL")]`, not produced by reverse codegen.

## Changes

- Add `discard_generated_columns()` in `sea-orm-cli`'s `generate` command. In the SQLite branch, each discovered table is passed through it before `TableDef::write()`, removing columns whose `ColumnVisibility` is `GeneratedVirtual` or `GeneratedStored`.
- Discovery itself is untouched, so schema sync keeps seeing generated columns and the #2995 fix is preserved — only the reverse-codegen path drops them.
- Unit test `test_discard_generated_columns` covering both `VIRTUAL` and `STORED` columns and confirming ordinary columns are preserved in order.

This restores the pre-`rc.34` `generate entity` behavior for generated columns. A faithful round-trip (emitting `#[sea_orm(extra = "GENERATED ...")]` and marking the field read-only) would be a larger, cross-crate change — `table_xinfo` does not return the generation expression — and is noted as a follow-up in #3094.

---
<sub>Authored by an AI agent (Claude Code, Anthropic), reviewed by a maintainer of a downstream project that hit this on a routine `sea-orm-cli` bump.</sub>

### DIFF
--- a/sea-orm-cli/src/commands/generate.rs
+++ b/sea-orm-cli/src/commands/generate.rs
@@ -249,6 +249,7 @@ pub async fn run_generate_command(
                             .filter(|schema| filter_tables(&schema.name))
                             .filter(|schema| filter_hidden_tables(&schema.name))
                             .filter(|schema| filter_skip_tables(&schema.name))
+                            .map(discard_generated_columns)
                             .map(|schema| schema.write())
                             .collect();
                         (None, table_stmts)
@@ -461,6 +462,29 @@ impl From<BannerVersion> for CodegenBannerVersion {
     }
 }
 
+/// Drop SQL generated columns (`GENERATED ALWAYS AS (...) VIRTUAL`/`STORED`) from a
+/// discovered SQLite table before it is reverse-generated into an entity.
+///
+/// Such columns are read-only at the storage layer: SQLite rejects any `INSERT`/`UPDATE`
+/// that targets them. Emitting them as ordinary entity fields makes them part of the
+/// derived `ActiveModel`, so a `Model::into_active_model()` round-trip would try to write
+/// them and fail. They are discovered on purpose (schema sync relies on it — see
+/// SeaQL/sea-schema#161 / #2995), but they must not be reverse-generated as columns.
+#[cfg(feature = "sqlx-sqlite")]
+fn discard_generated_columns(
+    mut table: sea_schema::sqlite::def::TableDef,
+) -> sea_schema::sqlite::def::TableDef {
+    use sea_schema::sqlite::def::ColumnVisibility;
+
+    table.columns.retain(|column| {
+        !matches!(
+            column.hidden,
+            ColumnVisibility::GeneratedVirtual | ColumnVisibility::GeneratedStored
+        )
+    });
+    table
+}
+
 #[cfg(test)]
 mod tests {
     use clap::Parser;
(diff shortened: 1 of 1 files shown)


tyt2y3:
Thanks for this, and sorry for the overlap — master already fixes #3094 via 27956a3c ("Skip generated columns in sea-orm-cli generate entity"), which landed just after you opened this. It takes the same approach (`sqlite_column_is_generated` + `retain`) and also covers the MySQL/Postgres branches, so this PR is now superseded. Closing as already-fixed. Appreciate the thorough writeup.

PR #3096, opened 2026-06-24
Fix sea-orm-cli generating duplicated columns for multi-column foreign keys

### LINKED ISSUE #2662: sea-orm-cli generates incorrect relation for multi-column foreign key
## Description

When a Postgres table has a foreign key constraint that includes more than one column, `sea-orm-cli` generates a relation with each column in the foreign key duplicated.  This leads to sea-orm generating SQL joins that cannot possibly match any rows.

## Steps to Reproduce

1. Create a two-table schema where the second table references two columns in the first table.
2. Run `sea-orm-cli generate entity` on the database.
3. Observe that the `Relation` enum generated for the second table has the columns duplicated in the `from` and `to` bits in the `sea_orm` macro attribute.
4. Try to do an inner join using sea-orm and see that the `ON` clause in the generated SQL has *four* equality constraints rather than two; two of them are correct, but the other two try to assert equality between non-corresponding columns.

### Expected Behavior

`sea-orm-cli generate entity` should generate the `Relation` enum without repeating the columns multiple times. 

### Actual Behavior

Incorrect `Relation` generated leading to incorrect SQL join queries.

### Reproduces How Often

Reproduces always

### Workarounds

Manually editing the generated code can fix the problem.

## Reproducible Example

I'll work up something more formal if needed, but here's a simple DB schema:

```sql
CREATE TABLE IF NOT EXISTS "first" (
    "auto_id" bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY NOT NULL,
    "a" varchar NOT NULL,
    "b" varchar NOT NULL
);
CREATE UNIQUE INDEX IF NOT EXISTS "idx_first_ab" ON "first" ("a", "b");

CREATE TABLE IF NOT EXISTS "second" (
    "auto_id" bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY NOT NULL,
    "f_a" varchar NOT NULL,
    "f_b" varchar NOT NULL,
    "c" varchar NOT NULL,
    CONSTRAINT "fk_ab" FOREIGN KEY ("f_a", "f_b") REFERENCES "first" ("a", "b") ON DELETE CASCADE
);
```

(The intent here is that there can be at most one row in `first` with the same `a`+`b`, and then there can be multiple rows in `second` with the same `f_a`+`f_b` that matches `a`+`b` in `first`, but with different values for `c`.)

This gives us the following as seen via `psql`:

```
                               Table "public.first"
 Column  |       Type        | Collation | Nullable |           Default            
---------+-------------------+-----------+----------+------------------------------
 auto_id | bigint            |           | not null | generated always as identity
 a       | character varying |           | not null | 
 b       | character varying |           | not null | 
Indexes:
    "first_pkey" PRIMARY KEY, btree (auto_id)
    "idx_first_ab" UNIQUE, btree (a, b)
Referenced by:
    TABLE "second" CONSTRAINT "fk_ab" FOREIGN KEY (f_a, f_b) REFERENCES first(a, b) ON DELETE CASCADE

                               Table "public.second"
 Column  |       Type        | Collation | Nullable |           Default            
---------+-------------------+-----------+----------+------------------------------
 auto_id | bigint            |           | not null | generated always as identity
 f_a     | character varying |           | not null | 
 f_b     | character varying |           | not null | 
 c       | character varying |           | not null | 
Indexes:
    "second_pkey" PRIMARY KEY, btree (auto_id)
Foreign-key constraints:
    "fk_ab" FOREIGN KEY (f_a, f_b) REFERENCES first(a, b) ON DELETE CASCADE
```

Running `sea-orm-cli generate entity` on this gives us the following inside `second.rs`:

```rust
pub enum Relation {
    #[sea_orm(
        belongs_to = "super::first::Entity",
        from = "(Column::FA, Column::FA, Column::FB, Column::FB)",
        to = "(super::first::Column::A, super::first::Column::B, super::first::Column::A, super::first::Column::B)",
        on_update = "NoAction",
        on_delete = "Cascade"
    )]
    First,
}
```

Note the duplication in `from` and `to`.  With sea-orm code like the following:

```rust
First::find()
    .inner_join(Second)
    .select_also(Second)
    .filter(...)
```

... we get a generated query with a join `ON` clause that looks like:

```sql
ON \
    "first"."a" = "second"."f_a" AND \
    "first"."b" = "second"."f_a" AND \
    "first"."a" = "second"."f_b" AND \
    "first"."b" = "second"."f_b"
```

The middle two equality bits will of course never work, and the query returns one row for each row in `first`, but with all of the columns in `second` empty.

If I then edit the generated code so the attributes on the macro instead read:

```
from = "(Column::FA, Column::FB",
to = "(super::first::Column::A, super::first::Column::B)",
```

... then I get generated queries with the correct join `ON` clause:

```sql
ON \
    "first"."a" = "second"."f_a" AND \
    "first"."b" = "second"."f_b"
```

... and I get back the rows I expect.

## Versions

* sea-bae v0.2.1
* sea-orm v1.1.13
* sea-orm-cli v1.1.13
* sea-orm-macros v1.1.13
* sea-orm-migration v1.1.13
* sea-query v0.32.6
* sea-query-derive v0.4.2
* sea-query-binder v0.7.0
* sea-schema v0.16.2
* sea-schema-derive v0.3.0

OS: Debian trixie
Database: PostgreSQL 17.0 (via postgres:17-alpine Docker container)

kenkoooo:
`sea-orm-cli generate entity` generated a malformed `Relation` for a multi-column foreign key, duplicating columns in the `from`/`to` attributes and producing SQL joins whose `ON` clause matches no rows. This de-duplicates the foreign-key column lists when building the relation, restoring the intended 1:1 column pairing.

## PR Info

- Closes #2662

## New Features

- N/A

## Bug Fixes

- [x] **Multi-column foreign keys that reference a bare unique index generated duplicated columns.**

  When a FK targets a bare unique index (`CREATE UNIQUE INDEX …`) rather than a named `UNIQUE` / `PRIMARY KEY` constraint, PostgreSQL's `information_schema` does not expose the referenced constraint (`referential_constraints.unique_constraint_name` is empty and the columns are absent from `key_column_usage`). Schema discovery cannot correlate the local and referenced columns by ordinal position and falls back to their **cartesian product** — local `[f_a, f_b]` paired with referenced `[a, b, a, b]`. `From<&TableForeignKey> for Relation` stored those lists verbatim, so the generated `belongs_to` paired two `from` columns against four `to` columns:

  ```rust
  // before
  from = "(Column::FA, Column::FB)",
  to   = "(super::first::Column::A, super::first::Column::B, super::first::Column::A, super::first::Column::B)",
  // after
  from = "(Column::FA, Column::FB)",
  to   = "(super::first::Column::A, super::first::Column::B)",
  ```

  Fixed by de-duplicating both column lists (preserving first-seen order). A well-formed never repeats a column on either side, so it is a no-op for valid input and only repairs the cartesian-product degeneracy. (The root cause is in sea-schema's discovery query; this is a defensive codegen-side fix that a future sea-schema change could complement.)

**Breaking Changes**

- None — backward compatible. The de-duplication is a no-op for well-formed foreign keys, so existing correct entities regenerate identically; only the previously-malformed output changes.

**Changes**

- [x] De-duplicate the local and referenced column lists (order-preserving) in From<&TableForeignKey> for Relation (sea-orm-codegen).
- [x] Regression tests: a unit test for the de-duplication, plus end-to-end transformer tests for the rendered belongs_to (2- and 3-column keys) and the inverse HasOne / HasMany classification.

Verified with cargo test --workspace and cargo test --manifest-path sea-orm-cli/Cargo.toml, and by running sea-orm-cli generate entity against a live PostgreSQL 17 instance with the issue's exact schema.

### DIFF
--- a/sea-orm-codegen/src/entity/relation.rs
+++ b/sea-orm-codegen/src/entity/relation.rs
@@ -208,14 +208,36 @@ impl Relation {
     }
 }
 
+/// Remove duplicated columns while preserving their first-seen order.
+///
+/// A foreign key references each of its columns exactly once on both the local
+/// and the referenced side, so de-duplication is always a no-op for a
+/// well-formed key. It only takes effect to repair malformed input produced by
+/// schema discovery: when a foreign key references a bare unique *index* (rather
+/// than a named unique / primary-key constraint), PostgreSQL's `information_schema`
+/// does not expose the referenced constraint, so `sea-schema` cannot correlate the
+/// columns by ordinal position and falls back to emitting their cartesian product
+/// (e.g. local `[f_a, f_b]` paired with referenced `[a, b, a, b]`). Collapsing both
+/// sides back to their distinct columns restores the intended 1:1 pairing. See
+/// <https://github.com/SeaQL/sea-orm/issues/2662>.
+fn dedup_preserving_order(columns: Vec<String>) -> Vec<String> {
+    let mut deduped: Vec<String> = Vec::with_capacity(columns.len());
+    for column in columns {
+        if !deduped.contains(&column) {
+            deduped.push(column);
+        }
+    }
+    deduped
+}
+
 impl From<&TableForeignKey> for Relation {
     fn from(tbl_fk: &TableForeignKey) -> Self {
         let ref_table = match tbl_fk.get_ref_table() {
             Some(s) => s.sea_orm_table().to_string(),
             None => panic!("RefTable should not be empty"),
         };
-        let columns = tbl_fk.get_columns();
-        let ref_columns = tbl_fk.get_ref_columns();
+        let columns = dedup_preserving_order(tbl_fk.get_columns());
+        let ref_columns = dedup_preserving_order(tbl_fk.get_ref_columns());
         let rel_type = RelationType::BelongsTo;
         let on_delete = tbl_fk.get_on_delete();
         let on_update = tbl_fk.get_on_update();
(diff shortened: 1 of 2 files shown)


Huliiiiii:
Is it possible to fix this directly in sea-schema?

kenkoooo:
@Huliiiiii 
Thanks for the suggestion! I've opened a PR on sea-schema to fix it at the source: https://github.com/SeaQL/sea-schema/pull/168
Let me close this PR.

PR #3097, opened 2026-06-25
Move `count()` to `CountTrait`

Huliiiiii:
## Breaking Changes

- [x] Move the count method into `CountTrait` to prevent accidental resets of limit and offset


tyt2y3:
let's not create churn right now. but if we have new methods, say `count_distinct` we should put them in a new trait, or reuse `SelectExt`

PR #3113, opened 2026-07-03
Add ActiveBelongsToNotNull + nullable belongs_to detach

tyt2y3:
Follow-up to #3110.

1. **Rename** (`026dfe1e`) — mechanical, word-boundary rename crate-wide:
   - `HasOneModel` -> `ActiveHasOne`
   - `HasManyModel` -> `ActiveHasMany`

   These are the active/write-side companions to `HasOne`/`HasMany`; the old `*Model` names read as loaded models. No behavior change.

2. **`ActiveBelongsToNotNull` + nullable detach** (`af16e86f`):
   - New type `ActiveBelongsToNotNull<E>` with variants `NotSet` / `Set` — **no `Delete`**. It is the active-side type for a **non-nullable `belongs_to`**, so trying to detach a relation whose self-FK cannot be nulled is now a **compile error** instead of a silent no-op / raw SQL constraint error at runtime.
   - The derive macro inspects each `belongs_to`'s self-FK column nullability:
     - non-nullable FK -> field typed `ActiveBelongsToNotNull` (no `delete_*` / `set_*_option` builders generated)
     - nullable FK -> field typed `ActiveHasOne`, and `Delete` now **detaches** by setting the self-FK to `NULL` (previously a no-op)
   - `from` is snake_cased before lookup, so both a Column name (`BakeryId`) and a field name (`user_id`) resolve to the FK field. Composite / unresolvable `from` values fall back to `ActiveHasOne` with no detach (keeps them compiling).

Previously, calling delete/detach on a non-nullable `belongs_to` silently did nothing (or surfaced a raw DB constraint error). This makes the invariant type-level for `belongs_to` — the only relation whose FK nullability is visible to the derive macro — while keeping codegen concrete per field so downstream users don't need generic bounds.

### DIFF
--- a/sea-orm-macros/src/derives/active_model_ex.rs
+++ b/sea-orm-macros/src/derives/active_model_ex.rs
@@ -1,6 +1,7 @@
 use super::active_model::DeriveActiveModel;
 use super::attributes::compound_attr;
 use super::util::{extract_compound_entity, field_not_ignored_compound, is_compound_field};
+use heck::ToSnakeCase;
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
 use std::collections::HashMap;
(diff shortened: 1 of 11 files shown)


tyt2y3:
Superseding this in favor of #3118

PR #3116, opened 2026-07-04
Fix: Fixed `has_related` condition building when providing a `Condition::any()` condition

### LINKED ISSUE #3115: `has_related` with `Condition::any()` end up generating a query with `OR A.id = B.Id` instead of `AND`
## Description

When building a query using `.has_related(entity, Condition::any().add(cond1).add(cond2))`, the generated SQL statement adds an `WHERE cond1 OR cond2 OR A.id = B.a_id` clause in the `EXISTS` subquery. This leads to incorrect results for the query.  

## Steps to Reproduce

1. Create two related entities using a One-To-Many relation
2. Create a query using `has_related` and a `Condition::any().add(cond1).add(cond2)`
3. Execute query

### Expected Behavior

The generated `EXISTS` subquery should have a `WHERE (cond1 OR cond2) AND A.id = B.a_id` clause, leading to the correct selection of rows

### Actual Behavior

The generated `EXISTS` subquery shows `WHERE cond1 OR cond2 OR A.id = B.a_id clause, leading to incorrect selection of rows

### Reproduces How Often

Reproduces every time

### Workarounds

Wrapping the `Condition::any()` in a `Condition::all()` seems to generating the correct query. Looking at the `src/query/join.rs` source file, the assumption made is that the provided condition will always be a single column condition, which will be cast to a `Conditon::all()` condition with the `.into_collection()` call. Unfortunately, providing a `Condition::any()` condition, the `condition.add()` call willnot change the condition type...

## Versions

sea-orm 2.0.0-rc40
SQLite
Linux Fedora 44

thuttinpasseron:
## PR Info

- Closes #3115

## Bug Fixes

- [x] #3115 - Wraps the generated condition in a `Condition::all()` condition to avoid any issue when adding the relation clause

### DIFF
--- a/sea-orm-sync/src/query/join.rs
+++ b/sea-orm-sync/src/query/join.rs
@@ -184,7 +184,7 @@ where
         C: IntoCondition,
     {
         let mut to = None;
-        let mut condition = condition.into_condition();
+        let mut condition = Condition::all().add(condition.into_condition());
         condition = condition.add(if let Some(via) = E::via() {
             to = Some(E::to());
             via
--- a/src/query/join.rs
+++ b/src/query/join.rs
@@ -184,7 +184,7 @@ where
         C: IntoCondition,
     {
         let mut to = None;
-        let mut condition = condition.into_condition();
+        let mut condition = Condition::all().add(condition.into_condition());
         condition = condition.add(if let Some(via) = E::via() {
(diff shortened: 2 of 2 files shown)


Huliiiiii (approved):

Huliiiiii:
Thanks!

PR #3118, opened 2026-07-07
Support non-null HasOne

Huliiiiii:
Adds support for non-nullable HasOne relations.

* Refactor a significant amount of the related code, as the current implementation was difficult to read and modify.
* Remove the `NotFound` representation for empty HasOne relations, as this state should never be possible for non-nullable relations.
* Remove the `Delete` variant added in the previous PRs, since the related behavior has now been fully replaced by `Set(Option::None)`.
* Remove the compatibility methods from the old implementation that automatically converted `Option` / `Vec` into `HasOne` / `HasMany`.

* Remove internal macro helpers
  * ActiveModelTrait::find_related_of
  * ActiveHasOne::as_slice
  * ActiveHasOne::empty_slice
  * ActiveHasOne::find
  * ActiveHasOne::replace


tyt2y3:
sadly, the API doesn't feel right. this is too much to write. has_one is always optional, and having `Option` as type parameter is explicit but kind of clunky.

```
#[sea_orm(has_one)]
pub profile: HasOne<Option<super::profile::Entity>>,
```

we tried to solve an edge case for belongs_to but it cascaded outwards

Issue #3119, opened 2026-07-07
Refactor macro implmentions


Huliiiiii:
I did some refactoring in #3118, but the current macro implementation is still quite messy.

The refactoring mainly split the code that mixed parsing and expansion into separate steps.

Ideally, we should remove `bae`. We could consider using `darling` instead, or just write the parsing code manually. The current macro parser supports all possible combinations, and I worked around that by adding an extra normalization step, but it would be better to clean up the related code.

There are also many other things that can be simplified. I noticed a lot of recurring patterns like duplicated logic across branches, using `parse(&str)` instead of `syn::parse_quote`, using manually constructed strings and `Ident`s instead of `format_ident`, and converting types such as `Type` or `Ident` into `String` for comparison and then converting them back, rather than comparing them directly.

I also noticed that some state can be derived rather than stored.

For contributors who want to help with this, I would suggest starting with small, focused changes, ideally by addressing one category of issues at a time. Furthermore, please don't open LLM-generated PRs.

shon0526:
Hi! I'd like to work on the "using `parse(&str)`instead of `syn::parse_quote!`" part of this issue. 

I noticed #3122 is already addressing a different part of this issue, replacing manually constructed `Ident`s with `format_indent!`. My undertstanding is that part I'd like to work on is separate from that PR, but please let me know if there is any potential overlap.

Before starting, I wanted to check whether anyone is already working on this part, and whether it would be okay for me to take it.

This would be my first contribution to this project, so any guidance would be appreciated.

Huliiiiii:
Note: Please start working after #3118 is merged.

TheShiveshNetwork:
Hi @Huliiiiii I see there has been no updates on this issue for long, I'm starting to work on it if there's no problem

Huliiiiii:
Sure, I've assigned it to you.

TheShiveshNetwork:
@Huliiiiii I've completed the migration from bae to darling. i just wanted to ask would it be fine if i draft a pr now and work on other refactoring as you mentioned in separate pr's? i also came across an issue that darling version `0.24` latest required syn version `3.0.0` but we're using syn `2`, for now i'm working with version `0.23` which seems stable. I could also work on that migration in another pr, if an issue can be opened.
Thanks, lmk whichever works

Huliiiiii:
If upgrading to syn 3 is easy enough, then upgrade to syn 3 first. Otherwise, we can use an older version of darling.

PR #3122, opened 2026-07-07
Use format_ident! instead of Ident::new(&format!(...)) in model_ex (#3119)

upuddu:
Small step toward #3119, which asks to clean up the macros one category at a time. This handles the "manually constructed `Ident`s" category: the two `Ident::new(&format!(...), span)` calls in `sea-orm-macros/src/derives/model_ex.rs` become `format_ident!(..., span = ...)`, matching how idents are already built elsewhere in the crate. Purely a readability change — the generated tokens are identical.


tyt2y3:
Thanks @upuddu — but master has moved past this. The `{model}Ex` line is already `format_ident!` (converted independently), and the `Reverse` construction this touched has since been refactored, hence the conflict. Master deliberately uses the call-site span for the generated `{model}Ex` ident, so re-adding `model.span()` isn't a change we want. One item from #3119's category does remain — the `Ident::new(&format!("{forward}Reverse"), ...)` at model_ex.rs:292 — worth a fresh one-liner if you'd like to keep chipping at #3119.

PR #3123, opened 2026-07-07
Add select_except

### LINKED ISSUE #3064: Add select_except
https://github.com/SeaQL/sea-orm/discussions/3026#discussioncomment-16994303
> I’d probably avoid `#[lazy]` on the normal `Model`.
> 
> If `file` is part of the model type but not actually loaded, then every access needs some hidden state like “loaded / not loaded / loading failed”. That makes the model less predictable and easy to misuse.
> 
> For this case I’d keep it explicit:
> 
> ```rust
> #[derive(Debug, FromQueryResult)]
> struct EntryWithoutFile {
>     id: i32,
>     name: String,
> }
> ```
> 
> ```rust
> let rows = Entity::find()
>     .select_only()
>     .column(Column::Id)
>     .column(Column::Name)
>     .into_model::<EntryWithoutFile>()
>     .all(db)
>     .await?;
> ```
> 
> Then load the blob only when needed:
> 
> ```rust
> let file: Option<Vec<u8>> = Entity::find_by_id(id)
>     .select_only()
>     .column(Column::File)
>     .into_tuple()
>     .one(db)
>     .await?;
> ```
> 
> So imo a `defer()` / `select_except()` style API would fit SeaORM better than lazy fields on `Model`.
> 
> Something like this would be really nice:
> 
> ```rust
> Entity::find()
>     .select_except([Column::File])
>     .all(db)
>     .await?;
> ```
> 
> For bigger files, I’d also consider moving the bytes to object storage / filesystem and keeping only metadata + path/key in DB. But if the files are small and DB storage is intentional, explicit projection + a helper method is probably the cleanest workaround for now.
> 
> Maybe you can wrap it like:
> 
> ```rust
> impl EntryWithoutFile {
>     async fn load_file(&self, db: &DatabaseConnection) -> Result<Option<Vec<u8>>, DbErr> {
>         Entity::find_by_id(self.id)
>             .select_only()
>             .column(Column::File)
>             .into_tuple()
>             .one(db)
>             .await
>     }
> }
> ```
> 
> Not as magic as `#[lazy]`, but much clearer and avoids accidentally fetching blobs in list queries.

WookiesRpeople2:
## PR Info

- Closes #3064 

## Features 
- [x] add select except, this will return all fields except the one that is passed as an argument, for example:
```
let bakers = Baker::find()
        .select_except([baker::Column::BakeryId])
        .all(&ctx.db)
        .await
        .unwrap()
```
The generated sql here will be
```
SELECT "baker"."id", "baker"."contact_details", "baker"."bakery_id" FROM "baker"
```
something that is worth noting is that select_exclude will only work on optional value's as the unwrap expects there to be a value present
```
`Result::unwrap()` on an `Err` value: Type("Missing value for column 'name'")
```

## Breaking Changes
- [x] no breaking changes this is purely additions 

## Changes

- [x] added select_except function
- [x] ran make-sync.sh
- [x] added an integration test case

### DIFF
--- a/sea-orm-sync/src/query/select.rs
+++ b/sea-orm-sync/src/query/select.rs
@@ -319,10 +319,7 @@ where
     }
 
     /// Select all columns of this entity except the given ones.
-    pub fn select_except<I>(mut self, except: I) -> Self
-    where
-        I: IntoIterator<Item = E::Column>,
-    {
+    pub fn select_except(mut self, except: impl IntoIterator<Item = E::Column>) -> Self {
         let except: Vec<&str> = except.into_iter().map(|col| col.as_str()).collect();
         self.query.clear_selects();
         for col in E::Column::iter() {
--- a/sea-orm-sync/src/query/select.rs
+++ b/sea-orm-sync/src/query/select.rs
@@ -1,6 +1,6 @@
 use crate::{
-    ColumnTrait, EntityTrait, Iterable, Order, PrimaryKeyToColumn, QueryFilter, QueryOrder,
-    QuerySelect, QueryTrait,
+    ColumnTrait, EntityTrait, IdenStatic, Iterable, Order, PrimaryKeyToColumn, QueryFilter,
+    QueryOrder, QuerySelect, QueryTrait,
 };
 use core::fmt::Debug;
 use core::marker::PhantomData;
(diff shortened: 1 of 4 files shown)


sea-orm-sync/src/query/select.rs
```diff
@@ -317,6 +317,21 @@ where
         }
         self
     }
+
+    /// Select all columns of this entity except the given ones.
+    pub fn select_except<I>(mut self, except: I) -> Self
```
Huliiiiii: Use impl trait

tyt2y3:
Three asks before merge:

1. **Doc that it resets the selection** — it `clear_selects()` + rebuilds, so it overrides any prior `select_only`/`column` and isn't composable with them. Say so in the rustdoc.
2. **Doc the nullable-only limit** — rows hydrate into the full `Model`, so excluding a non-nullable column fails at runtime (`Missing value for column`). Only `Option<_>` columns are safe to exclude.
3. **Add a negative test** covering that failure to pin the contract.

Impl looks good — `col.as_str()` matching is correct since `ColumnTrait` doesn't guarantee `PartialEq`, and the sync mirror + test are in place.

PR #3127, opened 2026-07-07
belongs_to detach: null the FK if nullable, clean error if not

tyt2y3:
Detaching a `belongs_to` relation via the generated `delete_<field>()` builder (`ActiveHasOne::Delete`) was **silently ignored** on save: `take()` maps `Delete` to `None`, so the belongs_to action did nothing and the FK was never touched. So `cake.delete_bakery().save(db)` was a no-op.

This is the deliberately-minimal fix, keeping the **single `ActiveHasOne<E>` type** — nullability drives behavior at runtime, rather than being pushed into the type parameter (which is the direction #3118 explored; closing that in favor of this).

### New Behavior

`delete_<field>()` on a `belongs_to` now:

- **nullable FK** → nulls the column (detaches), keeping both rows.
- **non-nullable FK** → returns a clean `DbErr::Type` ("relation cannot be detached") instead of attempting an UPDATE that the database's NOT NULL constraint would reject with a raw driver error.
- **never-set FK** (a freshly-built ActiveModel) → no-op (inserts NULL), rather than `AttrNotSet`.

`has_one` detach (delete/orphan the child) is unchanged.

### DIFF
--- a/sea-orm-macros/src/derives/active_model_ex.rs
+++ b/sea-orm-macros/src/derives/active_model_ex.rs
@@ -102,7 +102,8 @@ pub fn expand_derive_active_model_ex(
                                 } else if *entity_count.get(entity_path).unwrap() == 1 {
                                     // can only Related to another Entity once
                                     if compound_attrs.belongs_to.is_some() {
-                                        belongs_to_fields.push(ident.clone());
+                                        belongs_to_fields
+                                            .push((ident.clone(), entity_path.to_owned()));
                                     } else if compound_attrs.has_one.is_some() {
                                         has_one_fields.push(ident.clone());
                                     } else if compound_attrs.has_many.is_some()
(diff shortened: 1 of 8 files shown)


Huliiiiii:
I think a runtime error is much worse than an operation that is impossible to express. A foreign key is just a nullable ID column. `Set(Option<M>)` is more consistent.

tyt2y3:
I think we can have `BelongsTo<E>/BelongsTo<Option<E>>` to pair with `belongs_to` for better compile-time guarantee. but it's a trade-off in terms of complexity. `HasOne` should be usable as is, and shouldn't need `HasOne<Option<E>>`.

so I guess we can continue to experiment on #3118 with something like:
```
    #[sea_orm(belongs_to, from = "BakeryId", to = "Id")]
    pub bakery: BelongsTo<Option<super::bakery::Entity>>,
```

PR #3128, opened 2026-07-11
feat: dont require explicit pk for derive entity

romansorin:
allow relations that have no primary key defined to support the `DeriveEntityModel` macro for explicit querying


romansorin:
closing due to incorrect upstream base

PR #3146, opened 2026-07-24
`flatten` and `flatten_ref` methods on `ActiveValue`

### LINKED ISSUE #3145: Flatten methods on ActiveValue
## Motivation

I run into an issue quite frequently where I need to check both if an `ActiveValue` both contains a value and that the value is `Some`. Essentially, I need `Option::flatten`, but for `ActiveValues`. The current way to do this (with referenced `ActiveValues`) is
```rust
model.active_value.try_as_ref().and_then(Option::as_ref)
```
which gets _really annoying_ to keep doing, especially if I need to add another layer by checking if the `Some` value matched a condition:
```rust
model.active_value.try_as_ref().and_then(Option::as_ref).is_some_and(|x| x == "some value")
```
This is always going to get split into multiple lines by rustfmt, which is highly undesirable when all I'm trying to do is check if an ActiveValue's innermost value exists and passes a predicate.

## Proposed Solutions
`ActiveValue::flatten` and `ActiveValue::flatten_ref` methods, which establish a surjective relationship from `Set(Some)`, `Unchanged(Some)` to `Some` and `Set(None)`, `Unchanged(None)`, and `NotSet` to `None`. This can be done by way of sealed trait bounds for the generic `V` in `ActiveValue<V>`, called `AsRefOption<'a, T>` and `AsOwnedOption<T>`. These  would only be implemented for `&'a Option<T>` and `Option<T>` respectively, and would allow our trait bound to work only for `ActiveValues` contining an `Option`.

PacificBird:
Closes #3145

### DIFF
--- a/src/entity/active_value.rs
+++ b/src/entity/active_value.rs
@@ -466,8 +466,8 @@ where
 }
 
 mod private {
-    trait RefOptSealed {}
-    trait OwnedOptSealed {}
+    pub trait RefOptSealed {}
+    pub trait OwnedOptSealed {}
 }
 pub trait AsRefOption<'a, T>: private::RefOptSealed {
     fn as_ref_opt(self) -> &'a Option<T>;
--- a/src/entity/active_value.rs
+++ b/src/entity/active_value.rs
@@ -413,6 +413,74 @@ where
             ActiveValue::NotSet => None,
         }
     }
+    
+    /// Flattens `&ActiveValue<Option<T>>` into an `Option<&T>`, folding [NotSet][ActiveValue::NotSet]
+    /// and [None][Option::None] together, and [Set][ActiveValue::Set], [Unchanged][ActiveValue::Unchanged],
+    /// and [Some][Option::Some] together. This can replace `model.field.try_as_ref().and_then(Option::as_ref)`.
+    ///
+    /// For a version on owned [ActiveValue]s, see [ActiveValue::flatten].
+    ///
+    /// ```
+    /// use sea_orm::ActiveValue;
+    ///
+    /// let x: &ActiveValue<Option<i32>> = &ActiveValue::Set(Some(0));
+    /// let y: &ActiveValue<Option<i32>> = &ActiveValue::Set(None);
+    /// let z: &ActiveValue<Option<i32>> = &ActiveValue::NotSet;
+    ///
+    /// assert_eq!(x.flatten_ref(), Some(&0));
+    /// assert_eq!(y.flatten_ref(), None);
+    /// assert_eq!(z.flatten_ref(), None);
+    /// ```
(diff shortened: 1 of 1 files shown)


Huliiiiii:
Why not just:
```rust
impl<V> ActiveValue<Option<V>>
where
    V: Into<Value> + Nullable, 
{
...
}
```

Huliiiiii:
Naming them `as_option` and `into_option` might be better.

PacificBird:
> Why not just:
> ```rust
> impl<V> ActiveValue<Option<V>>
> where
>     V: Into<Value> + Nullable, 
> {
> ...
> }
> ```

This requires specialization afaik. The `impl<V> ActiveValue<V>` gives a conflicting implementation as far as the compiler is currently concerned.

PacificBird:
> Naming them `as_option` and `into_option` might be better.

That's definitely shorter and nicer.

Huliiiiii:
> > Why not just:
> > ```rust
> > impl<V> ActiveValue<Option<V>>
> > where
> >     V: Into<Value> + Nullable, 
> > {
> > ...
> > }
> > ```
> 
> This requires specialization afaik. The `impl<V> ActiveValue<V>` gives a conflicting implementation as far as the compiler is currently concerned.

Implementing it on `Option<V>` instead of `V` works correctly. (on my machine)
```rust
impl<V> ActiveValue<Option<V>>
where
    V: Into<Value> + Nullable,
{
    pub fn as_option(&self) -> Option<&V> {
        self.try_as_ref().and_then(Option::as_ref)
    }
 
    pub fn into_option(self) -> Option<V> {
        match self {
            Self::Set(value) | Self::Unchanged(value) => value,
            Self::NotSet => None,
        }
    }
}
```

tyt2y3:
Agree with @Huliiiiii's approach — implementing on `ActiveValue<Option<V>>` drops the two sealed helper traits entirely and, as a bonus, sidesteps the compile errors the current trait version has:

- the `AsRefOption<T>` impl is missing its lifetime arg (`AsRefOption<'a, T>`),
- the `flatten_ref` bound constrains the owned `V` instead of `&V`, so it's unsatisfiable for any real caller,
- `flatten(&self)` can't move out of `&self` to return an owned value.

CI is red on all three doctests.

The `as_option` / `into_option` naming is good. One thing the rename gets for free: `into_option` takes `self` by value (matching the `into_` convention), which is exactly the fix `flatten` needed.

tyt2y3:
Superseded by #3155, which landed `as_option` / `into_option` on `ActiveValue<Option<V>>` using @Huliiiiii's concrete-impl approach (no sealed traits) and the `as_option`/`into_option` naming. Thanks @PacificBird for driving this — the use case in #3145 is now covered on master.

PR #3149, opened 2026-07-24
Add `Send` bounds for transactions in `ActiveModelEx` methods

Huliiiiii:
- Closes #3147

PR #3153, opened 2026-07-30
Require `TransactionTrait::Transaction` to be a fixed point (#3147)

### LINKED ISSUE #3147: 2.0 - New Entity Format - Compiler (recursion) error
## Description

I wanted to migrate to the new entity format, but I cannot get the code to compile.

## Steps to Reproduce

1. Create Entities with the new entity format

E.g. :
```

#[sea_orm::model]
#[derive(Debug, Clone, PartialEq, Eq, DeriveEntityModel)]
#[sea_orm(table_name = "table")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: u32,
    pub one: Option<String>,
    pub two: Option<String>,
    pub three: bool,
    pub four: bool,
    pub rel_one_id: Option<u32>,
    // #[sea_orm(belongs_to, from = "rel_one_id", to = "id")]
    // pub rel_one: BelongsTo<Option<super::rel_ones::Entity>>,
    pub rel_two_id: Option<u32>,
    // #[sea_orm(belongs_to, from = "rel_two_id", to = "id")]
    // pub rel_two: BelongsTo<Option<super::rel_twos::Entity>>,
}

```

### Expected Behavior

Code should compile

### Actual Behavior

I get a compiler error:

```
error: future cannot be sent between threads safely
 --> entity/src/[redacted].rs:3:1
  |
3 | #[sea_orm::model]
  | ^^^^^^^^^^^^^^^^^ future returned by `cascade_delete` is not `Send`
  |
note: captured value is not `Send` because `&` references cannot be sent unless their referent is `Sync`
 --> entity/src/[redacted].rs:3:1
  |
3 | #[sea_orm::model]
  | ^^^^^^^^^^^^^^^^^ has type `&C` which is not `Send`, because `C` is not `Sync`
  = note: `C` doesn't implement `std::marker::Sync`
note: future is not `Send` as this value is used across an await
 --> entity/src/[redacted].rs:3:1
  |
3 | #[sea_orm::model]
  | ^^^^^^^^^^^^^^^^^
  | |
  | await occurs here, with `#[sea_orm::model]` maybe used later
  | has type `<C as sea_orm::TransactionTrait>::Transaction` which is not `Send`
  = note: `<C as sea_orm::TransactionTrait>::Transaction` doesn't implement `std::marker::Send`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#future_not_send
```

This error occurs also for other internal methods like `delete`, `action`, `save`, ...

If I add the relations I get even more errors:

```
error[E0275]: overflow evaluating the requirement `dyn std::future::Future<Output = std::result::Result<<<<<<... as sea_orm::TransactionTrait>::Transaction as sea_orm::TransactionTrait>::Transaction as sea_orm::TransactionTrait>::Transaction as sea_orm::TransactionTrait>::Transaction as sea_orm::TransactionTrait>::Transaction, ...>> + std::marker::Send: std::marker::Send`
    --> entity/src/[redacted].rs:4:1
     |
   4 | #[sea_orm::model]
     | ^^^^^^^^^^^^^^^^^
     |
     = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`entity`)
     = note: required for `Unique<dyn Future<Output = Result<<<<... as TransactionTrait>::Transaction as TransactionTrait>::Transaction as TransactionTrait>::Transaction, ...>> + Send>` to implement `std::marker::Send`
note: required because it appears within the type `Box<dyn Future<Output = Result<<<<... as TransactionTrait>::Transaction as TransactionTrait>::Transaction as TransactionTrait>::Transaction, ...>> + Send>`
```

### Reproduces How Often

Always

### Workarounds

Tried without success:
- Increase recursion limit
- Delete cache (target folder)

## Versions

sea-orm 2.0 developing on macOS

```
[workspace.dependencies.sea-orm]
version = "2.0.0"
default-features = false
features = [
    "macros",
    "with-chrono",
    "with-json",
    "with-uuid",
    "sqlx-sqlite",
    "sqlite-use-returning-for-3_35",
    "runtime-tokio-rustls"
]
```

tyt2y3:
Closes #3147. Alternative to #3149 by @Huliiiiii

### Problem

The methods generated by `#[sea_orm::model]` open a transaction and save related models inside it, so `action::<C>` recurses into `action::<C::Transaction>`. Proving those futures `Send` without a concrete `C` walks the unbounded `C::Transaction::Transaction::…` chain and overflows (E0275 / `clippy::future_not_send`) — the documented 2.0 entity format fails to compile under clippy.

### Fix

On the trait, not the generated methods. `TransactionTrait` now requires `Sync` and pins its associated `Transaction` to a `Send` fixed point (`Transaction::Transaction == Transaction`), collapsing the chain to one type so the obligation terminates.

### Why on the trait

- `Sync` / `Send` exclude no implementor — the `#[async_trait]` desugaring already forces both (`begin` holds `&self` across an await; `TransactionSession::commit` takes `self` by value). Only a non-fixed-point `Transaction` is newly rejected, and it errors at the offending `impl`, not at `#[sea_orm::model]`. All in-crate implementors are fixed points already.
- Putting the bounds on the generated methods (#3149) leaks into user call sites: `fn create<C: ConnectionTrait + TransactionTrait>(db: &C)` compiles on 2.0.0 and stops compiling, with no exported alias to restate. On the trait it is invisible to callers.


Huliiiiii:
Is it possible that TransactionTrait::Transition is different from itself?

tyt2y3:
I worried about this too. claude research says:

> To be broken by the fixed-point requirement you need a hand-written impl TransactionTrait whose Transaction type is itself not a fixed point — two custom layers deep. But to implement the trait at all you must produce a Transaction: ConnectionTrait + TransactionTrait + TransactionSession, and the only realistic way to get one is to delegate to DatabaseTransaction

meaning, unlikely anyone would have a custom impl that is not backed by a transaction manager

Huliiiiii:
> I worried about this too. claude research says:
> 
> > To be broken by the fixed-point requirement you need a hand-written impl TransactionTrait whose Transaction type is itself not a fixed point — two custom layers deep. But to implement the trait at all you must produce a Transaction: ConnectionTrait + TransactionTrait + TransactionSession, and the only realistic way to get one is to delegate to DatabaseTransaction
> 
> meaning, unlikely anyone would have a custom impl that is not backed by a transaction manager

Sounds reasonable to me.

sea-orm-sync/src/database/connection.rs
```diff
@@ -168,9 +168,18 @@ pub struct TransactionOptions {
 /// transactions via SAVEPOINTs). Use [`begin`](Self::begin) for a manually
 /// managed transaction, or [`transaction`](Self::transaction) for a closure
 /// that auto-commits on `Ok` and rolls back on `Err`.
+///
+/// [`Self::Transaction`] must be a fixed point: beginning a transaction on a
```
Huliiiiii: I think we can simplify the comment to explain that this is a requirement of methods on ActiveModelEx.

tyt2y3: Done — reworded to point at the `ActiveModelEx` methods and their recursion through `Self::Transaction`.

PR #3156, opened 2026-08-04
feat: Add built-in PostgreSQL geometric types support (#282)

### LINKED ISSUE #282: Add postbus types support and allow custom types
Hi team, 
I want to use postbus types, but I'm not sure how to do it.

I can use genotypes crate but it's not allowed in entity, 

Can you please support geotypes and also allow to add custom types?

kAif-gh:
# Add support for built-in PostgreSQL geometric types
Closes #282
## What
Adds support for PostgreSQL's built-in geometric types in entities:
`point`, `line`, `lseg`, `box`, `path`, `polygon`, `circle`.
These are the core Postgres geometric types (Postgres manual §8.8) — **not**
PostGIS, so no extension is required.
## How
sqlx already provides structs for these types (`PgPoint`, `PgPolygon`, …), but
Rust's orphan rules stop us from implementing sea-orm's traits on them directly.
So this adds a thin `Geo<T>` newtype wrapper that implements the needed traits
(`ValueType`, `TryGetable`, `Nullable`, `IntoActiveValue`, …) and derefs to the
inner sqlx type.
Values cross the DB boundary as canonical Postgres text using the existing
`save_as` / `select_as` casts, so the core `Value` enum is left unchanged.
Everything is gated behind a new `postgres-geometry` feature.
## Usage
```rust
#[sea_orm(table_name = "geo")]
pub struct Model {
    pub id: i32,
    #[sea_orm(column_type = r#"custom("point")"#, select_as = "text", save_as = "point")]
    pub location: Geo<PgPoint>,
    #[sea_orm(column_type = r#"custom("polygon")"#, select_as = "text", save_as = "polygon")]
    pub boundary: Geo<PgPolygon>,
}
```
## Tests
Adds `tests/postgres_geometry_tests.rs` covering insert/select round-trips of all
seven types (including a NULL case) and native spatial queries (`@>`, `<->`).
```bash
cargo test --features sqlx-postgres,postgres-geometry --test postgres_geometry_tests
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### New Features
 
+* Support built-in PostgreSQL geometric types (`point`, `line`, `lseg`, `box`, `path`, `polygon`, `circle`) https://github.com/SeaQL/sea-orm/issues/282
+
+  Behind the new `postgres-geometry` feature, geometric columns can be mapped
+  onto entity fields via a thin `Geo<T>` newtype wrapping sqlx's `Pg*` types
+  (`PgPoint`, `PgPolygon`, …). These are the core Postgres geometric types — no
+  PostGIS extension required. Values are stored/loaded as canonical Postgres
+  text using `save_as` / `select_as` casts, so the core `Value` enum is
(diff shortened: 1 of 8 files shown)


Huliiiiii:

We should not directly depend on sqlx types. Newtype wrapper is not a good idea.

PR #3159, opened 2026-08-05
Add a datetime-default-now helper function

goodspark:
I found I was writing this a lot in my migrations. So why not just put it in the library?

## New Features

- Adds a `date_time_default_now` helper function for migrations.

### DIFF
--- a/sea-orm-migration/src/schema.rs
+++ b/sea-orm-migration/src/schema.rs
@@ -293,6 +293,11 @@ pub fn date_time_uniq<T: IntoIden>(col: T) -> ColumnDef {
     date_time(col).unique_key().take()
 }
 
+/// A date time column with a server-side default of 'now'.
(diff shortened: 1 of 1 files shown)


sea-orm-migration/src/schema.rs
```diff
@@ -293,6 +293,11 @@ pub fn date_time_uniq<T: IntoIden>(col: T) -> ColumnDef {
     date_time(col).unique_key().take()
 }
 
+/// A date time column with a server-side default of 'now'.
+pub fn date_time_default_now<T: IntoIden>(col: T) -> ColumnDef {
```
Huliiiiii, line 297: now() != CURRENT_TIMESTAMP

Furthermore, CURRENT_TIMESTAMP has different semantics in different databases.

goodspark: Ah sorry, late to this. What do you suggest? Just changing the comment? Can do in a follow-up PR.

PR #3164, opened 2026-08-08
Add `require_one`: fetch exactly one row, error if none (#2974)

### LINKED ISSUE #2974: Query one non-optional
## Motivation

I frequently query one row of something using `.one(db)` method and return an error if its `None`.

It would be nice to have a distinction between fetching one row and fetching optional row since its quite common(like in sqlx).

## Proposed Solutions

I propose to add another method to various query implementations, to query non-optional and return
https://github.com/SeaQL/sea-orm/blob/4f89d1bc16deb4bcba47d776db9393003003de34/src/error.rs#L51-L53
this error on None.

## Additional Information

 I think this feature would be relevant after error refactor proposed in one of the recent issues.
Naming this method seems a big problem as `.one(db)` is already taken for optional variant.
I can implement myself if needed.

tyt2y3:
Closes #2974.

## Motivation

Fetching a single row that is *expected* to exist is extremely common, and today it requires unwrapping the `Option` that `one()` returns:

```rust
let user = User::find_by_id(id).one(db).await?.ok_or(MyErr::NotFound)?;
```

Other query APIs distinguish "fetch optional" from "fetch one" (e.g. sqlx `fetch_optional` vs `fetch_one`, Diesel `.optional()` vs `.get_result()`). SeaORM only had the optional form.

## What this adds

`require_one()` — the non-optional counterpart to `one()`. It returns the item directly and yields `DbErr::RecordNotFound` when no row matches, so the call site can just use `?`:

```rust
let user = User::find_by_id(id).require_one(db).await?;
```

Added on the base `Selector` and `SelectorRaw` executors, with ergonomic wrappers on `Select`, `SelectTwo`, and `SelectTwoRequired` (the types that already expose `one()`). All route through a single `RecordNotFound` construction.

## Notes

- Purely additive; no breaking changes.
- Mirrored to `sea-orm-sync` via `build-tools/make-sync.sh`.
- Covered by a runnable mock doc-test on `Select::require_one` exercising both the found and not-found paths (passes in async and sync crates).
- Name chosen as `require_one` (reads as "this query requires a row"; no panic connotation, unlike `expect_one`).

No changelog entry — leaving next-version slotting to the release workflow.

### DIFF
--- a/sea-orm-sync/src/executor/select.rs
+++ b/sea-orm-sync/src/executor/select.rs
@@ -28,6 +28,12 @@ type PinBoxStream<'b, S> = std::pin::Pin<Box<dyn Stream<Item = Result<S, DbErr>>
 #[cfg(feature = "sync")]
 type PinBoxStream<'b, S> = Box<dyn Iterator<Item = Result<S, DbErr>> + 'b>;
 
+/// The error returned by the `require_one` family when a query that must match a
+/// row matches none.
+fn record_not_found() -> DbErr {
+    DbErr::RecordNotFound("None of the models match the query".to_owned())
+}
+
 /// A ready-to-execute `SELECT` query backed by a [`SelectStatement`]. The
 /// type parameter `S` (a [`SelectorTrait`]) determines what each row is
 /// decoded into. Build one via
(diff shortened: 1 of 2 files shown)


svenstaro:
Nice. Are you also planning to add this to the EntityLoader?

PR #3169, opened 2026-08-12
Refactor macro attribute parsing: migrate from bae to darling and remove string/Ident round-trips

### LINKED ISSUE #3119: Refactor macro implmentions

TheShiveshNetwork:
- Closes #3119

- Dependencies:
  - The `bae` -> `darling` migration is part of this branch: `bae` (`sea-bae`) is fully removed and all five attribute structs (`derive_attr`, `relation_attr`, `compound_attr`, `value_type_attr`, `related_attr`) now use `darling::FromMeta`. `darling` is pinned to `0.23` (latest release still on `syn 2.0`; `0.24` would require a separate `syn` 2->3 migration). The original `try_from_attributes` / `from_attributes` signatures are preserved so call sites are unchanged.
  - Went back and forth on darling vs just hand-parsing with `parse_nested_meta` (a few derives here already do that, e.g. `column.rs`/`derive_iden.rs`, so it's not like it'd be a foreign pattern). Ended up going with darling because `compound_attr` has 14 fields and actual co-occurrence rules between them (`from`+`to` need each other, `has_one`/`belongs_to` are mutually exclusive, etc.), and writing that validation by hand for all 5 structs felt like we'd just be badly reinventing what `FromMeta` already does. Also figured darling being a known quantity in the ecosystem is nicer for anyone picking this crate up later than a parser only sea-orm uses.

- Dependents:
  - Working on syn migration to `3.0.3`, yet to create the pr

**New Features**
- None

**Bug Fixes**
- None (refactor only)

**Breaking Changes**
- None. Attribute syntax (`#[sea_orm(...)]`) is untouched, this is purely how the macros parse it internally.

**Changes**
- [x] **Migrate macro attribute parsing from `bae` to `darling`.** Removed the `bae`/`sea-bae` dependency; the five attribute structs in `sea-orm-macros/src/derives/attributes.rs` now derive `darling::FromMeta`, with validation deferred so unknown attributes still warn rather than error.
- [x] Replace manual `Ident::new(&format!(..))` with `format_ident!` for literal and transformed idents across the derive macros (`model_ex`, `active_model_ex`, `entity_model`, `typed_column`, `value_type_match`, `related_entity`, `entity_loader`, `partial_model`, `into_active_model`, `raw_sql`).
- [x] Build derived types with `syn::parse_quote!` instead of `syn::parse_str` on a formatted string (e.g. the `#[sea_orm(compact)]` related `ActiveModel` path in `partial_model`).
- [x] Compare `Ident`s directly via `*id == "some_str"` instead of stringifying with `.to_string()` first, mainly in `util.rs`'s `CompoundType::matches_type` / `CompoundType::from_type` (worst offender, 5 stringify-then-match sites in one function), plus a few smaller ones in `active_enum.rs`, `column.rs`, and `derive_iden.rs`.


sea-orm-macros/src/derives/active_model_ex.rs
```diff
@@ -192,7 +192,7 @@ struct ScalarField<'a> {
 impl<'a> ScalarField<'a> {
     fn from_field(field: &'a syn::Field, ident: &Ident) -> syn::Result<Self> {
         let column = trim_starting_raw_identifier(ident).to_upper_camel_case();
-        let mut column = Ident::new(&escape_rust_keyword(column), ident.span());
+        let mut column = format_ident!("{}", escape_rust_keyword(column));
```
Huliiiiii: This will cause the span lost.

sea-orm-macros/src/derives/attributes.rs
```diff
@@ -1,92 +1,263 @@
...
+            let is_bare = matches!(meta, NestedMeta::Meta(Meta::Path(_)));
+            if is_bare && !flags.contains(&name.as_str()) {
+                return false;
+            }
+            true
+        })
+        .collect()
+}
+
+fn darling_err(e: darling::Error) -> syn::Error {
```
Huliiiiii: Doesn't Darling already have a method for this?

sea-orm-macros/src/derives/attributes.rs
```diff
@@ -1,92 +1,263 @@
+use darling::FromMeta;
+use darling::ast::NestedMeta;
+use syn::punctuated::Punctuated;
+use syn::{Attribute, Meta, MetaList, MetaNameValue, Token};
+
+fn parse_sea_orm_attrs(attrs: &[Attribute]) -> syn::Result<Vec<NestedMeta>> {
```
Huliiiiii: These helpers seem not right to me. Darling should have a way to do this.

sea-orm-macros/src/derives/attributes.rs
```diff
@@ -1,92 +1,263 @@
...
+        "has_many",
+        "belongs_to",
+        "self_ref",
+        "skip_fk",
+        "reverse",
+    ];
 
     /// Attributes for compound model fields
-    #[derive(Default, FromAttributes)]
+    #[derive(Default, FromMeta)]
```
Huliiiiii: This doesn't solve the issue. You should replace the current manually parsed compatibility layer with darling.

sea-orm-macros/src/raw_sql.rs
```diff
@@ -28,11 +28,11 @@ pub fn expand(input: proc_macro::TokenStream) -> syn::Result<TokenStream> {
         ..
     } = syn::parse(input)?;
 
-    let builder = match backend.to_string().as_str() {
-        "MySql" => quote!(MysqlQueryBuilder),
-        "Postgres" => quote!(PostgresQueryBuilder),
-        "Sqlite" => quote!(SqliteQueryBuilder),
-        backend => panic!("Unsupported backend {backend}"),
+    let builder = match &backend {
```
Huliiiiii, line 31: Does deref ident still allocate on the heap?

TheShiveshNetwork: No allocation here. this compares the Ident against string literals via Ident: PartialEq<str>, which reads the ident's name as a &str and byte-compares

TheShiveshNetwork:
@Huliiiiii this pr should be ready for review. moreover, i've also opened the pr for migration of syn 2 to syn 3 #3171 we can complete that first and then, i'll migrate darling to latest once that is merged.

Huliiiiii:
```
pub fn try_from_attributes(attrs: &[Attribute]) -> syn::Result<Option<Self>> {
    super::try_from_attributes(attrs)
}
```

Please, I don't want to see this kind of LLM slop. You also haven't understood what we're trying to fix. Please put more thought into the code. If you're going to rely on LLM, it'd be much easier for me to just do it myself.

PR #3177, opened 2026-08-20
fix: handle missing time columns in ActiveModel::from_json

### LINKED ISSUE #3175: `ActiveModel::from_json` fails to deserialize TimeDateTimeWithTimezone fields to NotSet when field is missing from JSON payload
## Description
This is a continuation of #3160, which was closed due to thinking that things were closed by [SeaQL/sea-query#1093](https://github.com/SeaQL/sea-query/pull/1093). This sea-query fixed the issues happening for optional values, but didn't remedy the deserialization problems for non-optional ones.

When deserializing an ActiveValue using `from_json`, if a `TimeDateTimeWithTimezone` field doesn't have a corresponding value in the JSON payload, somewhere along the way it defaults to a stringified version of the unix epoch, which causes the following error down the line:

```
Json Error: invalid type: string "'1970-01-01 00:00:00.000000 +00:00'", expected an `OffsetDateTime`
```

## Steps to Reproduce
This is a minimal example of the exact problem I'm having:
```rust
        #[sea_orm::model]
        #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, DeriveEntityModel)]
        #[sea_orm(table_name = "example")]
        pub struct Model {
            #[serde(rename = "STATION_ID")]
            pub unit_id: i32,
        }

        let payload = r#"{ "data": [
            {"STATION_ID": 100000}
        ]}"#;

        let result = serde_json::from_str::<serde_json::Value>(&payload)
            .unwrap()
            .get("data")
            .and_then(serde_json::Value::as_array)
            .unwrap()
            .into_iter()
            .filter_map(|val| {
                let x = <ActiveModel>::from_json(val.clone());
                dbg!(&x); // Json Error: invalid type: string "'1970-01-01 00:00:00.000000 +00:00'", expected an `OffsetDateTime`
                x.ok()
            })
            .collect::<Vec<_>>();

        assert!(!result.is_empty());
```

### Expected Behavior

Any missing fields from the JSON payload should just deserialize as `NotSet` in the ActiveModel, which they did in 2.0.0-rc32.

aniketch07:
`ActiveModel::from_json` fails when a JSON payload is missing a time-typed
column (e.g. `TimeDateTimeWithTimeZone`, `TimeDateTime`, chrono equivalents).
The trait-default implementation serializes dummy values with
`sea_value_to_json_value`, which produces SQL-literal strings (`'1970-01-01
00:00:00.000000 +00:00'`) that the model's serde deserializer cannot parse
back into a typed value.

This moves `from_json` into the derive macro, as suggested in #3175. The
generated implementation serializes the dummy model with serde, looks each
column's value up by its serialize name (honouring `serde(rename)` /
`serde(rename_all)`), and re-keys it to the column's json key, so the model
round-trip always succeeds and missing columns become `NotSet`.

Includes a regression test covering both a missing and a present
`TimeDateTimeWithTimeZone` column.

- fixes #3175

### DIFF
--- a/sea-orm-macros/src/derives/active_model.rs
+++ b/sea-orm-macros/src/derives/active_model.rs
@@ -1,8 +1,12 @@
-use super::util::{escape_rust_keyword, field_not_ignored, trim_starting_raw_identifier};
+use super::case_style::CaseStyle;
+use super::util::{
+    consume_meta, escape_rust_keyword, field_not_ignored, trim_starting_raw_identifier,
+};
 use heck::ToUpperCamelCase;
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
-use syn::{Data, DataStruct, Expr, Fields, LitStr, Type, Visibility};
+use std::str::FromStr;
+use syn::{Attribute, Data, DataStruct, Expr, Fields, LitStr, Type, Visibility};
 
 pub(crate) struct DeriveActiveModel {
     model: Ident,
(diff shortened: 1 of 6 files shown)


Huliiiiii:

slop

PR #3184, opened 2026-08-24
Most elegant way I could figure to fix from_json

### LINKED ISSUE #3175: `ActiveModel::from_json` fails to deserialize TimeDateTimeWithTimezone fields to NotSet when field is missing from JSON payload

PacificBird:
## PR Info
Closes #3175

It's honestly not the prettiest fix in the world, but as I mention in the comments of the original issue this is a very difficult problem to solve cleanly. Modified-deserialization-by-proxy is very much not an intended or supported use case of serde. I tried a fully macro-based implementation of `from_json`, but ran into big problems of the duplication of work between the derives for Models and ActiveModels. In any case I do believe that this is the least-bad way to solve this general design problem short of a very clever workaround and a big refactor, so I'm happy with 

I took the idea of making the dummy map by doing a round trip from the dummy active model, to model, into a JSON map using `serde_json::to_values`, and then merging with `serde_json::Map::append` with the input JSON values from the AI slop pull, although I hand-coded all of the active model and macro logic without the use of any AI (I straight up don't have a subscription to any and my laptop is not powerful enough to meaningful run a local model...). I did initially copy and then heavily modify the regression test from that PR to test both `time` and `chrono`. I left the original comment explaining the test though, I hope that's okay.

The addition of `ColumnTrait::serialize_json_key` was a necessary evil, as I needed a way to map the serialized keys from roundtripping the dummy values into their deserialize keys so that the final conversion of the merged map into a model (and ultimately an active model) would actually see the dummy values. The way I did it is pretty simple, I just made a `HashMap` by iterating over `Column::iter` keyed by the serialize keys from `json_key_serialize` to the deserialize keys from `json_key`. That's then used to remap the keys from the dummy values.

This pull may not be my cleanest Rust work, sorry about that. Running rust-analyzer on this codebase eats all of my RAM and soft-locks Linux for me...

## Checklist

- [x] I have read the [AI Policy](https://github.com/SeaQL/.github/blob/master/AI_POLICY.md) and confirm that this contribution complies with it.

## Release Notes

- Fix incorrect deserialization of missing DateTime (both `time` and `chrono` fields) for `ActiveModel::from_json`
- Add method: `ColumnTrait::serialize_json_key` as well as macro implementations for it in `DeriveEntityTrait`.

Because the new trait method has a default implementation, the addition does not constitute a breaking version change!

### DIFF
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -508,36 +508,34 @@ pub trait ActiveModelTrait: Clone + Debug {
             )));
         };
 
-        let dummy_am = Self::default_values();
-        let dummy_model = dummy_am.clone().try_into_model().map_err(json_err)?;
+        let dummy_model = Self::default_values().try_into_model().map_err(json_err)?;
         let dummy_value = serde_json::to_value(dummy_model).map_err(json_err)?;
-
-        let len = <<Self::Entity as EntityTrait>::Column>::iter().len();
-        // Mark down which attribute exists in the JSON object
-        let mut json_keys = Vec::with_capacity(len);
-        let serde_json::Value::Object(merged) = dummy_value else {
+        let serde_json::Value::Object(dummy_value) = dummy_value else {
             return Err(DbErr::Json(format!(
                 "invalid type: expected JSON object for dummy model for {}",
                 <<Self as ActiveModelTrait>::Entity as IdenStatic>::as_str(&Default::default())
             )));
         };
 
-        let mut ser_de_map = <<Self::Entity as EntityTrait>::Column>::iter()
-            .map(|col| (col.json_key_serialize(), col.json_key()))
+        let ser_de_map = <<Self::Entity as EntityTrait>::Column>::iter()
+            .map(|col| (col.serialize_json_key(), col.json_key()))
             .collect::<std::collections::HashMap<_, _>>();
 
-        let mut merged = merged
+        let mut merged = dummy_value
             .into_iter()
             .map(|(key, val)| {
                 // map seralized keys into deserialize keys
                 let new_key = ser_de_map
-                    .remove(key.as_str())
+                    .get(key.as_str())
                     .map(ToString::to_string)
                     .unwrap_or(key);
                 (new_key, val)
             })
             .collect::<serde_json::Map<_, _>>();
 
+        let len = <<Self::Entity as EntityTrait>::Column>::iter().len();
+        // Mark down which attribute exists in the JSON object
+        let mut json_keys = Vec::with_capacity(len);
         for col in <<Self::Entity as EntityTrait>::Column>::iter() {
             let key = col.json_key();
             let has_key = input.contains_key(key);
--- a/src/entity/active_model.rs
+++ b/src/entity/active_model.rs
@@ -509,21 +509,36 @@ pub trait ActiveModelTrait: Clone + Debug {
         };
 
         let dummy_am = Self::default_values();
+        let dummy_model = dummy_am.clone().try_into_model().map_err(json_err)?;
+        let dummy_value = serde_json::to_value(dummy_model).map_err(json_err)?;
+
         let len = <<Self::Entity as EntityTrait>::Column>::iter().len();
         // Mark down which attribute exists in the JSON object
         let mut json_keys = Vec::with_capacity(len);
-        let mut merged = serde_json::Map::with_capacity(len);
+        let serde_json::Value::Object(merged) = dummy_value else {
+            return Err(DbErr::Json(format!(
+                "invalid type: expected JSON object for dummy model for {}",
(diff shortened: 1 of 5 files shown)


src/entity/column.rs
Huliiiiii: Should we update the comment here as well?

src/entity/column.rs
```diff
@@ -537,6 +537,12 @@ pub trait ColumnTrait: IdenStatic + Iterable + FromStr {
     fn json_key(&self) -> &'static str {
         self.as_str()
     }
+
+    /// JSON key used for this column when serializing the model.
+    #[cfg(feature = "with-json")]
+    fn json_key_serialize(&self) -> &'static str {
```
Huliiiiii: serialization_json_key?

src/entity/active_model.rs
```diff
@@ -509,21 +509,39 @@ pub trait ActiveModelTrait: Clone + Debug {
         };
 
         let dummy_am = Self::default_values();
+        let dummy_model = dummy_am.clone().try_into_model().map_err(json_err)?;
```
Huliiiiii: Remove this clone or inline dummy_am

src/entity/active_model.rs
```diff
@@ -509,21 +509,39 @@ pub trait ActiveModelTrait: Clone + Debug {
...
+        let mut ser_de_map = <<Self::Entity as EntityTrait>::Column>::iter()
+            .map(|col| (col.json_key_serialize(), col.json_key()))
+            .collect::<std::collections::HashMap<_, _>>();
+
+        let mut merged = merged
+            .into_iter()
+            .map(|(key, val)| {
+                // map seralized keys into deserialize keys
+                let new_key = ser_de_map
+                    .remove(key.as_str())
```
Huliiiiii: Why use remove instead of get?

src/entity/active_model.rs
```diff
@@ -509,21 +509,39 @@ pub trait ActiveModelTrait: Clone + Debug {
...
 
         let dummy_am = Self::default_values();
+        let dummy_model = dummy_am.clone().try_into_model().map_err(json_err)?;
+        let dummy_value = serde_json::to_value(dummy_model).map_err(json_err)?;
+
         let len = <<Self::Entity as EntityTrait>::Column>::iter().len();
         // Mark down which attribute exists in the JSON object
         let mut json_keys = Vec::with_capacity(len);
-        let mut merged = serde_json::Map::with_capacity(len);
+        let serde_json::Value::Object(merged) = dummy_value else {
```
Huliiiiii: We can merge this into the definition of dummy_value.

src/entity/active_model.rs
PacificBird: The original rationale was to avoid having to do a `get+clone`, but since I have to do `to_string` anyways you're right that it doesn't end up saving an extra allocation anyways.

PacificBird:
Added changes for these review comments and also ensured that for `#[serde(rename/-all = "...")]` (i.e., symmetric ser/de names), it will apply to both serialize and deserialize json keys.

PacificBird:
not sure if maintainers can rerun tests, the failing test is some sort of CI error, not a failing test

PR #3191, opened 2026-09-02
fix: pass through Eq derive in ModelEx to fix clippy::derive_partial_eq_without_eq (issue #3172)

### LINKED ISSUE #3172: `Eq` skipped in `ModelEx`
### Discussed in https://github.com/SeaQL/sea-orm/discussions/3170

<div type='discussions-op-text'>

<sup>Originally posted by **kapalla** August 13, 2026</sup>
A simple model like this

```rust
#[sea_orm::model]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    pub name: String,
}
```

causes a clippy error with the `-D clippy::derive-partial-eq-without-eq` flag or the `clippy::nursery` group.

https://github.com/SeaQL/sea-orm/blob/13be0db50c85f633c810d920cffa2d5d79233a6a/sea-orm-macros/src/derives/model_ex.rs#L100-L102</div>

yunaremaia:
The model_ex macro was silently stripping the Eq derive from the generated ModelEx struct, which triggers clippy::derive_partial_eq_without_eq when -D warnings or clippy::nursery is enabled.

Eq is a trivial marker trait that is always valid when PartialEq is present, so passing it through instead of skipping it is safe and correct.

Fixes #3172.

Assisted-by: poolside/laguna-s-2.1

### DIFF
--- a/sea-orm-macros/src/derives/model_ex.rs
+++ b/sea-orm-macros/src/derives/model_ex.rs
@@ -98,7 +98,8 @@ pub fn expand_sea_orm_model(input: ItemStruct, compact: bool) -> syn::Result<Tok
 
         list.parse_nested_meta(|meta| {
             if meta.path.is_ident("Eq") {
-                // skip
+                // Pass through Eq so clippy's `derive_partial_eq_without_eq` doesn't fire
+                new_list.push(parse_quote!(Eq));
             } else if meta.path.is_ident("DeriveEntityModel") {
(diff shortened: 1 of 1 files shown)


Huliiiiii:
Slop.

> Eq is a trivial marker trait that is always valid when PartialEq is present, so passing it through instead of skipping it is safe and correct.

Have you heard of floats?

PR #3194, opened 2026-09-04
Fix select_as aliases in combined selects

### LINKED ISSUE #3174: find_also_related panics on entities generated by sea-orm-cli for citext columns
## Description

`sea-orm-cli generate entity` emits `select_as = "text"` for columns of a custom
Postgres type such as `citext`. Any entity carrying that attribute panics when
used with `find_also_related` / `select_also`:
 
```
thread panicked at sea-orm-2.0.1/src/query/combine.rs:72:30:
cannot apply alias for expr other than Column or AsEnum
```
 
The panic happens while building the query, before any SQL is sent to the
database.
 
This makes two shipped features mutually exclusive: entities produced by the
official generator cannot be used with typed joins. Since `citext` is a common
choice for emails, usernames and slugs, this is reached with default tooling and
no unusual configuration.
 
This was previously reported in #1749 (converted to discussion #1750) on 0.11.3
in July 2023 and has not been addressed. Reproduced here on 2.0.1 with a
narrower diagnosis: `select_as` alone is responsible, and removing it fixes the
join without breaking reads.

## Steps to Reproduce

Schema:
 
```sql
CREATE EXTENSION IF NOT EXISTS citext;
 
CREATE TABLE users (
  id       UUID PRIMARY KEY,
  username CITEXT NOT NULL UNIQUE
);
 
CREATE TABLE user_emails (
  id      UUID PRIMARY KEY,
  user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
  email   CITEXT NOT NULL UNIQUE
);
```
 
Entity as generated by `sea-orm-cli generate entity`:
 
```rust
#[sea_orm(table_name = "user_emails")]
pub struct Model {
    pub id: Uuid,
    pub user_id: Uuid,
    #[sea_orm(
        column_type = r#"custom("citext")"#,
        select_as = "text",
        save_as = "citext",
        unique
    )]
    pub email: String,
}
```
 
Query:
 
```rust
user_emails::Entity::find()
    .filter(user_emails::Column::Email.eq("someone@example.com"))
    .find_also_related(users::Entity)
    .one(db)
    .await
```

### Expected Behavior

The query returns `Option<(user_emails::Model, Option<users::Model>)>`.

### Actual Behavior

Panic in `Select::apply_alias`:
 
```
thread panicked at sea-orm-2.0.1/src/query/combine.rs:72:30:
cannot apply alias for expr other than Column or AsEnum

   2: <sea_orm::query::select::Select<user_emails::Entity>>::apply_alias::{closure#0}
        at sea-orm-2.0.1/src/query/combine.rs:72:30
   5: <sea_orm::query::select::Select<user_emails::Entity>>::apply_alias
        at sea-orm-2.0.1/src/query/combine.rs:47:22
   6: <sea_orm::query::select::Select<user_emails::Entity>>::select_also::<users::Entity>
        at sea-orm-2.0.1/src/query/combine.rs:87:21
   7: <sea_orm::query::select::Select<user_emails::Entity>>::find_also_related::<users::Entity>
        at sea-orm-2.0.1/src/query/join.rs:63:27
```
`select_as` turns the selected column into a cast expression rather than a plain
column, and `apply_alias` only handles `Expr::Column` and `AsEnum`.

### Reproduces How Often

Always. Any entity with `select_as` on a selected column panics on every
`find_also_related` / `select_also` call, before any query reaches the database.

### Workarounds

Removing `select_as` from the generated entity fixes the join. Reads still
decode correctly: `citext` comes back as `text` over the wire without an
explicit cast, and `save_as = "citext"` is enough for writes.
 
Since the attribute is emitted by the generator, working around this means
post-processing every generated entity file on each regeneration.

## Versions
```
│   ├── sea-orm v2.0.1
│   │   ├── sea-orm-macros v2.0.1 (proc-macro)
│   │   │   ├── sea-bae v0.2.2 (proc-macro)
│   │   ├── sea-query v1.0.1
│   │   │   ├── sea-query-derive v1.0.0 (proc-macro)
│   │   ├── sea-query-sqlx v0.9.1
│   │   │   ├── sea-query v1.0.1 (*)
│   │   ├── sea-schema v0.18.1
│   │   │   ├── sea-query v1.0.1 (*)
│   │   │   ├── sea-query-sqlx v0.9.1 (*)
│   │   │   ├── sea-schema-derive v0.3.0 (proc-macro)
│   ├── sea-orm-migration v2.0.1
│   │   ├── sea-orm v2.0.1 (*)
│   │   ├── sea-orm-cli v2.0.1
│   │   ├── sea-schema v0.18.1 (*)
├── sea-orm v2.0.1 (*)
│   ├── sea-orm v2.0.1 (*)
```
sea-orm-cli 2.0.1
PostgreSQL: 18.4
OS: Fedora 44
Rust: 1.97.1

boopy253:
Fixes #3174.

## Root cause

A column-level select_as transform changes the projection from a column expression into a cast expression. Combined selects prefix the first-entity projection aliases, but can only infer an unaliased projection name from a plain column or an enum cast. As a result, find_also_related and the other combined-select paths panic before executing SQL.

## Fix

Preserve the column logical name as the projection alias whenever select_as produces a transformed expression. Plain columns and the existing enum path retain their previous SQL shape. Apply the same behavior to default entity projections and QuerySelect::column, and mirror it in sea-orm-sync.

## Verification

- Reproduced the original panic with the reported citext mapping in an isolated crate.
- Executed the fixed query end to end against PostgreSQL 18 with the citext extension, including model decoding and case-insensitive filtering.
- Added regression coverage for default and select_only combined selects.
- Passed cargo test --workspace, the full PostgreSQL integration matrix, the sea-orm-sync SQLite matrix, feature checks, Clippy with warnings denied, rustfmt, and Taplo formatting.

### DIFF
--- a/sea-orm-sync/src/entity/column.rs
+++ b/sea-orm-sync/src/entity/column.rs
@@ -843,7 +843,7 @@ mod tests {
         {
             assert_eq!(
                 E::find().build(DbBackend::Postgres).to_string(),
-                r#"SELECT "hello"."id", "hello"."one1", CAST("hello"."two" AS integer), "hello"."three3" FROM "hello""#,
+                r#"SELECT "hello"."id", "hello"."one1", CAST("hello"."two" AS integer) AS "two", "hello"."three3" FROM "hello""#,
             );
             assert_eq!(
                 Update::one(active_model)
@@ -1116,7 +1116,7 @@ mod tests {
         {
             assert_eq!(
                 E::find().build(DbBackend::Postgres).to_string(),
-                r#"SELECT "hello"."id", "hello"."one1", CAST("hello"."two" AS integer), "hello"."three3" FROM "hello""#,
+                r#"SELECT "hello"."id", "hello"."one1", CAST("hello"."two" AS integer) AS "two", "hello"."three3" FROM "hello""#,
             );
             assert_eq!(
                 Update::one(active_model)
--- a/sea-orm-sync/src/entity/column.rs
+++ b/sea-orm-sync/src/entity/column.rs
@@ -736,10 +736,9 @@ mod tests {
         );
     }
 
-    #[test]
     #[cfg(feature = "macros")]
-    fn select_as_1() {
-        use crate::{ActiveModelTrait, ActiveValue, Update};
+    mod select_as {
+        use super::*;
 
         mod hello_expanded {
             use crate as sea_orm;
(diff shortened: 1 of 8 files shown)


Huliiiiii:
Thanks!

Issue #3183, opened 2026-08-22
Nested ActiveModelEx::save() fails to insert a new has_many child with a composite, non-auto-increment primary key


ziimakc:
## Description

Nested `ActiveModelEx::save()` (SeaORM 2.0) picks insert-vs-update per related row via `is_update()`, which just checks whether the primary key columns are `Set`/`Unchanged`. For an entity with a composite, non-auto-increment primary key (e.g. a join/ownership table keyed by `(parent_id, tag)`), the key is always fully set when building a new row, so a brand-new child is misdiagnosed as an update. The resulting `UPDATE` matches zero rows and `save()` fails with `RecordNotFound`, instead of inserting the row.

This works fine for auto-increment primary keys (a fresh row naturally has `NotSet` on the pk), but composite non-auto-increment keys are a common shape for `has_many` children (join tables, per-user records, etc.), so `save()` can't be used to add a new child to an already-persisted parent in that case. The same misdiagnosis also hits any entity whose (single-column) primary key is a UUID assigned application-side rather than by the database — a very common pattern — since that pk is likewise always `Set` on a fresh row, not `NotSet`.

## Steps to reproduce

```rust
use sea_orm::entity::prelude::*;
use sea_orm::{ConnectionTrait, Database, Schema};

mod parent {
    use sea_orm::entity::prelude::*;

    #[sea_orm::model]
    #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
    #[sea_orm(table_name = "parent")]
    pub struct Model {
        #[sea_orm(primary_key, auto_increment = false)]
        pub id: i32,
        #[sea_orm(has_many)]
        pub children: HasMany<super::child::Entity>,
    }

    impl ActiveModelBehavior for ActiveModel {}
}

mod child {
    use sea_orm::entity::prelude::*;

    // Composite, non-auto-increment primary key.
    #[sea_orm::model]
    #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
    #[sea_orm(table_name = "child")]
    pub struct Model {
        #[sea_orm(primary_key, auto_increment = false)]
        pub parent_id: i32,
        #[sea_orm(primary_key, auto_increment = false)]
        pub tag: i32,
        #[sea_orm(belongs_to, from = "parent_id", to = "id")]
        pub parent: HasOne<super::parent::Entity>,
    }

    impl ActiveModelBehavior for ActiveModel {}
}

#[tokio::main]
[...]

PR #3195, opened 2026-09-07
Add ActiveHasMany Mutation (ability to delete)

Jacob-32587:
## Problem

I started upgrading my project to use SeaORM 2.0 and was able to cut out a lot of code that the new `ActiveModelEx` allows. After looking more closely at the docs, I realized that the new `ActiveHasMany` does not have a way to represent “deleted” in its state. This was a huge problem for my project since it uses PATCH requests that send only the information needed to update. The goal of these kind of requests are to reduce the amount of database calls and processing needed.

## Description

I have introduced a third variant to the `ActiveHasMany` enum called `Mutate`. This variant keeps track of models that need to be deleted and models that need to be saved. There are now two new functions exposed on each has many relationship: `delete_{RELATION_NAME}` & `delete_existing_{RELATION_NAME}`. The first function is designed to work if the `ActiveModel` that needs to be deleted is not loaded in, the second is meant to be used if the `ActiveModel` that needs to be deleted is loaded. The `push_*` functions on the `ActiveHasMany enum` also ensure that a model with the same primary key does not appear in the save and delete state at the same time.

## Notes

* I chose to use `swap_remove` when deleting elements from the save/delete vectors so these operations stay fast
for active models with a large number of related entities. This comes with the caveat that elements in a `ActiveHasMany` enum may not retain their order after performing `push_*` operations.

Checklist

- [x] I have read the AI Policy and confirm that this contribution complies with it.

Release Notes

* Add the ability for `ActiveHasMany` models to be marked as deleted.


Huliiiiii:
Thanks for your contribution!

However, there are issues with our existing API design, and the API you built upon it also has some semantic problems. I’d prefer to address these together in a future refactor (see #3183). 

Before implementing it, I’d also like us to first discuss and reach a consensus on the API design.

PR #3199, opened 2026-09-13
Fix left_join_linked joining multi-hop links onto the wrong alias

Dev-next-gen:
Since 994759b8 `left_join_linked` can be chained, and the join aliases are taken from `linked_index` so the second call starts at `r2`, `r3` and so on. The alias for the previous hop was left as `r{i-1}` though, so only the first hop of each call is correct. With a single-hop link (the case covered by `partial_model_join_two_linked`) that never shows, but a multi-hop link chained after another one joins onto the first link's tables.

I hit it with the test config entities:

```rust
cake::Entity::find()
    .left_join_linked(entity_linked::CakeToFilling)
    .left_join_linked(entity_linked::CakeToFillingVendor)
```

which currently builds

```sql
LEFT JOIN `cake_filling` AS `r2` ON `cake`.`id` = `r2`.`cake_id`
LEFT JOIN `filling` AS `r3` ON `r0`.`filling_id` = `r3`.`id`
LEFT JOIN `vendor` AS `r4` ON `r1`.`vendor_id` = `r4`.`id`
```

so `r3` and `r4` are joined through the first link's `r0`/`r1` instead of `r2`/`r3`. No error, just wrong rows. The fix uses `r - 1`, which equals `i - 1` on the first call, so existing single-call queries produce the same SQL as before.

I added `join_26` for this case. It fails on master with the diff above and passes with the change; the other 25 `query::join` tests still pass, in `sea-orm` and in `sea-orm-sync`, where I copied the same change since the two files were identical.

AI tools used


Huliiiiii:
Thanks!

PR #3200, opened 2026-09-15
sea-orm-migration: don't depend on stream feature

ggirol-rc:
The stream feature was introduced in https://github.com/SeaQL/sea-orm/pull/3075 to be able to compile sea-orm without the `aliasable` crate. Unfortunately sea-orm-migration depends on fully featured sea-orm.

This commit makes sea-orm-migration depend only on the features it requires from sea-orm.

## Checklist

- [x] I have read the [AI Policy](https://github.com/SeaQL/.github/blob/master/AI_POLICY.md) and confirm that this contribution complies with it.


Huliiiiii:
Thanks!

Issue #3205, opened 2026-09-17
Per-backend column type overrides, e.g. json on mysql / jsonb on postgres


xue-ding-e:
## Motivation

Column types are not portable across databases, and today SeaORM/sea-query can only express a single fixed column type per column. Two concrete consequences:

- The common "JSON" column needs `json` on mysql and `jsonb` on postgres. There is no single column definition that produces the native right type on both; you either pick one or work around it outside the entity.
- Types that only some databases have (`tinyint`, `longblob`, postgres `interval`, ...) force either a custom `ColumnType::Custom("...")` string that is wrong on every other backend, or a bigger portable type. Long-standing issues #426 (interval), #631 (longblob), #213 (DATE) and the newer composite/domain type requests (#2895, #2863) all sit around this gap.

## Proposed Solutions

I'd like to start a discussion on a per-backend column type override, roughly:

```rust
ColumnDef::new(Column::Payload)
    .string() // abstract fallback for backends without an override
    .per_backend(|backend| match backend {
        Backend::MySql => ColumnType::Json,
        Backend::Postgres => ColumnType::JsonB,
        _ => ColumnType::String(None),
    })
```

(or any equivalent shape — a map from backend to `ColumnType`, an extension of `ColumnType::Custom`, or derivable helpers; the ergonomics are for maintainers to steer).

The semantics that have proven themselves in other ORMs: the backend-specific type wins on that backend, other backends fall back to the abstract type, and columns without overrides render exactly as today.

- ent (Go) has `field.String("s").SchemaType(map[string]string{dialect.MySQL: "varchar(255)", dialect.Postgres: "text"})` as a built-in
- SQLAlchemy has `String(255).with_variant(VARCHAR(255, charset="utf8"), "mysql")`
- gorm recently merged dialect-qualified type tags (go-gorm/gorm#7867)

Happy to collaborate on a design and implementation if maintainers think this fits sea-query/SeaORM's direction.

## Additional Information

The same need shows up across ecosystems; the ent and SQLAlchemy links above are the closest prior art. No behavior would change for existing entities — the fallback path is the current rendering.

Huliiiiii:
I agree with this direction, but the current implementation has some coupling that may affect other parts of the system. It would be even better if you could look into this more deeply and investigate what it might affect, then write up an RFC based on your findings.

xue-ding-e:
[...]

Issue #3106, opened 2026-07-01
Sync: Model column type changes fail silently


berniev:
Entity-First is fantastic! (Using SQLite, sync). New user.

Sync doesn't propagate to database a column type changed in model (eg f32 -> Decimal). I get this could be complicated.

But surely it would not be difficult to detect and alert, and perhaps abort sync?

I think silent failures are almost always a bug. But lodge as feature (benefit of the doubt re design).

PR #3206, opened 2026-09-18
feat: per-backend column type overrides

xue-ding-e:
Follow-up to #3205 and the [RFC comment](https://github.com/SeaQL/sea-orm/issues/3205#issuecomment-5725528712). @Huliiiiii was right that the RFC's impact inventory was incomplete — it never mentioned the 2.0 schema sync pipeline, which is the most important consumer of this path. While implementing this I read through `src/schema/builder.rs` and the picture is now clearer:

- `TableSyncHandle::new` builds its statements via `Schema::create_table_from_entity`, so resolving overrides in `column_def_from_entity_column` covers **both** the direct schema builder and the entity-first sync/apply flows with a single dispatch point.
- The type-divergence warning added for #3106 compares `render_column_type(backend, ...)` on the already-dispatched column defs, so with an override in place, the warning automatically reports the *physical* type the entity intends for that backend — sync stays consistent without extra changes.
- The runtime paths (`select_as` / `save_as` / `cast_enum_as`) keep matching on the logical type, which is what value casting should follow: a `Json` column stored as `jsonb` on Postgres still casts as JSON.

This implements the RFC's step 1 + 2:

```rust
#[sea_orm(table_name = "payload_item")]
pub struct Model {
    #[sea_orm(primary_key)]
    pub id: i32,
    #[sea_orm(column_type = "Json", column_type_postgres = "JsonBinary")]
    pub payload: Json,
    #[sea_orm(column_type = "TinyInteger", column_type_mysql = "SmallInteger")]
    pub level: i16,
}
```

produces `"payload" jsonb` on Postgres, `` `payload` json `` on MySQL, `"payload" json_text` on SQLite — and `` `level` smallint `` on MySQL where the logical `TinyInteger` would render `tinyint`.

Also available at runtime:

```rust
ColumnType::Json.def().type_for(DbBackend::Postgres, ColumnType::JsonBinary)
```

What changed:

- `ColumnDef` (src/entity/column_def.rs): a `pub(crate) col_type_overrides: Vec<(DbBackend, ColumnType)>` field plus `type_for` / `get_column_type_override_for`. All fields there are already `pub(crate)`, so this is additive; columns without overrides carry an empty Vec and behave byte-identically to before.
- `column_def_from_entity_column` (src/schema/entity.rs): the override lookup sits ahead of the existing match, so the Enum rewriting keeps working for columns without overrides.
- `DeriveEntityModel` (sea-orm-macros): `column_type_mysql` / `column_type_postgres` / `column_type_sqlite` attributes, parsed with the same helper as `column_type` and emitted as `.type_for(...)` in the generated `def()`.

What deliberately did not change: sea-query (no new `ColumnType` variant, so no new match arms in the backend renderers, codegen `get_rs_type`, or the hand-written `PartialEq`), the runtime cast paths, sea-schema discovery, Seaography, and sea-orm-migration (SchemaManager statements are hand-written sea-query; authors who branch per backend already can).

Local checks: `cargo test --workspace`, `cargo clippy --all -- -D warnings` (both feature sets from CI) and `cargo fmt --check` on the touched files all pass. The remaining open questions from the RFC (attribute naming, Enum × override precedence, codegen policy) still stand — happy to adjust to whatever shape you prefer.


Huliiiiii:

The current PR description is unreadable.
Please take the time to rewrite it in your own words and make it clear and understandable.

PR #3207, opened 2026-09-20
docs: drop redundant install block from CLI and migration READMEs

toyeshhm:
`cargo install --bin sea` in both READMEs can't work on its own — cargo needs a crate name.

Per review, dropping the block rather than fixing the command: `cargo install sea-orm-cli` already installs `sea` as well. Both `[[bin]]` targets in `sea-orm-cli/Cargo.toml` are gated on `cli`/`codegen`, and `codegen` is a default feature, so the second block was never doing anything the first didn't.


Huliiiiii:
Should we just remove the `or` paragraph? I did some refactors to the CLI earlier, which probably caused this issue. Does it still install an additional `sea` binary when no explicit `--bin` is specified?

toyeshhm:
Yeah, it still installs `sea`. Both `[[bin]]` targets in `sea-orm-cli/Cargo.toml` are gated on `cli`/`codegen`, and `codegen` is a default feature, so `cargo install sea-orm-cli` builds `sea-orm-cli` and `sea` both.

So the second block wasn't buying anything. Dropped it from both READMEs.

If the `sea` alias is worth advertising I can put a line under the first block instead, just say the word.

Huliiiiii:
> just say the word

😅

Fine. Thanks bot.

